PR #15, opened 2015-03-19
create RestTemplate with LoadBalancerInterceptor using @LoadBalanced qua...

spencergibb:
...lifier.

fixes https://github.com/spring-cloud/spring-cloud-netflix/issues/267

@dsyer is this along the lines of what you were thinking?


dsyer:
I like the `@LoadBalanced`. Where does it get used though? Don't you need to inject it somewhere?

spencergibb:
We don't inject `RestTemplate` anywhere, just make it available to users.  When you want the load balanced `RestTemplate` you do:

```
@Autowired
@LoadBalanced
private RestTemplate loadBalanced;
```

dsyer:
OK, I like. Can we do the same thing with `OAuth2RestTemplate`?

spencergibb:
yup

spencergibb:
in the case of `OAuth2RestTemplate` should we create both one with `@LoadBalanced` and one without?

spencergibb:
https://github.com/spring-cloud/spring-cloud-security/issues/51

PR #27, opened 2015-04-25
A proposed fix for issue #26 fixes #26

### LINKED ISSUE #26: PropertySourceBootstrapConfiguration: Would it be possible to allow the override of any property from the config server
The current implementation of the PropertySourceBootstrapConfiguration. Only allows the overriding of values from the bootstrap configured PropertySources by source before the System Environment Property source. Would it be possible to allow the overriding from any property source e.g standard application.properties

Also the configuration properties to allow overriding

```
spring.cloud.config.allowOverride
spring.cloud.config.overrideSystemProperties
```

Look like they must come from the bootstrap configured sources as the standard bootstrap property source is removed before calling `PropertySourceBootstrapConfiguration.insertPropertySources()`.

Would be nice if these value if present in the bootstrap.yaml or bootstrap.properties were respected.

bstick12:
This would be the functionality I was suggesting.

If you are happy with the changes and property names I'll do up some unit tests.

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/config/PropertySourceBootstrapConfiguration.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/config/PropertySourceBootstrapConfiguration.java
@@ -83,7 +83,7 @@ public class PropertySourceBootstrapConfiguration implements
 		}
 		if (!empty) {
 			MutablePropertySources propertySources = applicationContext.getEnvironment()
-					.getPropertySources();
+					.getPropertySources();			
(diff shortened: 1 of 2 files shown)


bstick12:
I see I've made an invalid assumption in my pull request that the Property Source named 'bootstrap' actually contains the properties from bootstrap.properties or bootstrap.yml. 

I also see that this issue was previously requested - https://github.com/spring-cloud/spring-cloud-config/issues/57 and marked as Closed but I don't see how this would allow me to set the bootstrapped propertysources as the last PropertySources.

Just for some background I'm not using the ConfigServer but a custom property source that delegates to ZooKeeper for properties but we want the ability to specify the overrides locally.

PR #41, opened 2015-06-24
service capabilities

mbenson:
spring-cloud-commons portion of https://github.com/spring-cloud/spring-cloud-commons/issues/19

PR #56, opened 2015-11-01
Discovery metadata

jmnarloch:
This pull request introduces the metadata map that can be associated with each service instance registered within discovery service (currently covered by Eureka and Zookeper, Consul is planned to have such support in version 2.0) allowing to use that information on the client side.

I had to also align the version with latest spring boot snapshot where some class has been renamed, this is way there are two commits.

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
@@ -16,9 +16,12 @@
 
 package org.springframework.cloud.client;
 
-import java.net.URI;
-
 import lombok.Data;
+import lombok.RequiredArgsConstructor;
+
+import java.net.URI;
+import java.util.Collections;
+import java.util.Map;
 
 /**
  * Default implementation of {@link ServiceInstance}.
@@ -26,6 +29,7 @@ import lombok.Data;
  * @author Spencer Gibb
  */
 @Data
+@RequiredArgsConstructor
 public class DefaultServiceInstance implements ServiceInstance {
 
 	private final String serviceId;
@@ -36,9 +40,20 @@ public class DefaultServiceInstance implements ServiceInstance {
 
 	private final boolean secure;
 
+	private final Map<String, String> metadata;
+
+	public DefaultServiceInstance(String serviceId, String host, int port, boolean secure) {
+		this(serviceId, host, port, secure, Collections.<String, String>emptyMap());
+	}
+
 	@Override
 	public URI getUri() {
 		return getUri(this);
+	}
+
+	@Override
+	public Map<String, String> getMetadata() {
+		return metadata;
 	}
 
 	/**
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/ServiceInstance.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/ServiceInstance.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.client;
 
 import java.net.URI;
+import java.util.Map;
 
 /**
  * Represents an instance of a Service in a Discovery System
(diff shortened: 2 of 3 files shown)


jmnarloch:
Probably it's worth getting back to the issue despite the https://github.com/spring-cloud/spring-cloud-netflix/issues/419 and https://github.com/spring-cloud/spring-cloud-commons/pull/41

My idea is simply, if the backing discovery service can provide the client with the metadata (like Eureka and Zookeeper) we can cover that, for Consul it could return the tags information for version 1.x and when version 2.0 will be covered the metadata could be provided from there (issues: https://github.com/hashicorp/consul/issues/1131, https://github.com/hashicorp/consul/issues/697, https://github.com/hashicorp/consul/issues/1107).

spencergibb:
At a glance, I think I like this better than #41.

jmnarloch:
The only question is whether a "static" Map would be sufficient and won't be a need to do some lookup on the fly and there therefor introduce some InstanceMetadata abstraction.

spencergibb:
When you say, "static" map, do you mean just a `java.util.Map`?

jmnarloch:
Yes, exactly. I was only deliberating whether introducing there would be a need to introduce some wrapper on top of the map, for instance if you could declare some interface (let's call if InstanceMetadata) then the implementation could have much more flexibility. Although I not sure whether there will be ever need for that, so for now I just propose simple key/value map for storing the metadata.

spring-cloud-commons/src/test/java/org/springframework/cloud/client/loadbalancer/LoadBalancerAutoConfigurationTests.java
```diff
@@ -116,7 +113,8 @@ LoadBalancerClient loadBalancerClient() {
 
         @Override
         public ServiceInstance choose(String serviceId) {
-            return new DefaultServiceInstance(serviceId, serviceId, random.nextInt(40000), false);
+            return new DefaultServiceInstance(serviceId, serviceId, random.nextInt(40000),
+                    false, Collections.<String, String>emptyMap());
```
spencergibb: I think `DefaultServiceInstance` needs to retain the original constructor signature as well as the new one. Hadn't noticed this before.

jmnarloch: Done, thanks for noticing.

spencergibb:
Thanks @jmnarloch!

spencergibb:
@jmnarloch so I started looking at this. The two netflix implementations in [`EurekaDiscoveryClient`](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/eureka/EurekaDiscoveryClient.java) were simple to implement. I ran into trouble in [`RibbonLoadBalancerClient`](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonLoadBalancerClient.java#L138-L180). The ribbon `Server` doesn't even expose the underlying `InstanceInfo` where the metadata is located. Wondering if you had any thoughts?

spencergibb:
I'd have to do something like this

		protected RibbonServer(String serviceId, Server server, boolean secure) {
			this.serviceId = serviceId;
			this.server = server;
			this.secure = secure;

			if (this.server instanceof DiscoveryEnabledServer) {
				Field instanceInfoField = ReflectionUtils.findField(DiscoveryEnabledServer.class, "instanceInfo");
				ReflectionUtils.makeAccessible(instanceInfoField);
				try {
					InstanceInfo instanceInfo = (InstanceInfo) instanceInfoField.get(this.server);
					this.metadata = instanceInfo.getMetadata();
				} catch (IllegalAccessException e) {
					log.error("Error getting instanceInfo for serviceId: "+serviceId, e);
				}
			}
		}

@dsyer @adriancole @marcingrzejszczak thoughts?

spencergibb:
The above stinks because it won't work for s-c-zk and s-c-consul, which will use `RibbonServer` but not `DiscoveryEnabledServer`.

spencergibb:
Actually we already have a `ServerIntrospector` abstraction we could add ` Map<String, String> getMetadata(Server server)`.

spencergibb:
And there is a `getInstanceInfo` method, so no reflection needed, I'm happy.

spencergibb:
Closed with some polish via aa2ddce31e6e293210df5dda456dfb82402fe16f

marcingrzejszczak:
The metadata feature is a very food idea. It can be combined together with the "dependency" approach that we have in Zookeeper.

PR #67, opened 2015-12-16
Skips docker's interfaces

zanella:
The valid IP can be used for communicating w/ docker and w/ the outside, which is probably what you want when developing

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/util/InetUtils.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/util/InetUtils.java
@@ -42,6 +42,18 @@ public class InetUtils {
 		return hostInfo;
 	}
 
+	/*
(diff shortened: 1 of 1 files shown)


spencergibb:
@dsyer @adriancole @marcingrzejszczak This pains me that I wrote this as a static method. I'd like to make the interfaces to skip configurable. Should I add a parameter that takes a configuration properties object, or move this to a bean? As I type the question, the answer seems to me to move it to a bean if I'm going to make it configurable.

dsyer:
Yep. It's not too late for 1.1.0. The ship has sailed on Brixton.M4 I think (so please don't change the architecture in anything that depends on commons till after that).

spencergibb:
@dsyer will do.

spencergibb:
Closing in favor of #75. Thanks for getting the ball rolling on this.

pcornelissen:
So this won't go into Brixton at all?

PR #68, opened 2015-12-17
Polish doc

snicoll:
I did not manage to fix the include for `https://github.com/spring-cloud/spring-cloud-build/blob/master/docs/src/main/asciidoc/contributing-docs.adoc` which still show up as plain text in the generated doc.

Maybe the `RibbonAutoConfiguration` reference is a bit out of place in an introduction?

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-commons.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-commons.adoc
@@ -1,12 +1,13 @@
 = Cloud Native Applications
-:github: https://github.com/spring-cloud/spring-cloud-commons
+:githubroot: https://github.com/spring-cloud
+:github: {githubroot}/spring-cloud-commons
 :githubmaster: {github}/tree/master
 :docslink: {githubmaster}/docs/src/main/asciidoc
 :toc:
 
 include::intro.adoc[]
 
-include::https://raw.githubusercontent.com/spring-cloud/spring-cloud-build/master/docs/src/main/asciidoc/contributing-docs.adoc[]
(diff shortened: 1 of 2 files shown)


docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -1,12 +1,13 @@
...
+:githubroot: https://github.com/spring-cloud
+:github: {githubroot}/spring-cloud-commons
 :githubmaster: {github}/tree/master
 :docslink: {githubmaster}/docs/src/main/asciidoc
 :toc:
 
 include::intro.adoc[]
 
-include::https://raw.githubusercontent.com/spring-cloud/spring-cloud-build/master/docs/src/main/asciidoc/contributing-docs.adoc[]
+include::https://github.com/spring-cloud/spring-cloud-build/blob/master/docs/src/main/asciidoc/contributing-docs.adoc[]
```
spencergibb: Not sure why this doesn't get included. The `blob` link renders a github page so if it does get included it would break.

snicoll: Look at the current doc. It just doesn't work. I don't know either.

spencergibb: As far as I can see this is the only place we try to include content via a url in our main docs. The readme's do it, but they get processed differently. Maybe this should just be removed or copied in for now until that get's figured out.

spencergibb:
Closed via 78e4786aa95a033a4eac4f52b88d6532ac338cf5
Thanks @snicoll!

PR #75, opened 2015-12-21
Allow configurable ignoring of network interfaces.

spencergibb:
fixes gh-67

Waiting to merge until Brixton.M4 is released.

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/util/InetUtils.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/util/InetUtils.java
@@ -26,6 +26,103 @@ import lombok.extern.apachecommons.CommonsLog;
 @CommonsLog
 public class InetUtils {
 
+	private final ExecutorService executorService;
+	private final InetUtilsProperties properties;
+
+	public InetUtils(final InetUtilsProperties properties) {
+		this.properties = properties;
+		this.executorService = Executors
+				.newSingleThreadExecutor(new ThreadFactory() {
+					@Override
+					public Thread newThread(Runnable r) {
+						Thread thread = new Thread(r);
+						thread.setName(properties.getExecutorThreadName());
+						thread.setDaemon(true);
+						return thread;
+					}
+				});
+	}
+
+	public HostInfo findFirstNonLoopbackHostInfo() {
+		InetAddress address = findFirstNonLoopbackAddress();
+		if (address != null) {
+			return convertAddress(address);
+		}
+		HostInfo hostInfo = new HostInfo();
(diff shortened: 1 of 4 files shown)


spring-cloud-commons/src/test/java/org/springframework/cloud/util/InetUtilsTests.java
```diff
@@ -31,23 +34,43 @@
...
-		HostInfo info = InetUtils.getFirstNonLoopbackHostInfo();
+		HostInfo info = new InetUtils(new InetUtilsProperties()).findFirstNonLoopbackHostInfo();
 		assertNotNull(info.getIpAddressAsInt());
+
+		info = InetUtils.getFirstNonLoopbackHostInfo();
+		assertNotNull(info.getIpAddressAsInt());
+	}
+
+	@Test
+	public void testIgnoreInterface() {
```
adriancole: supernit: add javadoc or comment or rename test method to say this is to support docker overlay networks. Not sure the impact of blacklist vs whitelist on usability.

adriancole: actually maybe just lift description from here? https://github.com/zanella/spring-cloud-commons/commit/f54600d0d785e60d09bb079cb2c3a4687302ce1a#diff-019d46b3915e9bcb2b0db64b0d5219f9R46

adriancole:
LGTM

spring-cloud-commons/src/main/java/org/springframework/cloud/util/InetUtils.java
```diff
@@ -26,6 +26,103 @@
 @CommonsLog
 public class InetUtils {
 
+	private final ExecutorService executorService;
+	private final InetUtilsProperties properties;
+
+	public InetUtils(final InetUtilsProperties properties) {
+		this.properties = properties;
+		this.executorService = Executors
```
adriancole: feel free to not address this now, but this will eagerly create a thread, even if never used. This is a way out, if interested: https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/framed/FramedConnection.java#L156

fiunchinho:
:+1:

spencergibb:
Closed via a42e8eaab5a4ba29b7119fba8fc7bcb0577ec900

pcornelissen:
Will this be included in brixton.m5? 
I am in desperate need for this, because as soon as I start a vpn connection on my machine all locally running spring boot instances reconfigure to this IP and are not accessible afterwards (which in itself is strange, but the routes that come with the VPN seem to be a problem). So blacklisting certain Interfaces would be great.

PR #84, opened 2016-01-30
Added support for service IP address

nickcodefresh:
I need the ability to retrieve the IP address of a service instance in Consul, so the first part of the change I need to make to Spring Cloud Consul is to add the required "address" field to org.springframework.cloud.client.ServiceInstance.


spencergibb:
Can you tell me why you need it? There's a setting to prefer ipaddress which would end up in host.

nickcodefresh:
I didn't know they're was a setting for this.  What's it the key name?

dsyer:
https://github.com/spring-cloud/spring-cloud-consul/blob/master/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java#L88

Issue #87, opened 2016-02-03
Decrypted values from environment variables not being picked up


tanob:
Given a Spring bean like this:

```java
@Component
public class Foo {
    private final String fooPassword;

    @Autowired
    public Foo(@Value("${foo.password}") String fooPassword) {
        System.out.println("foo password is " + fooPassword);
        this.fooPassword = fooPassword;
    }
}
```

when we run the app providing an `ENCRYPT_KEY` and an encrypted value for `foo.password` using the relaxed binding ([reference 1](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-relaxed-binding) and [reference 2](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-application-property-files)) through `FOO_PASSWORD`, it is not printing the decrypted value for `foo.password`:

```
$ ENCRYPT_KEY=foo FOO_PASSWORD={cipher}23cd3b5b92db6791c531fda2e2bdd51f2dd3d94d9f1de7b5ba932b5501dd72fc ./gradlew bootRun | grep "foo password is"
foo password is {cipher}23cd3b5b92db6791c531fda2e2bdd51f2dd3d94d9f1de7b5ba932b5501dd72fc
```

The expected output would be:

```
foo password is mypassword
```

The same works if we change the environment variable name to match the dot-syntax one:

```
$ env ENCRYPT_KEY=foo foo.password={cipher}23cd3b5b92db6791c531fda2e2bdd51f2dd3d94d9f1de7b5ba932b5501dd72fc ./gradlew bootRun | grep "foo password is"
foo password is mypassword
```

(Notice I'm using the `env` command to be able to pass an env variable with a dot in the name, [more info here](http://unix.stackexchange.com/a/93533))

You can see the full example here:
https://github.com/tanob/spring-relaxed-binding-and-decrypted-values

dsyer:
This is nothing to do with encryption is it? You can't bind `FOO_VALUE` to `foo.value` with an `@Value`. Relaxed binding is a Spring Boot feature and `@Value` is from Spring Core. And none  of it is relevant for encryption or Spring Cloud.

dsyer:
I see from the PR that you are making cunning use of `SystemEnvironmentPropertySource`. I'm not sure that's entirely kosher, but I'll have a think about it.

tanob:
[...]

PR #89, opened 2016-02-04
Fix property source for decrypted variables

tainacaetano:
This fixes the issue described at #87.

When resolving a property that comes as an environment variable and is encrypted (MY_PASSWORD={cipher}ahsbcssb235563ssc), [PropertySourcesPropertyResolver](https://github.com/spring-projects/spring-framework/blob/64ce8a81c3f6d5ef99ee6cd8220e3c8ad9e5f846/spring-core/src/main/java/org/springframework/core/env/PropertySourcesPropertyResolver.java) searches for the properties in the ``decrypted`` property source. 

The decrypted property (MY_PASSWORD) is not found, however, because [MapPropertySource](https://github.com/spring-projects/spring-framework/blob/d5ee787e1e6653257720afe31ee3f8819cd4605c/spring-core/src/main/java/org/springframework/core/env/MapPropertySource.java) doesn't have an implementation to solve relaxed binding. This leads PropertySourcesPropertyResolver to move on and find ``MY_PASSWORD`` in the ``systemEnvironment`` property source, which still has the encrypted value.

Changing ``EnvironmentDecryptApplicationInitializer`` implementation to use a ``SystemEnvironmentPropertySource`` internally enables it to lookup property names that use relaxed binding.

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializer.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializer.java
@@ -33,10 +33,10 @@ import org.springframework.core.Ordered;
 import org.springframework.core.env.CompositePropertySource;
 import org.springframework.core.env.ConfigurableEnvironment;
 import org.springframework.core.env.EnumerablePropertySource;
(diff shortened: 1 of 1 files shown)


dsyer:
Thanks. Did you fill in the contributor's agreement (link in README)?

tainacaetano:
Just did :)

PR #103, opened 2016-03-21
Fixes #101: Moved existing RefreshEndpoint logic to a new RefreshSupport class.

### LINKED ISSUE #101: Configuration refreshing depends on actuator
See original discussion https://github.com/spring-cloud/spring-cloud-config/issues/352.

@dsyer I agree with @kurtinaitis, I use the `RefreshEvent` which has a listener in commons that does
`RefreshEndpoint.refresh()`. I have to guard that `RefreshEndpoint` is a bean in consul and zookeeper for their watches to trigger a refresh.

`RefreshEndpoint.refresh()` does the following, could be extracted.

		Map<String, Object> before = extract(
				this.context.getEnvironment().getPropertySources());
		addConfigFilesToEnvironment();
		Set<String> keys = changes(before,
				extract(this.context.getEnvironment().getPropertySources())).keySet();
		this.context.publishEvent(new EnvironmentChangeEvent(keys));
		this.scope.refreshAll();
		return keys.toArray(new String[keys.size()]);

venilnoronha:
Hi,

I've moved RefreshEndpoint's logic to RefreshSupport and delegated the _refresh_ call to it. One can now use RefreshSupport outside actuator. Please verify and merge.

My ICLA Number: 149720151120072904

Thanks,
Venil Noronha

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/RefreshEndpointAutoConfiguration.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/RefreshEndpointAutoConfiguration.java
@@ -35,14 +35,13 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.cloud.bootstrap.config.PropertySourceBootstrapConfiguration;
 import org.springframework.cloud.context.environment.EnvironmentChangeEvent;
 import org.springframework.cloud.context.properties.ConfigurationPropertiesRebinder;
-import org.springframework.cloud.context.refresh.RefreshSupport;
+import org.springframework.cloud.context.refresh.ContextRefresher;
 import org.springframework.cloud.context.restart.RestartEndpoint;
 import org.springframework.cloud.context.scope.refresh.RefreshScope;
 import org.springframework.cloud.endpoint.RefreshEndpoint;
 import org.springframework.cloud.endpoint.event.RefreshEventListener;
 import org.springframework.cloud.health.RefreshScopeHealthIndicator;
 import org.springframework.context.ApplicationListener;
-import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.core.env.ConfigurableEnvironment;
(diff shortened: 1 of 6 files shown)


spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/RefreshSupport.java
```diff
@@ -0,0 +1,206 @@
...
+import org.springframework.core.env.MutablePropertySources;
+import org.springframework.core.env.PropertySource;
+import org.springframework.core.env.StandardEnvironment;
+import org.springframework.web.context.support.StandardServletEnvironment;
+
+/**
+ * 
+ * @author Venil Noronha
+ */
+public class RefreshSupport {
```
spencergibb: Not so keen on the name. Maybe `EnvironmentRefresher`? @dsyer?

spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/RefreshEndpointAutoConfiguration.java
```diff
@@ -112,9 +117,15 @@ public RestartEndpoint restartEndpoint() {
 
 		@Bean
 		@ConditionalOnMissingBean
-		public RefreshEndpoint refreshEndpoint(ConfigurableApplicationContext context,
+		public RefreshSupport refreshSupport(ConfigurableApplicationContext context,
```
spencergibb: This bean is still guarded by `@ConditionalOnClass(Endpoint.class)`. It should be reworked so it is not.

spencergibb: It should probably go in `RefreshAutoConfiguration`.

spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/RefreshSupport.java
dsyer: Yeah, need a better name. We already have `EnvironmentManager`. What about `ContextRefresher`? I like the way that `ApplicationEvents` provide an extensibility already, though (you can listen for the event if you want to refresh more stuff). Maybe the `ContextRefresher` could just emit an event, and all the other features can be listeners?

spencergibb: And log the `keys` as a different type of event?

venilnoronha: I've renamed the class to `ContextRefresher`. I didn't get the `ApplicationEvents` part. Are we talking about publishing a new event along with `EnvironmentChangeEvent`?

spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/RefreshEndpointAutoConfiguration.java
venilnoronha: I've moved it to `RefreshAutoConfiguration`.

venilnoronha:
@spencergibb I've done the necessary changes. Could you please review? Also, do guide me with the `ApplicationEvents` change if need be.

spencergibb:
@dsyer is this ok after we've gone into RC mode?

dsyer:
Fine I think (it's not really a new feature).

spencergibb:
Squashed and merged via 1e71c5632c01736fbe5be3b399318de23f0f0c3a

venilnoronha:
Thank you!

PR #110, opened 2016-05-04
Catch and log exceptions that occur when attempting to deregister a s…

cfritz0127:
…ervice during shutdown.

Currently, if an exception is thrown when deregistering the service (maybe the registry is not available) the latch used by the DefaultLifecycleProcessor is not decremented and you have to wait for the latch timeout (30 seconds) for shutdown to complete. With this fix, the exception is logged and normal shutdown execution continues.


spencergibb:
Have you signed the [Contributor License Agreement](https://support.springsource.com/spring_committer_signup)?

cfritz0127:
I have now :)

spencergibb:
Closed with a bit of polish via 8c9317bd7e5f5f62ab761d3b27aa24173d394fea

Issue #9, opened 2015-01-28
Create ServiceRegistry interface with register/unregister


spencergibb:

dsyer:
I would argue that register/unregister is a "registry" concern not a "discovery" concern. New interface?

spencergibb:
Sure `s-c-commons-client-registry.Registry`?   I don't like that `client` is in the package structure in commons.

dsyer:
`ServiceRegistry` is what most people call it I guess? We can take the "client" out of the package name if you prefer, but isn't it better to be explicit about that (in case someone thinks they can use that package to implement a server)?

spencergibb:
`ServiceRegistry` it is.  I was worried about having "client" in the package, just second guessing myself.

spencergibb:
hmm, there doesn't seem to be a way to arbitrarily register services with eureka v1.  All their usage of statics limits registration to the current instance.  I saw register looking something like

```
public void register(ServerInstance);
```

PR #119, opened 2016-07-02
Create a ServiceRegistry interface.

spencergibb:
Allows service discovery systems to register more than one instance.

Automatic registration will still take place.

fixes gh-9

/cc @dsyer @adriancole @marcingrzejszczak

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/AbstractDiscoveryLifecycle.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/AbstractDiscoveryLifecycle.java
@@ -33,16 +33,14 @@ import org.springframework.context.ApplicationListener;
 import org.springframework.core.env.Environment;
 
 /**
- * Lifecycle methods that may be useful and common to {@link ServiceRegistry} implementations.
+ * Lifecycle methods that may be useful and common to various DiscoveryClient implementations.
  *
- * TODO: document the lifecycle
- *
- * @param <R> registration type passed to the {@link ServiceRegistry}.
+ * @deprecated use {@link org.springframework.cloud.client.serviceregistry.AbstractAutoServiceRegistration} instead. This class will be removed in the next release train.
  *
  * @author Spencer Gibb
  */
-//TODO: rename to AbstractServiceRegistryLifecycle or AbstractAutoServiceRegistration?
-public abstract class AbstractDiscoveryLifecycle<R> implements DiscoveryLifecycle,
+@Deprecated
+public abstract class AbstractDiscoveryLifecycle implements DiscoveryLifecycle,
 		ApplicationContextAware, ApplicationListener<EmbeddedServletContainerInitializedEvent> {
 
 	private static final Log logger = LogFactory.getLog(AbstractDiscoveryLifecycle.class);
@@ -58,12 +56,6 @@ public abstract class AbstractDiscoveryLifecycle<R> implements DiscoveryLifecycl
 	private Environment environment;
 
 	private AtomicInteger port = new AtomicInteger(0);
-
-	private ServiceRegistry<R> serviceRegistry;
-
-	protected AbstractDiscoveryLifecycle(ServiceRegistry<R> serviceRegistry) {
-		this.serviceRegistry = serviceRegistry;
-	}
 
 	protected ApplicationContext getContext() {
 		return context;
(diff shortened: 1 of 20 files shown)


spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/AbstractDiscoveryLifecycle.java
```diff
@@ -26,6 +26,8 @@
 import org.springframework.beans.BeansException;
 import org.springframework.boot.context.embedded.EmbeddedServletContainerInitializedEvent;
 import org.springframework.cloud.client.discovery.event.InstanceRegisteredEvent;
+import org.springframework.cloud.client.serviceregistry.Registration;
```
spencergibb: I don't like how this is mixed up here. How about I move `*Lifecycle` to it's own package?

spencergibb: In fact, `DiscoveryLifecycle` doesn't even make sense anymore. Maybe `AutoRegistration`. It's not a function of discovery at all.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/AbstractDiscoveryLifecycle.java
```diff
@@ -122,28 +130,44 @@ protected boolean shouldRegisterManagement() {
 	/**
 	 * @return the object used to configure the DiscoveryClient
```
adriancole: or null when ____

spring-cloud-commons/src/main/java/org/springframework/cloud/client/serviceregistry/Registration.java
```diff
@@ -0,0 +1,7 @@
+package org.springframework.cloud.client.serviceregistry;
+
+/**
+ * @author Spencer Gibb
+ */
+public interface Registration {
```
adriancole: why is a marker interface needed?

spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/AbstractDiscoveryLifecycle.java
```diff
@@ -35,7 +37,7 @@
  * Lifecycle methods that may be useful and common to various DiscoveryClient implementations.
```
adriancole: is it still various? or is registration (the `R` type) intended to work only with one sort of client now? I'd probably doc the type param regardless of whether it stays a marker or not.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/AbstractDiscoveryLifecycle.java
```diff
@@ -122,28 +130,44 @@ protected boolean shouldRegisterManagement() {
...
+	protected abstract R getRegistration();
+
+	protected abstract R getManagementRegistration();
+
+	protected ServiceRegistry<R> getServiceRegistry() {
+		return this.serviceRegistry;
+	}
 
 	/**
 	 * Register the local service with the DiscoveryClient
```
adriancole: does this doc need to be changed now?

spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/AbstractDiscoveryLifecycle.java
```diff
@@ -122,28 +130,44 @@ protected boolean shouldRegisterManagement() {
...
 
 	/**
 	 * Register the local management service with the DiscoveryClient
 	 */
 	protected void registerManagement() {
+		this.serviceRegistry.register(getManagementRegistration());
 	}
 
 	/**
 	 * De-register the local service with the DiscoveryClient
```
adriancole: ditto

spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/AbstractDiscoveryLifecycle.java
```diff
@@ -122,28 +130,44 @@ protected boolean shouldRegisterManagement() {
...
 	 */
-	protected abstract void deregister();
+	protected void deregister() {
+		this.serviceRegistry.deregister(getRegistration());
+	}
 
 	/**
 	 * De-register the local management service with the DiscoveryClient
```
adriancole: ditto

spring-cloud-commons/src/main/java/org/springframework/cloud/client/serviceregistry/ServiceRegistry.java
```diff
@@ -0,0 +1,10 @@
+package org.springframework.cloud.client.serviceregistry;
+
+/**
+ * @author Spencer Gibb
+ */
+public interface ServiceRegistry<R extends Registration> {
```
adriancole: here or somewhere else, we should qualify with an example of what a registration might be

spring-cloud-commons/src/main/java/org/springframework/cloud/client/serviceregistry/Registration.java
marcingrzejszczak: Yeah, cause looking at the test I don't really see any benefit of having this `Registration` interface. Can you explain that?

marcingrzejszczak:
Can you explain in a few words how do you see it work? Like what the flow is cause I don't quite follow it ATM :/

spring-cloud-commons/src/main/java/org/springframework/cloud/client/serviceregistry/Registration.java
spencergibb: Yeah, leftover from a refactor where this interface had a method and I removed it.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/AbstractDiscoveryLifecycle.java
spencergibb: `R` will be specific to the impl (though it won't extend `Registration`).

spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/AbstractDiscoveryLifecycle.java
spencergibb: Yeah, `DiscoveryClient` was never right, since it wasn't used.

yidongnan:
Is there any plan to merge and release it?

spencergibb:
Closed via bf4b6e32938745cb57ee795c4f7544cd417eb9fa

@yidongnan see spring-cloud-release [milestones](https://github.com/spring-cloud/spring-cloud-release/milestones?direction=asc&sort=due_date&state=open) for tentative release dates.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/serviceregistry/Registration.java
spencergibb: I ended up putting this back in, because I needed to lookup the registration in the spring context without knowing the implementation.

PR #123, opened 2016-07-21
consider spring.profiles.include values from bootstrap PropertySources

mbenson:
Discussed on gitter; if remote `PropertySourceLocator`s such as `ConfigServicePropertySourceLocator` can recursively pull in resources based on the content of their locally defined `spring.profiles.include` property values (this is actually done by the config server itself, but it _is_ done), it follows that those same values should be considered for addition to the `Environment`'s active profiles.

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/config/PropertySourceBootstrapConfiguration.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/config/PropertySourceBootstrapConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2014 the original author or authors.
+ * Copyright 2013-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,8 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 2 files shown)


spencergibb:
Trying to understand the need. Did you discuss with @dsyer?

mbenson:
It might have helped if I hadn't mistyped the description of the problem. Here is the short exchange from gitter:

> Matt Benson @mbenson Jul 21 11:22
> so the thing I brought up yesterday seems like a @dsyer thing... since you are the author of record
> for PropertySourceBootstrapConfiguration. For example, the ConfigServicePropertySourceLocator 
> returns profile documents picked up due to a spring.profiles.include property returned by the config 
> service. Now, I don't really think it makes sense that spring.profiles.active could be modified in the 
> PSBC, but it would seem a straightforward code change to pick up spring.profiles.include after 
> modifying the [Configurable]Environment.
> (if one that requires a bit of code duplication from e.g. ConfigFileApplicationListener)

> Dave Syer @dsyer Jul 21 12:29
> I didn't really follow what the problem was, sorry.
> But if you think you can solve it, go ahead

mbenson:
To better explain the need, this supports composition of profiles, where one profile can include another. This works with local property resources; it is my contention that the behavior should be the same for remote property resources.

spencergibb:
Thanks @mbenson for the context.

spencergibb:
Trying to understand. This adds profiles set in `spring.profiles.include` on the client, then those profiles get added to the set that gets sent to the config server?

mbenson:
Nope. ;) e.g., application sends `prod` to the config server, where perhaps there is:

`application.yml`:
```
spring:
  profiles: prod
spring.profiles.include=aws
---
spring:
  profiles: aws
...
```

The config server includes the settings from the `aws`-specific subdocument, respecting the `spring.profiles.include` property included for the `prod` profile; therefore it seems to me reasonable that `PropertySourceBootstrapConfiguration` handle the same property setting(s) and include the specified profiles into the local `Environment`.

spencergibb:
Thanks again.

spencergibb:
Sorry for the delay, looking to clear the queue this week.

PR #130, opened 2016-09-09
Fixes indexed properties merging problem.

spencergibb:
Given the following in one property source
```
mine[0].someValue=Foo
mine[0].someKey={cipher}XXXX
mine[1].someValue=Bar
mine[1].someKey={cipher}XXXX
yours[0].anotherKey=Baz
```
only `mine[*].someKey` were added to "decrypted" property source. Due to house spring boot manages indexed properties, they can not be merged across property source. This change includes all indexed properties in the "decrypted" property source if any one of them is encrypted.

In the above example, `yours[*].*` is included even though unrelated.

fixes https://github.com/spring-cloud/spring-cloud-config/issues/466

/cc @dsyer @philwebb


ryanjbaxter:
@spencergibb was finally able to test this fix with my sample and it worked.  Everything looks good to me!

PR #137, opened 2016-10-04
Retry implementation for LoadBalanced RestTemplates

ryanjbaxter:
Helps address spring-cloud/spring-cloud-netflix#1290

### DIFF
--- a/spring-cloud-commons/pom.xml
+++ b/spring-cloud-commons/pom.xml
@@ -6,7 +6,7 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-commons-parent</artifactId>
-		<version>1.1.4.BUILD-SNAPSHOT</version>
+		<version>1.1.5.BUILD-SNAPSHOT</version>
 		<relativePath>..</relativePath>
 	</parent>
 	<artifactId>spring-cloud-commons</artifactId>
@@ -51,6 +51,16 @@
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-hateoas</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-aop</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.retry</groupId>
+			<artifactId>spring-retry</artifactId>
 			<optional>true</optional>
 		</dependency>
 		<dependency>
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalanceChooser.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalanceChooser.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.client.loadbalancer;
+
+import org.springframework.cloud.client.ServiceInstance;
+
+/**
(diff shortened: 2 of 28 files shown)


william-tran:
How would I specify retry on 404 behaviour? Is there still the issue of the underlying client not throwing on such errors and so not triggering retry behaviour?

ryanjbaxter:
Right now there is no retry logic with load balanced `RestTemplates` so the existing behavior never retries.  Unlike with the Netflix `RestClient` this will retry on any type of error.

spring-cloud-commons/pom.xml
```diff
@@ -54,6 +54,14 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-aop</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.retry</groupId>
+			<artifactId>spring-retry</artifactId>
```
spencergibb: Should these be optional?

ryanjbaxter: All set

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerAutoConfiguration.java
```diff
@@ -25,9 +25,11 @@
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.http.client.ClientHttpRequestInterceptor;
+import org.springframework.retry.support.RetryTemplate;
```
spencergibb: I guess when I asked if the dependencies should be optional, that implied that this should work without retry. Either it should work without retry or the depencencies shouldn't be optional. @dsyer?

dsyer: I like optional.

ryanjbaxter: Maybe it is better to just move the load balancer code to another module?

dsyer: Why?

ryanjbaxter: Not sure I understand how you would remove the imports for spring retry and still add the functionality to retry the request.  

If we moved the load balancer code to its own module in commons than that module can depend on spring retry.  If downstream user's of commons dont need the load balancer stuff than they would not get the spring retry dependency.

ryanjbaxter: @spencergibb @dsyer with the latest changes spring-retry should be truly optional

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalanceChooser.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.springframework.cloud.client.ServiceInstance;
+
+/**
+ * Implemented by classes which use a load balancer to choose a server to
+ * send a request to.
+ *
+ * @author Ryan Baxter
+ */
+public interface LoadBalanceChooser {
```
spencergibb: Maybe a better name? `ServiceInstanceChooser`?

spencergibb: Actually, why the separation? Something not implemented yet? I don't see it in s-c-netflix.

ryanjbaxter: `LoadBalancerClient` extends this interface.  `RibbonLoadBalancerClient` in s-c-netflix implements `LoadBalancerClient`.

ryanjbaxter: I am fine with renaming it to `ServiceInstanceChooser`.

spencergibb: cool, I'd say that's the last thing.

ryanjbaxter:
@spencergibb since spring-cloud/spring-cloud-netflix#1375 is related I will wait for your approval there as well before merging this.  Thanks!

PR #148, opened 2016-11-23
Allow lifecycle endpoints to be globally disabled

drei01:
Currently, turning off lifecycle endpoints does not work with the `endpoints.enabled` property. This means the lifecycle endpoints have to be individually disabled 

```
endpoints:
   enabled: false
pause:
   enabled: false
restart:
   enabled: false
resume:
   enabled: false
```

This change allows the lifecycle endpoints observe the `endpoints.enabled` property


spencergibb:
I'm not sure that's a good idea adding the same `@ConfigurationProperties` with the same prefix. @dsyer or @philwebb how does boot handle this?

philwebb:
@spencergibb I found out the other day we do have a few objects that bind properties with the same prefix. See https://github.com/spring-projects/spring-boot/issues/7875 for background. I don't think it's a great idea but I haven't found a better way to solve the problem yet.

PR #164, opened 2017-01-05
Customize load balanced requests according to the chosen ServiceInstance

### LINKED ISSUE #162: Allow the Ribbon request to be customized according to the chosen ServiceInstance
I want to be able to use the [Surgical Routing](https://docs.cloudfoundry.org/devguide/deploy-apps/routes-domains.html#surgical-routing) CF Router feature so that Ribbon can target specific instances of an app when they're behind the Go Router. In this case, the CF App GUID and Instance Index would need to be in the ServiceInstance's metadata.

There doesn't seem to be an easy way to add to the behaviour of transforming a given request depending on the chosen ServiceInstance. It looks like it would need to be done in [LoadBalancerInterceptor](https://github.com/spring-cloud/spring-cloud-commons/blob/v1.1.6.RELEASE/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerInterceptor.java#L50-L51) and likewise in the RetryLoadBalancerInterceptor. 

Ideally, an application would be able to define its own beans of some interface that would encapsulate the transformations to be applied, and autoconfiguration would pick those up and hook them into the above spots.

william-tran:
Applications can define their own LoadBalancerRequestTransformer beans
which can modify the HttpRequest to be executed. These beans can be
@Ordered in case of multiple transformers.

Fixes #162

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerAutoConfiguration.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2015 the original author or authors.
+ * Copyright 2013-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,7 +26,6 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingClass;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -39,6 +38,7 @@ import org.springframework.web.client.RestTemplate;
  *
  * @author Spencer Gibb
  * @author Dave Syer
+ * @author Will Tran
  */
 @Configuration
 @ConditionalOnClass(RestTemplate.class)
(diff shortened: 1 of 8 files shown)


william-tran:
There are public signature changes that I'm not sure we need to be concerned with:

- Constructors for LoadBalancerInterceptor and RetryLoadBalancerInterceptor now require a LoadBalancerRequestFactory
- ServiceRequestWrapper is now UriReconstructingRequestWrapper

If we need to be concerned with backwards compatibility for anyone whose directly using these classes outside of what gets autoconfigured, we could be additive and create a default LoadBalancerRequestFactory in the existing constructor.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerAutoConfiguration.java
```diff
@@ -65,12 +67,35 @@ public void afterSingletonsInstantiated() {
...
 	}
 
+	@Bean
+	public LoadBalancerRequestTransformer uriReconstructingRequestTransformer() {
+		return new LoadBalancerRequestTransformer() {
+
+			@Override
+			public HttpRequest transformRequest(HttpRequest request, ServiceInstance instance,
+					LoadBalancerClient loadBalancer) {
+				return new UriReconstructingRequestWrapper(request, instance, loadBalancer);
```
ryanjbaxter: Why is this a transformer?  I would think Spring Cloud would create the initial request and then pass that request to any transformers that might be present.  Its not really transforming the request at all.

william-tran: I'm not sure if your issue is with nomenclature or structure but I'll try to address both.

I chose the term transformer to mean something that takes in an HttpRequest, and spits out a HttpRequest (that could be a different object).  In this case, this is the default behaviour of returning a request that has its URI reconstructed. 

This default behaviour is provided as a LoadBalancingRequestTransformer so if an application wanted to replace it entirely it could do so. See https://github.com/spring-cloud/spring-cloud-commons/pull/164/files#diff-be91ae702a070a885afe88254370b793R184

william-tran: >This default behaviour is provided as a LoadBalancingRequestTransformer so if an application wanted to replace it entirely it could do so.

Another approach would be to hard code this behaviour into LoadBalancerRequestFactory, and if you wanted to replace that behaviour, provide your own implementation of LoadBalancerRequestFactory that overrides the behaviour

PR #172, opened 2017-01-29
Add support for looking up services by metadata in DiscoveryClient

darylrobbins:
Added a new getInstances() method to DiscoveryClient, which also accepts metadata. This can be used, for example, to lookup a service with a given tag in Consul. Also added AbstractDiscoveryClient to minimize the impact of future enhancements to DiscoveryClient, so that all implementation will no longer need to be updated.

This change will require corresponding changes to all the existing implementations.

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/AbstractDiscoveryClient.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/AbstractDiscoveryClient.java
@@ -1,0 +1,19 @@
+package org.springframework.cloud.client.discovery;
+
+import org.springframework.cloud.client.ServiceInstance;
+
+import java.util.List;
+import java.util.Map;
+
+/**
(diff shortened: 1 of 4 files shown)


spencergibb:
Debating this one. You could just inject `ConsulDiscoveryClient` that has a method to query by metadata.

darylrobbins:
How about a `MetadataAwareDiscoveryClient`?

darylrobbins:
I figured this was likely to be somewhat contentious, but I really need mainstream support for this functionality to make it at all viable.

spencergibb:
Definitely don't want to force the change to other clients when they won't support it. What's the difference between a new interface here and using the consul implementation directly?

darylrobbins:
I was hoping to refer to this interface in ribbon. This would also allow adding support for other discovery providers that support some kind of metadata-based search.

spencergibb:
Each `DiscoveryClient` impl also as a Ribbon `ServerList` implementation.

spencergibb:
Closing in favor of #572

PR #179, opened 2017-02-28
GH-178: Fixes edge case of a infinite loop in decrypting properties

### LINKED ISSUE #178: EnvironmentDecryptApplicationInitializer.insert(...) infinite loop possibility
The ApplicationContext overload of insert ([currently at line 124](https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializer.java#L124)) has a while loop that looks like it could get stuck.

This probably can't happen in reality, but it's easy to fix, just move the 
```
				parent = parent.getParent();
````
one line down, **outside** the if-instance-of block:
```
		while (parent != null) {
			if (parent.getEnvironment() instanceof ConfigurableEnvironment) {
				ConfigurableEnvironment mutable = (ConfigurableEnvironment) parent
						.getEnvironment();
				insert(mutable.getPropertySources(), propertySource);
			}
			parent = parent.getParent();
		}
```

### PR DESCRIPTION
**Pull request by bijukunjummen**
Fixes #178 : Edge case where if there is a parent context which returns an environment that is not a `ConfigurableEnvironment`, can lead to an infinite loop.

### DIFF
--- a/spring-cloud-context/src/test/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializerTests.java
+++ b/spring-cloud-context/src/test/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializerTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2014 the original author or authors.
+ * Copyright 2013-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,26 +15,30 @@
  */
 package org.springframework.cloud.bootstrap.encrypt;
 
-import java.util.Collections;
-import java.util.Map;
-
 import org.junit.Test;
+import org.springframework.context.ApplicationContext;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.annotation.AnnotationConfigApplicationContext;
+import org.springframework.core.env.Environment;
 import org.springframework.core.env.MapPropertySource;
 import org.springframework.core.env.MutablePropertySources;
 import org.springframework.core.env.PropertySource;
 import org.springframework.security.crypto.encrypt.Encryptors;
 
+import java.util.Collections;
+import java.util.Map;
+
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 import static org.springframework.boot.test.util.EnvironmentTestUtils.addEnvironment;
 import static org.springframework.cloud.bootstrap.encrypt.EnvironmentDecryptApplicationInitializer.DECRYPTED_PROPERTY_SOURCE_NAME;
 
 /**
  * @author Dave Syer
- *
+ * @author Biju Kunjummen
  */
 public class EnvironmentDecryptApplicationInitializerTests {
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: spring-cloud-context/src/test/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializerTests.java [hunk 1]
```diff
@@ -32,6 +34,8 @@
 import static org.springframework.boot.test.util.EnvironmentTestUtils.addEnvironment;
 import static org.springframework.cloud.bootstrap.encrypt.EnvironmentDecryptApplicationInitializer.DECRYPTED_PROPERTY_SOURCE_NAME;
 
+import static org.mockito.Mockito.*;
```
spencergibb: No `*` imports please.

spring-cloud-context/src/test/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializerTests.java
```diff
@@ -115,4 +119,88 @@ public void indexedPropertiesCopied() {
...
+	public void testDecryptNonStandardParent() {
+		ConfigurableApplicationContext ctx = new AnnotationConfigApplicationContext();
+		EnvironmentDecryptApplicationInitializer initializer =
+				new EnvironmentDecryptApplicationInitializer(Encryptors.noOpText());
+
+		addEnvironment(ctx, "key:{cipher}value");
+
+
+		ApplicationContext ctxParent = mock(ApplicationContext.class);
+		when(ctxParent.getEnvironment()).thenReturn(new Environment() {
```
spencergibb: Use `MockEnvironment` here.

spring-cloud-context/src/test/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializerTests.java
bijukunjummen: Fixed, thanks

spring-cloud-context/src/test/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializerTests.java
bijukunjummen: Not sure if that will work @spencergibb, this issue is a real corner case and can only happen if the environment is not a `ConfigurableEnvironment`. `MockEnvironment` is a `ConfigurableEnvironment` so will not simulate the bug?

spencergibb: Good point, didn't realize that.

spring-cloud-context/src/test/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializerTests.java
```diff
@@ -15,23 +15,27 @@
  */
 package org.springframework.cloud.bootstrap.encrypt;
 
+import static org.hamcrest.Matchers.is;
```
spencergibb: statics to the bottom

bijukunjummen: Fixed @spencergibb

bijukunjummen:
Hi @spencergibb, I think all your review comments have now been addressed. A quick question: is this typically the right way to send in changes based on review comments, I have been committing, then squashing and pushing. Artem from the spring integration who I worked with last week has suggested not squashing on submitting updates for a PR.

spencergibb:
@bijukunjummen doesn't matter much. We can squash easily too. You can see individual changes if you don't squash. So there is more context that way.

PR #181, opened 2017-02-28
GH-151: HostInfo is now extensible

### LINKED ISSUE #151: Better mock-ability for InetUtils.HostInfo
The `InetUtils.HostInfo` class is both `final` and has a package-private constructor.  This makes it impossible to use for JUnit testing.  

Most Spring code I've dealt with in the past does not use final classes, so I'm hoping the design philosophy has not changed.

bijukunjummen:
Fixes #151 : HostInfo now has public constructors and is not marked as final


bijukunjummen:
Not sure if there anything more to it :-)

PR #182, opened 2017-02-28
GH-184: To support eagerly creating the set of child contexts

bijukunjummen:
Fixes: #184 
Downstream Issue: https://github.com/spring-cloud/spring-cloud-netflix/issues/1334

* Provides a base method for eagerly instantiating all the contexts registered in `NamedContextFactory`.  The expectation is that `SpringClientFactory` in spring-cloud-netflix project will invoke this to instantiate all ribbon clients on application startup.

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/context/named/NamedContextFactory.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/context/named/NamedContextFactory.java
@@ -67,6 +67,12 @@ public abstract class NamedContextFactory<C extends NamedContextFactory.Specific
 		return new HashSet<>(contexts.keySet());
 	}
 
+	protected void createAndCacheContexts() {
+		for (String key: this.configurations.keySet()) {
+			getContext(key);
+		}
+	}
+
 	@Override
 	public void destroy() {
 		Collection<AnnotationConfigApplicationContext> values = this.contexts.values();
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/context/named/NamedContextFactory.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/context/named/NamedContextFactory.java
@@ -11,7 +11,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanFactoryUtils;
 import org.springframework.beans.factory.DisposableBean;
-import org.springframework.boot.autoconfigure.PropertyPlaceholderAutoConfiguration;
+import org.springframework.boot.autoconfigure.context.PropertyPlaceholderAutoConfiguration;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.annotation.AnnotationConfigApplicationContext;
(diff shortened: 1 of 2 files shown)


spring-cloud-context/src/main/java/org/springframework/cloud/context/named/NamedContextFactory.java
```diff
@@ -67,6 +67,12 @@ public void setConfigurations(List<C> configurations) {
 		return new HashSet<>(contexts.keySet());
 	}
 
+	protected void createAndCacheContexts() {
```
ryanjbaxter: Something would need to call this method, right?  When would it be called?

bijukunjummen: Yes, still not thought of that @ryanjbaxter - it would have to reside in `spring-cloud-netflix` as it is to support this downstream issue https://github.com/spring-cloud/spring-cloud-netflix/issues/1334. I am guessing some kind of listener that calls this overridden method of `SpringClientFactory`

bijukunjummen: Also, a question @ryanjbaxter, how are such changes which is to fix a downstream issue handled, is it better to open an issue in `spring-cloud-commons` too that shadows the downstream issue and open a PR against that?

spencergibb: For ribbon there would have to be either a caveat that only `@RibbonClient` could be eagerly cached, or provide configuration to enumerate all clients that are to be eagerly created. Feign wouldn't have this problem

ryanjbaxter: I would create an issue in s-c-commons for this that would reference the one in netflix.

bijukunjummen: Sounds good @ryanjbaxter, @spencergibb , let me open an issue here to track downstream spring-cloud-netflix issue, will clean up this PR.

spring-cloud-context/src/test/java/org/springframework/cloud/context/named/NamedContextFactoryTests.java
```diff
@@ -62,8 +61,38 @@ public void testChildContexts() {
 		assertThat("bar context wasn't closed", barContext.isActive(), is(false));
 	}
 
+	@Test
+	public void testEagerlyCreateContexts() {
+		long time1 = lazyGetFromContext();
+		long time2 = cacheAndGetBeanFromContext();
+		assertThat(time2, lessThan(time1));
```
spencergibb: I worry that this is not a deterministic test. A more valid test would check when creation was done. Not sure how you'd set that up though.

bijukunjummen: Yes, good point - let me revisit the test.

bijukunjummen:
On second thoughts, this won't really work for the requirements of spring-cloud/spring-cloud-netflix#1334 where the client name may not even be available ahead of time in the `configurations` field. Will try and send a PR directly in `spring-cloud-netflix` that does it a little differently.

PR #194, opened 2017-03-30
Set spring.cloud.loadbalancer.retry.enabled to true by default

### LINKED ISSUE #193: Make RestTemplate retry opt out.
We have recently moved to Camden.SR6 and converted our code over to use the new retry functionality in Camden (We were using the deprecated Netflix Http client). One thing that tripped us up initially was that the rest client still requires us to "opt in" to the retry logic via the property `spring.cloud.loadbalancer.retry.enabled=true`. It would be helpful to either change this, such that this property is defaulted true (and you need to opt-out if you don't want retry) or to add something to the documentation that clarifies that this property must be set to enable retry within the RestTemplate. 

I think it would make more sense to default this to true to keep it consistent with Zuul, which has a property to opt-out of retry logic. 

Everything is working great, once we figured out we needed to enable the new functionality.

ryanjbaxter:
Fixes #193

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-commons.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-commons.adoc
@@ -363,7 +363,9 @@ for details of how the `RestTemplate` is set up.
 
 A load balanced `RestTemplate` can be configured to retry failed requests.
 By default this logic is disabled, you can enable it by adding link:https://github.com/spring-projects/spring-retry[Spring Retry] to your application's classpath.  The load balanced `RestTemplate` will
-honor some of the Ribbon configuration values related to retrying failed requests.
+honor some of the Ribbon configuration values related to retrying failed requests.  If
+you would like to disable the retry logic with Spring Retry on the classpath
+you can set `spring.cloud.loadbalancer.retry.enabled=false`.
 The properties you can use are `client.ribbon.MaxAutoRetries`,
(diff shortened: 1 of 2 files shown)


docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -362,8 +362,7 @@ for details of how the `RestTemplate` is set up.
 ==== Retrying Failed Requests
 
 A load balanced `RestTemplate` can be configured to retry failed requests.
-By default this logic is disabled, you can enable it by setting
-`spring.cloud.loadbalancer.retry.enabled=true` and adding link:https://github.com/spring-projects/spring-retry[Spring Retry] to your application's classpath.  The load balanced `RestTemplate` will
+By default this logic is disabled, you can enable it by adding link:https://github.com/spring-projects/spring-retry[Spring Retry] to your application's classpath.  The load balanced `RestTemplate` will
```
spencergibb: Add "To disable set `spring.cloud.loadbalancer.retry.enabled=false`." or something similar.

ryanjbaxter: I thought of that but then I didnt really see the point, why would they include spring retry and then set `spring.cloud.loadbalancer.retry.enabled=false`?

spencergibb: If the option is there, shouldn't it be documented? We have people include eureka client but turn it off for testing, for example.

PR #197, opened 2017-04-04
Add retryable status code check

ryanjbaxter:
Part of the fix to address spring-cloud/spring-cloud-netflix#1540


ryanjbaxter:
Is this OK to merge back to 1.1.x?  It is an interface change, so could possibly break some folks...

spencergibb:
I hesitate to introduce a breaking change in a SR.

ryanjbaxter:
So should we just make this feature available in Dalston then?  I am fine with that

spencergibb:
I think so

Issue #140, opened 2016-10-19
Support composite DiscoveryClient


spencergibb:

JagmohanSharma:
We have implemented this to have composite DiscoveryClient(Consul discovery and ConfigPropertyBasedDiscovery) , Can I make any contribution if that is good to have?

spencergibb:
PR's are welcome

PR #198, opened 2017-04-10
GH-140: Support for Composite Discovery client

bijukunjummen:
#140 - Added support for a composite Discovery Client

### DIFF
--- a/spring-cloud-commons/src/test/java/org/springframework/cloud/client/discovery/simple/DiscoveryClientAutoConfigurationDefaultTests.java
+++ b/spring-cloud-commons/src/test/java/org/springframework/cloud/client/discovery/simple/DiscoveryClientAutoConfigurationDefaultTests.java
@@ -10,12 +10,11 @@ import org.springframework.cloud.client.discovery.composite.CompositeDiscoveryCl
 import org.springframework.context.annotation.Configuration;
 import org.springframework.test.context.junit4.SpringRunner;
 
-import java.util.List;
-
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
- * DiscoveryClient implementation defaults to {@link SimpleDiscoveryClient}
+ * DiscoveryClient implementation defaults to {@link CompositeDiscoveryClient}
+ * 
  * @author Biju Kunjummen
  */
 @RunWith(SpringRunner.class)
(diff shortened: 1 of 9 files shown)


spring-cloud-commons/src/test/java/org/springframework/cloud/client/discovery/simple/DiscoveryClientAutoConfigurationDefaultTests.java
```diff
@@ -3,36 +3,40 @@
...
  * @author Biju Kunjummen
  */
 @RunWith(SpringRunner.class)
-@SpringBootTest(classes = DiscoveryClientAutoConfigurationDefaultTests.App.class)
+@SpringBootTest(classes = DiscoveryClientAutoConfigurationDefaultTests.Config.class)
 public class DiscoveryClientAutoConfigurationDefaultTests {
 
 	@Autowired
-	DiscoveryClient discoveryClient;
+	private List<DiscoveryClient> discoveryClients;
```
spencergibb: I think this test should still work as is.

spencergibb:
Apps need to just be able to inject `DiscoveryClient` without any changes.

spring-cloud-commons/src/test/java/org/springframework/cloud/client/discovery/simple/DiscoveryClientAutoConfigurationDefaultTests.java
bijukunjummen: Yes, it does @spencergibb, I have marked the `CompositeDiscoveryClient` as primary and auto-configuring the `SimpleDiscoveryClient`, a different test captures that behavior. I will restore this test to show that the defaults are maintained

bijukunjummen:
Anything else here @spencergibb, not sure if the assumptions are right - now there are two DiscoveryClients created by default, the `CompositeDiscoveryClient` and the `SimpleDiscoveryClient`(if client defined bean does not exist). CompositeDiscoveryClient is marked `@Primary` so that it is the one injected when a client requests it.

spencergibb:
how about a rebase @bijukunjummen?

bijukunjummen:
Will do @spencergibb - will revisit the implementation once more, rebase and update the PR.

bijukunjummen:
Done @spencergibb, rebased now.

HelloDhero:
Could you help me modify this code to make the getInstances method collect all ServiceInstance objects from every DiscoveryClient?"

PR #208, opened 2017-05-14
Disables endpoint creation

### LINKED ISSUE #117: Endpoints instances created even though they are disabled
Hi,

I'll try to disable /pause, /resume and /restart endpoint but always instanced .

endpoints.pause.enabled=false
endpoints.restart.enabled=false
endpoints.resume.enabled=false

If you disable this endpoints, is necessary load this endpoints(Objects) at memory?

It's only for improve the performance at startup.

I opened the same issue in other repository (my mistake) I attach the link : [https://github.com/spring-cloud/spring-cloud-config/issues/420](url)

@spencergibb said me that it's better if I'll open the issue here.

Thank you for your time.

matosf:
Fixes #117

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/RefreshEndpointAutoConfiguration.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/RefreshEndpointAutoConfiguration.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.autoconfigure;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.actuate.autoconfigure.ConditionalOnEnabledHealthIndicator;
 import org.springframework.boot.actuate.autoconfigure.EndpointAutoConfiguration;
+import org.springframework.boot.actuate.condition.ConditionalOnEnabledEndpoint;
 import org.springframework.boot.actuate.endpoint.Endpoint;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
@@ -56,16 +57,16 @@ public class RefreshEndpointAutoConfiguration {
 		return new RefreshScopeHealthIndicator(scope, rebinder);
 	}
 
-	@ConditionalOnProperty(value = "endpoints.restart.enabled", matchIfMissing = true)
-	protected static class RestartEndpointConfiguration {
+	@ConditionalOnClass(IntegrationMBeanExporter.class)
+	@ConditionalOnEnabledEndpoint(value = "restart", enabledByDefault = false)
+	protected static class RestartEndpointWithIntegration {
 
 		@Autowired(required = false)
 		private IntegrationMBeanExporter exporter;
 
-		@Bean("restartEndpoint")
+		@Bean
 		@ConditionalOnMissingBean
-		@ConditionalOnClass(IntegrationMBeanExporter.class)
-		public RestartEndpoint restartEndpointWithIntegration() {
+		public RestartEndpoint restartEndpoint() {
 			RestartEndpoint endpoint = new RestartEndpoint();
 			if (this.exporter != null) {
 				endpoint.setIntegrationMBeanExporter(this.exporter);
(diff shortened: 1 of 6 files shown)


pivotal-issuemaster:
@matosf Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-commons&pullRequestId=208)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-commons&pullRequestId=208) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@matosf Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-commons&pullRequestId=208)!

dsyer:
LGTM

ryanjbaxter:
Have you looked at the build failures?

spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/RefreshEndpointAutoConfiguration.java
```diff
@@ -56,45 +56,45 @@ RefreshScopeHealthIndicator refreshScopeHealthIndicator(RefreshScope scope,
 		return new RefreshScopeHealthIndicator(scope, rebinder);
 	}
 
-	@ConditionalOnClass(IntegrationMBeanExporter.class)
-	protected static class RestartEndpointWithIntegration {
+	@ConditionalOnProperty(value = "endpoints.restart.enabled", matchIfMissing = true)
```
spencergibb: This should be `ConditionalOnEnabledEndpoint`

spring-cloud-context/src/main/java/org/springframework/cloud/context/restart/RestartEndpoint.java
```diff
@@ -114,7 +114,7 @@ public ResumeEndpoint getResumeEndpoint() {
 		return new ResumeEndpoint();
 	}
 
-	@ConfigurationProperties("endpoints")
+	@ConfigurationProperties("endpoints.pause")
```
spencergibb: This line was added to make sure `endpoints.enabled` still worked. Is that still the case?

spring-cloud-context/src/test/java/org/springframework/cloud/autoconfigure/RefreshAutoConfigurationClassPathTests.java
```diff
@@ -24,14 +24,14 @@ public void refreshEventListenerCreated() {
...
+			assertFalse(context.containsBean("refreshEndpoint"));
 		}
 	}
 
 	private static ConfigurableApplicationContext getApplicationContext(
 			Class<?> configuration, String... properties) {
-		return new SpringApplicationBuilder(configuration).web(false)
-				.properties(properties).run();
+		return new SpringApplicationBuilder(configuration)
+				.web(false).registerShutdownHook(true).properties(properties).run();
```
spencergibb: Why add `.registerShutdownHook(true)`?

spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/RefreshEndpointAutoConfiguration.java
matosf: OK, will change it, including the other places where @ConditionalOnProperty was already used.

spring-cloud-context/src/main/java/org/springframework/cloud/context/restart/RestartEndpoint.java
matosf: I know, but this IMO not the best solution, since I would like to also be able to disable the pause endpoint independently, on top of globally. The same goes for the resume endpoint.

The global endpoints.enable logs is now handled byt @ConditionalOnProperty in RefreshEndpointAutoConfiguration.

spring-cloud-context/src/test/java/org/springframework/cloud/autoconfigure/RefreshAutoConfigurationClassPathTests.java
matosf: I was trying to fix the exact exceptions shown by CircleCI failing tests. I worked locally, but I see now that was a coincidence. Will undo the change.

Regarding the exceptions I cannot find the relation with my changes, and seems similar to:
https://github.com/spring-cloud/spring-cloud-config/issues/118

matosf:
@spencergibb can you please have another look?

spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/RefreshEndpointAutoConfiguration.java
```diff
@@ -45,6 +45,7 @@
  */
 @Configuration
 @ConditionalOnClass(Endpoint.class)
+@ConditionalOnProperty(value = "endpoints.enabled", matchIfMissing = true)
```
spencergibb: I don't think this is required. See https://github.com/spring-projects/spring-boot/blob/master/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/condition/OnEnabledEndpointCondition.java#L67-L75

`@ConditionalOnEnabledEndpoint` should take it into account.

spencergibb:
Ping @matosf

matosf:
@spencergibb PR updated.

Issue #204, opened 2017-05-01
Adjusting `ServiceInstance.Metadata`


smcvb:
Would it be possible to have some form of `put()` or `set()` function for the `Map<String,String> metadata` in the `ServiceInstance`?
I'd like to be able to **not** rely on a specific Spring Cloud implementation for adjusting the `metadata` (like through Eureka's properties, or the tags in Consul), but just use the api in spring-cloud-commons to update a `ServiceIntance` it's `metadata`.

This could for example give the possibility for other frameworks to set framework-specific knowledge in the `ServiceInstance.metadata` field and not having to know anything of specific implementation. Thus, only needing the `spring-cloud-commons` dependency.

ryanjbaxter:
It doesn't seem unreasonable to me.  @spencergibb what do you think?

spencergibb:
It sounds reasonable, getting something that works across impls might prove tricky.

ryanjbaxter:
@smcvb want to submit a PR?

smcvb:
@ryanjbaxter sure thing! 
Just a change to the `ServiceInstance` interface to add a `setMetadata(Map)` and `putMetadata(key, value)` I'd guess?

ryanjbaxter:
That seems OK.  Just make sure you also update the implementations as well.

smcvb:
@ryanjbaxter did you also mean for me to adjust the specific Spring Cloud `ServiceInstance` implementations by Eureka/Consul/etc, or just the implementations in `spring-cloud-commons`?

smcvb:
Ok, I'll see what time I can free up to help with that :+1:

dsyer:
I don't really like `putMetaData()` much, since all it does is mirror the call to the underlying map. We might as well just document that the map is mutable and make sure that all interfaces support that somehow (or break in an informative way).

spencergibb:
@dsyer, not all impls have a simple map backing them (consul in particular).

dsyer:
I guess that’s why I said “break in an informative way”. But if it can support the putValue method, I’d say it can support a mutable backing map.

spencergibb:
After looking at consul, `ServiceInstance` is not used by consul during registration. Updating metadata there means nothing. It happens to work in eureka, since the backing object happens to be one eureka uses and sends deltas during heartbeats.

[...]

PR #209, opened 2017-05-16
gh-204 Introduce putMetadata() and setMetadata() function to ServiceInstance

smcvb:
This PR is to solve #204, and introduces a `void putMetadata(String key, String value)` and `void setMetadata(Map<String, String> metadata)` function to the `ServiceInstance` interface.

This addition might prove interesting for other frameworks, which want to rely on `spring-cloud-commons` it's `DiscoveryClient`/`ServiceInstance` set up to do service discovery, but also want to adjust the `ServiceInstance.metadata` without the need of knowing which Spring Cloud implementation is used.


spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
```diff
@@ -57,6 +59,23 @@ public URI getUri() {
 		return this.metadata;
 	}
 
+	@Override
+	public void putMetadata(String key, String value) {
+		this.metadata.put(key, value);
+	}
+
+	@Override
+	public void setMetadata(Map<String, String> metadata) {
```
smcvb: I took the approach of removing all the old fields and adding them because `metadata` is `final` for this class.

Removing `final` from the field seemed more problematic because of the `@RequiredArgsConstructor`.

spencergibb:
Wouldn't you be able to do all of this, assuming the impl returned a read/write version of the metadata?

smcvb:
@spencergibb yes I'd say so. That was my initial assumption of the `getMetadata()` function too.
But after trying out the Eureka and Consul implementation of the `DiscoveryClient / ServiceInstance` I found out this did hold for Eureka, but didn't for Consul.

Hence the idea to add a `set()/put()` function to the interface.

smcvb:
Updated this PR to point to align with the current master branch. 
@spencergibb am I still required to provide specific `ServiceInstance` implementations for all specific version like Consul/Eureka?

spencergibb:
Since we're on Java 8, create empty default methods on the interface

smcvb:
Added empty default methods per suggestion.

smcvb:
@spencergibb is this PR still being hold back by something you'd like me to do? 
Or is @dsyer his comment on #204 such that we should debate whether this should still go in or not?

spencergibb:
Closing, if we were to support something like this, it would be with a mutable backing map, but I have my doubts even for that https://github.com/spring-cloud/spring-cloud-commons/issues/204#issuecomment-374969458

PR #221, opened 2017-06-26
Add option to control management registration

jihor:
In continuation of [this PR to spring-cloud-consul](https://github.com/spring-cloud/spring-cloud-consul/pull/323), I've added option to control management registration to spring-cloud-commons.


ryanjbaxter:
We are trying to move away from Lombok, can you please remove the annotations?

jihor:
Done.

[spring-cloud-netflix page](https://github.com/spring-cloud/spring-cloud-netflix) stated that is uses Lombok, so I took it as a sign that using Lombok was also ok for other spring-cloud projects. But I'm curious why you are moving away from Lombok.

I should also explain that this PR is born from [this one](https://github.com/spring-cloud/spring-cloud-consul/pull/323) at spring-cloud-consul. The idea is to control whether the management should be registered as a separate service, and @spencergibb suggested that this property should go to commons library.

ryanjbaxter:
Thanks.  Yes we use Lombok but we have been slowly removing it from the code base.

PR #222, opened 2017-06-28
Implementation of factories to provide Apache HTTP Clients

ryanjbaxter:
Adds factory beans for downstream projects to consistently create Apache HTTP Clients


spring-cloud-commons/src/main/java/org/springframework/cloud/commons/httpclient/DefaultOkHttpClientFactory.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Default implementation of {@link OkHttpClientFactory}.
+ * @author Ryan Baxter
+ */
+public class DefaultOkHttpClientFactory implements OkHttpClientFactory {
+
+	private static final Log LOG = LogFactory.getLog(DefaultOkHttpClientFactory.class);
+
+	@Override
+	public OkHttpClient create(boolean disableSslValidation, long connectTimeout,
```
spencergibb: I'm a little worried about such a long list of parameters. I wonder if something like a builder customizer might make more sense?

ryanjbaxter:
@spencergibb I ended up just using the provided builders for the clients instead of reinventing the wheel.  I left the connection pools as is since the number of arguments is rather fixed.

ryanjbaxter:
@spencergibb want to take another pass at this?

ryanjbaxter:
ping @spencergibb

PR #227, opened 2017-07-08
WIP! DO NOT MERGE! Commons Load Balancer

spencergibb:
/cc @twicksell
This is a POC load balancer implementation.

The `DiscoveryClient` is the source of the instances. The only two interfaces are `ScopedLoadBalancer` and `ScopedDiscoveryClient`. Those remove the serviceId from the parameters because it is known in the environment. Load balancing strategies and algorithms can be implemented as a `ScopedLoadBalancer` (see `RandomScopedLoadBalancer`). Filtering and other things can come from a decorated `ScopedDiscoveryClient` (see `MyScopedDiscoveryClient`).

There are lots of things missing, like metrics and actuator support, but I wanted to see if this is the right direction to go.

Something like this would replace ribbon.


spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/scoped/ScopedLoadBalancer.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.springframework.cloud.client.ServiceInstance;
+
+/**
+ * Implemented by classes which use a load balancer to choose a server to
+ * send a request to.
+ *
+ * @author Ryan Baxter
+ */
+public interface ScopedLoadBalancer {
```
elandau: Have you considered doing this as a generic API that also includes quality of service?  Incorporating request lifecycle into the load balancer API would allow for more advanced load balancing strategies that take quality of service into account.  Many libraries fall short on the load balancing API by only focusing on the server selection but don't provide a feedback mechanism for more advanced load balancing algorithm.  For example, a least loaded load balancing algorithm would need to know when an operation completed so it can decrement the pending request count.  A latency aware algorithm would need to collect latency samples to update internal statistics. 

For example,

```java
public interface LoadBalancer<T> {
    enum Status {
        SUCCESSS,  // Request was handled successfully
        FAILED, // Request reached the server but failed due to timeout or internal error
        DISCARD, // Request did not go off box and should not be counted for statistics
    }

    // Context created for each request.
    interface Context {
        boolean hasServer();

        T getServer();

        // Notification that the request completed
        void complete(Status status);
    }

     // Choose the next server based on the load balancing algorithm
     Context<T> choose();
}
```

Regarding using generics, it would be much easier to test a new algorithm without having to tie it to a specific service instance API.

spencergibb: @elandau see 9d23d77ddac584f22f2744c080c68442ec0b8eb2. `complete()` seems like a good place for metrics and other operations.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/context/RoundRobinLoadBalancer.java
```diff
@@ -0,0 +1,90 @@
...
+import org.springframework.cloud.client.ServiceInstance;
+import org.springframework.cloud.client.loadbalancer.impl.scoped.ScopedDiscoveryClient;
+import org.springframework.cloud.client.loadbalancer.impl.support.LoadBalancerClientFactory;
+
+import static org.springframework.util.CollectionUtils.isEmpty;
+
+/**
+ * @author Spencer Gibb
+ */
+public class RoundRobinLoadBalancer implements LoadBalancer<ServiceInstance> {
```
ryanjbaxter: More of an observation than anything else.  This round robin implementation is dependent on the list of services returned from the discovery client.  There is no guarantee that the order of that list will be the same.  So in some cases we might not always be following a round robin algorithm.  Specifically once the service list is refreshed from the service discovery service we cant guarantee that when we go to select the next server we wont select the previous server we just tried.  This happens with Ribbon today as well.  Like I said just a comment on the way this works more than anything else.

spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/KeyProperties.java
```diff
@@ -137,15 +137,15 @@ public void setSecret(String secret) {
 	public static class Rsa {
 
 		/**
-		 * The RSA algorithm to use (DEFAULT or OEAP). Once it is set do not change it (or
```
dsyer: Typo here? Maybe search and replace error?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/CommonsLoadBalancer.java
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	@Override
+	public ServiceInstance choose(String serviceId) {
+		ScopedLoadBalancer scopedLoadBalancer = this.clientFactory.getInstance(serviceId, ScopedLoadBalancer.class);
+
+		return scopedLoadBalancer.choose();
+	}
+
+	public LoadBalancerClientFactory getClientFactory() {
```
dsyer: Isn't this an implementation detail (no public method should be exposed)?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/scoped/CachedDiscoveryClient.java
```diff
@@ -0,0 +1,31 @@
...
+
+	private final DiscoveryClient delegate;
+	private final Environment environment;
+
+	public CachedDiscoveryClient(DiscoveryClient delegate, Environment environment) {
+		this.delegate = delegate;
+		this.environment = environment;
+	}
+
+	@Cacheable(cacheNames = "discovery-client", key = "${loadbalancer.client.name}")
```
ryanjbaxter: When will this cache expire?  I read that if the parameters change than the cache wont be returned but since parameters are not changing how will it be expired?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/scoped/ScopedLoadBalancer.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Choose a ServiceInstance from the LoadBalancer for the specified service
+     * @return a ServiceInstance that matches the serviceId
+     */
+    ServiceInstance choose();
```
dsyer: I'm a bit confused about why we need 2 interfaces with the same methods `LoadBalancer` and `ScopedLoadBalancer`.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/scoped/ScopedLoadBalancer.java
elandau: Indeed.  Complete should probably just get a single stats object that contains both the status code as well as any metrics.  Having a single parameter (instead of an enum) would allow us to add more info in the future without breaking the interface.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/context/LoadBalancer.java
```diff
@@ -0,0 +1,25 @@
...
+		boolean hasServer();
+
+		T getServer();
+
+		// Notification that the request completed
+		void complete(Status status);
+	}
+
+	// Choose the next server based on the load balancing algorithm
+	Context<T> choose();
```
twicksell: Thinking this choose method should take a parameter along the lines of `choose(RequestContext<T> requestContext)`. Things I'd be looking for in the RequestContext are the resource (uri?), headers, etc.

elandau: +1

That would be a good place to also include things like deadline and affinity

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/context/RoundRobinLoadBalancer.java
twicksell: Indeed. In fact in some of our new work involving gRPC we enforce a consistent sorting on the servers coming back from eureka. Debating now on if/how we should roll that out into the OSS eureka server.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/scoped/CachedDiscoveryClient.java
spencergibb: It's configurable. I just wanted to try and use the framework caching instead of trying to do something custom. There are all sorts of options for this at the framework level.

spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/KeyProperties.java
spencergibb: haha!

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/CommonsLoadBalancer.java
spencergibb: 👍

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/scoped/ScopedLoadBalancer.java
spencergibb: One has `serviceId` as a parameter, the other doesn't, since we already know. `ScopedLoadBalancer` is more like `IRule` from ribbon. Name suggestions welcome.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/context/LoadBalancer.java
spencergibb: How would we not get HTTP specific here?

spencergibb: @marcingrzejszczak @adriancole this gets similar to baggage. How do you deal with the protocol specifics without those leaking into the API?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/context/RoundRobinLoadBalancer.java
spencergibb: Currently, the list is cached. Once the list has changed, shouldn't a load balancer like this basically reset?

spencergibb:
In `LoadBalancer` I have `ChosenContext` chose(RequestContext)` and in ChosenContext `void complete(CompletionContext).

I'd love more descriptive name suggestions.

Should there be any relationship between the three?

spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/simple/SimpleDiscoveryProperties.java
```diff
@@ -122,5 +122,18 @@ public URI getUri() {
 		public Map<String, String> getMetadata() {
 			return this.metadata;
 		}
+
+		@Override
+		public String toString() {
+			final StringBuffer sb = new StringBuffer("SimpleServiceInstance{");
```
adriancole: ps iirc javac will stringbuffer concatenated strings, so whichever reads easier

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/CommonsLoadBalancer.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.springframework.cloud.client.ServiceInstance;
+import org.springframework.cloud.client.loadbalancer.LoadBalancer;
+import org.springframework.cloud.client.loadbalancer.impl.scoped.ScopedLoadBalancer;
+import org.springframework.cloud.client.loadbalancer.impl.support.LoadBalancerClientFactory;
+
+/**
+ * @author Spencer Gibb
+ */
+public class CommonsLoadBalancer implements LoadBalancer {
```
adriancole: why Commons? Maybe Default? unless we are anticipating a tragedy of the commons where users eventually spoil the loadbalancer by using it without regards to others :)

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/annotation/LoadBalancerClient.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright 2013-2014 the original author or authors.
```
adriancole: nit: update the template?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/annotation/LoadBalancerClient.java
```diff
@@ -0,0 +1,64 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.cloud.client.loadbalancer.impl.support.LoadBalancerClientFactory;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+
+/**
+ * Declarative configuration for a ribbon client. Add this annotation to any
```
adriancole: nit: scrub ribbon

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/annotation/LoadBalancerClient.java
```diff
@@ -0,0 +1,64 @@
...
+	 */
+	String name() default "";
+
+	/**
+	 * A custom <code>@Configuration</code> for the load balancer client. Can contain override
+	 * <code>@Bean</code> definition for the pieces that make up the client.
+	 *
+	 * @see LoadBalancerClientConfiguration for the defaults
+	 */
+	Class<?>[] configuration() default {};
```
adriancole: like!

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/annotation/LoadBalancerClientConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * @author Dave Syer
+ */
+@SuppressWarnings("deprecation")
+@Configuration
+@EnableConfigurationProperties
+public class LoadBalancerClientConfiguration {
+
+	@Value("${loadbalancer.client.name}")
```
adriancole: maybe javadoc why this expression would work (ex it is scoped or there's only one client expected etc?)

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/annotation/LoadBalancerClientConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+	private String name = "client";
+
+	@Bean
+	@ConditionalOnMissingBean
+	public CachedDiscoveryClient cachingDiscoveryClient(DiscoveryClient discoveryClient, Environment env) {
```
adriancole: nit: match bean names w/impl names?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/annotation/LoadBalancerClientConfigurationRegistrar.java
```diff
@@ -0,0 +1,88 @@
...
+		}
+		Map<String, Object> client = metadata.getAnnotationAttributes(
+				LoadBalancerClient.class.getName(), true);
+		String name = getClientName(client);
+		if (name != null) {
+			registerClientConfiguration(registry, name, client.get("configuration"));
+		}
+	}
+
+	private String getClientName(Map<String, Object> client) {
```
adriancole: nit: I think these methods here and below can be static?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/annotation/LoadBalancerClientSpecification.java
```diff
@@ -0,0 +1,79 @@
...
+
+	private String name;
+
+	private Class<?>[] configuration;
+
+	public LoadBalancerClientSpecification() {
+	}
+
+	public LoadBalancerClientSpecification(String name, Class<?>[] configuration) {
+		this.name = name;
```
adriancole: feels we should check null here and on the setters?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/annotation/LoadBalancerClients.java
```diff
@@ -0,0 +1,45 @@
...
+@Configuration
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE })
+@Documented
+@Import(LoadBalancerClientConfigurationRegistrar.class)
+public @interface LoadBalancerClients {
+
+	LoadBalancerClient[] value() default {};
+
+	Class<?>[] defaultConfiguration() default {};
```
adriancole: maybe a see javadoc tag for LoadBalancerClient.configuration if that's what's implied

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/context/LoadBalancer.java
```diff
@@ -0,0 +1,50 @@
+package org.springframework.cloud.client.loadbalancer.impl.context;
+
+/**
+ * @author Spencer Gibb
+ */
+public interface LoadBalancer<T> {
+	enum Status {
+		SUCCESSS,  // Request was handled successfully
+		FAILED, // Request reached the server but failed due to timeout or internal error
+		DISCARD, // Request did not go off box and should not be counted for statistics
```
adriancole: switch comments to javadoc. can discard mean served from cache? note: typo on sucesss

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/context/LoadBalancer.java
```diff
@@ -0,0 +1,50 @@
...
+	interface ChosenContext<T> {
+		boolean hasServer();
+
+		T getServer();
+
+		// Notification that the request completed
+		void complete(CompletionContext completionContext);
+	}
+
+	interface RequestContext {
```
adriancole: javadoc what this might contain?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/context/LoadBalancer.java
```diff
@@ -0,0 +1,50 @@
...
+		SUCCESSS,  // Request was handled successfully
+		FAILED, // Request reached the server but failed due to timeout or internal error
+		DISCARD, // Request did not go off box and should not be counted for statistics
+	}
+
+	//TODO: add metrics
+	class CompletionContext {
+		final Status status;
+
+		public CompletionContext(Status status) {
```
adriancole: should this include a nullable throwable (ex a normal bug could abend and might be helpful)

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/context/LoadBalancer.java
```diff
@@ -0,0 +1,50 @@
...
+			sb.append('}');
+			return sb.toString();
+		}
+	}
+
+	// ChosenContext created for each request.
+	interface ChosenContext<T> {
+		boolean hasServer();
+
+		T getServer();
```
adriancole: never null? of if !hasServer this would be null. Is there a race between hasServer and getServer?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/context/RoundRobinLoadBalancer.java
adriancole: could this be solved by supplying the previous completion context with the request? (or some similar token?) Internally, it could peek at details for how to continue the list.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/context/RoundRobinLoadBalancer.java
```diff
@@ -0,0 +1,90 @@
...
+		}
+
+		int nextServerIndex = incrementAndGetModulo(instances.size());
+
+		ServiceInstance instance = instances.get(nextServerIndex);
+
+		return new DefaultChosenContext(instance);
+	}
+
+	public static class DefaultChosenContext implements ChosenContext<ServiceInstance> {
```
adriancole: can things like this still operate package private? if we aren't exposing new methods, I'd prefer them hidden

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/scoped/CachedDiscoveryClient.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.cloud.client.ServiceInstance;
+import org.springframework.cloud.client.discovery.DiscoveryClient;
+import org.springframework.core.env.Environment;
+
+import java.util.List;
+
+/**
+ * @author Spencer Gibb
+ */
+public class CachedDiscoveryClient implements ScopedDiscoveryClient {
```
adriancole: javadoc the impact of cached things.. for example, I'm guessing this means the instance list will be available when the backend is down, but might be stale. Is there anything in the backend to invalidate cache entries? If so, note behaviors folks can rely on. otherwise, don't

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/scoped/RoundRobinScopedLoadBalancer.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.cloud.client.ServiceInstance;
+import org.springframework.cloud.client.loadbalancer.impl.support.LoadBalancerClientFactory;
+
+import static org.springframework.util.CollectionUtils.isEmpty;
+
+/**
+ * @author Spencer Gibb
+ */
+public class RoundRobinScopedLoadBalancer implements ScopedLoadBalancer {
```
adriancole: looks like c&p for some of these things.. if meaty code, maybe an internal utility that they share. otherwise, keep an eye on it.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/scoped/ScopedDiscoveryClient.java
```diff
@@ -0,0 +1,12 @@
...
+
+import org.springframework.cloud.client.ServiceInstance;
+
+import java.util.List;
+
+/**
+ * @author Spencer Gibb
+ */
+public interface ScopedDiscoveryClient {
+	List<ServiceInstance> getInstances();
```
adriancole: should the caller anticipate runtime exceptions or are they protected from them?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/scoped/ScopedLoadBalancer.java
adriancole: maybe also consider which types in this PR could be served as abstract classes. we can easily add methods to abstract classes, but adding methods to interfaces revlocks those using extensions

spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/simple/SimpleDiscoveryProperties.java
spencergibb: Auto-gen by intelij :-)

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/CommonsLoadBalancer.java
spencergibb: Lol. AFAIC names at this point are very fluid.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/annotation/LoadBalancerClient.java
spencergibb: Copy past I'm sure.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/annotation/LoadBalancerClient.java
spencergibb: yup, copy/paste

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/context/LoadBalancer.java
spencergibb: Yup, got to figure that out :-)

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/context/RoundRobinLoadBalancer.java
ryanjbaxter: @spencergibb how would the cache know the list changed without fist fetching the list from the service?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/scoped/CachedDiscoveryClient.java
ryanjbaxter: OK I just dove a little deeper on how it works.  By default it will use JCache right?  But apps could plug in their own caching backend if they want?

spencergibb: I think the ultimate default is a concurrent hash map, but yes, it's customizable.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/client/loadbalancer/impl/context/LoadBalancer.java
spencergibb: @elandau thoughts?

PR #229, opened 2017-07-11
Fix empty keys in EnvironmentChangedEvent on resetting the Environment

joshiste:
When the Environment is resetted the EnvironmentChangedEvent always
contains an empty keySet, since the live view of the cleared map was
used. This commit fixes this.


spencergibb:
I guess it depends on what `reset()` should do. The way you define it, is all keys that are in the map that are cleared get sent as part of an `EnvironmentChangedEvent`. The previous behavior was empty. What is the use case you have that doesn't fit?

/cc @dsyer

joshiste:
I was debugging another issue where my ConfigurationProperties hasn't been rebound after a reset (using Finchley) and stumbled across that code. In my opinion the code expresses an intention but doesn't behaves this way -  so I "fixed" this.

PR #231, opened 2017-07-20
Remove dependence on lombok

bijukunjummen:
Removed dependence on lombok library and changed code to account for the lombok generated methods/fields.


bijukunjummen:
Sorry, went for a low hanging fruit @spencergibb, I like the fact that lombok dependence is being removed from spring-cloud-netflix and thought of doing the same here.

spencergibb:
No worries. We like help with low hanging fruit.

PR #232, opened 2017-07-22
Show better message for missing config keystore props

### LINKED ISSUE #230: Improve error handling if location of key store in encryption configuration is wrong
If the encryption of configurations is setup like this:
```properties
encrypt:
  key-store:
    location: file:/config/app/keystore.jceks
    password: xxxx
    secret: xxx
```
but the location to the key store is wrong, this results in a NullPointerException in org.springframework.cloud.context.encrypt.EncryptorFactory#create.

The reason is this logic in EncryptionBootstrapConfiguration:
```java
		public TextEncryptor textEncryptor() {
			KeyStore keyStore = this.key.getKeyStore();
			if (keyStore.getLocation() != null && keyStore.getLocation().exists()) {
				return new RsaSecretEncryptor(
						new KeyStoreKeyFactory(keyStore.getLocation(),
								keyStore.getPassword().toCharArray()).getKeyPair(
										keyStore.getAlias(),
										keyStore.getSecret().toCharArray()),
						this.key.getRsa().getAlgorithm(), this.key.getRsa().getSalt(),
						this.key.getRsa().isStrong());
			}
			return new EncryptorFactory().create(this.key.getKey());
		}
```
The code just ignores, that a keystore location is set, but the keystore file doesn't exists and continues with the assumption, that a key was set directly in the "key" property, but without checking this.

Please 
- throw a meaningful exception, if the keystore location doesn't exists,
- check if the "key" is null and throw an exception too

bijukunjummen:
Fixes #230  - Throws an `IllegalStateException` if the keystore location is present but wrong. The second requested issue of handling a case where key is null does not happen as `KeyCondition` would fail.

### DIFF
--- a/spring-cloud-context/src/test/java/org/springframework/cloud/bootstrap/encrypt/EncryptionBootstrapConfigurationTests.java
+++ b/spring-cloud-context/src/test/java/org/springframework/cloud/bootstrap/encrypt/EncryptionBootstrapConfigurationTests.java
@@ -1,13 +1,19 @@
 package org.springframework.cloud.bootstrap.encrypt;
 
-import static org.junit.Assert.assertEquals;
-
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.security.crypto.encrypt.TextEncryptor;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertEquals;
+
 public class EncryptionBootstrapConfigurationTests {
+
+	@Rule
+	public ExpectedException expected = ExpectedException.none();
 
 	@Test
 	public void rsaKeyStore() {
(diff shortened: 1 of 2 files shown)


spring-cloud-context/src/test/java/org/springframework/cloud/bootstrap/encrypt/EncryptionBootstrapConfigurationTests.java
```diff
@@ -1,14 +1,20 @@
...
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.security.crypto.encrypt.TextEncryptor;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertEquals;
+
 public class EncryptionBootstrapConfigurationTests {
 
+	@Rule
+	public ExpectedException expected = ExpectedException.none();
```
spencergibb: Unused

bijukunjummen: Fixed now @spencergibb

PR #234, opened 2017-07-31
Support resettable configuration properties

bijukunjummen:
@spencergibb , @dsyer, @ryanjbaxter - this is more a POC to support the issues that we had discussed about PR https://github.com/spring-cloud/spring-cloud-netflix/pull/2158. 

One of the underlying issue is that any `@ConfigurationProperties` annotated bean supports refreshing of new properties or update of existing properties, but if properties are deleted it is not accounted for - this is because the previously bound bean is retrieved and re-bound with the properties again and existing state remains. This PR tries to fix it by introducing a `PropertiesResettable` interface - if a `@ConfigurationProperties` annotated class implements this interface then the internal state can be reset by `reset()` method which the bean is expected to implement. 

Can you please review and see if this approach looks okay - it does behave like `@RefreshScope` though, so any feedback on how to better implement this is welcome.

If this approach works then I can as a follow up work on the PR on the s-c-netflix side to fix ZuulHandlerMapping on ZuulProperties route deletion.

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/ConfigurationPropertiesRebinderAutoConfiguration.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/ConfigurationPropertiesRebinderAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2015 the original author or authors.
+ * Copyright 2013-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


spencergibb:
I think we're looking for some changes in boot 2.0 for this.

holy12345:
Hi @spencergibb @bijukunjummen 
I have been concerned about this issue, in addition I think there is a place that can be improved.I found that when zuul is used with eureka, the triggering of zuul's dynamic refresh route is via the Eureka cacheRefreshed event.
The corresponding source code is as follows:
```
@Override
public void onApplicationEvent(ApplicationEvent event) {
    if (event instanceof InstanceRegisteredEvent) {
        reset();
    } else if (event instanceof ParentHeartbeatEvent) {
	ParentHeartbeatEvent e = (ParentHeartbeatEvent) event;
	resetIfNeeded(e.getValue());
    } else if (event instanceof HeartbeatEvent) {
	HeartbeatEvent e = (HeartbeatEvent) event;
	resetIfNeeded(e.getValue());
    }
}
```
`HeartBeatEvent` this event comes from `Eureka Client` getting registration information from `eureka server` and updating the local cache. In most cases, the local cache list information has not changed. This will cause zuul to refresh the route even if there is no change.
Whether we can use finer-grained events (in fact we expect zuul to dynamically refresh the route when the cache on the local eureka service list changes).
I did not think of a good way to optimize this problem, so I would like to consult.Any answer is a great reward for me.

spencergibb:
@holy12345 please don't comment on unrelated pull requests

holy12345:
@spencergibb Okay, I deleted it. Sorry.

dsyer:
What's wrong with `@RefreshScope`?

spencergibb:
Closing this due to inactivity. Please re-open if there's more to discuss.

PR #245, opened 2017-08-23
Remove the need for @EnableDiscoveryClient

### LINKED ISSUE #218: Make use of `@EnableDiscoveryClient` optional.
More devs are putting it on a profile enabled configuration. In all impls it just creates a marker bean and then auto configuration kicks in. I don't see why this couldn't be  just auto configuration and properties. Would blend better with sidecar proxies. @dsyer, @ryanjbaxter

ryanjbaxter:
Fixes #218


spencergibb:
Does `@EnableDiscoveryClient(autoRegister=false)` still work?

ryanjbaxter:
Just a start right now.  Need to discuss how to disable auto registration.  I think the best way might be to use a property rather than continue to support `@EnableDiscoveryClient(autoRegister=false)`.  I ran these changes against sc-netflix and removed all references to `@EnableDiscoveryClient` and `@EnableEurekaClient` and all tests passed.

ryanjbaxter:
@spencergibb you are too quick, see my comment above about the property ;)

spencergibb:
If it's going to be in Edgware we can't break it. The import selector could set a property if it was false.

ryanjbaxter:
OK let me look at that too.

ryanjbaxter:
@spencergibb I added some tests.  However when I was looking to update the documentation I realized there is already a property to disable registration, `spring.cloud.service-registry.auto-registration.enabled`.  I am guessing we should reuse this property rather than introduce another one.

spencergibb:
Yes, I agree

Issue #251, opened 2017-09-13
"disableSslValidation" means NO SSL support


michael-wirth:
DefaultApacheHttpClientConnectionManagerFactory gives only SSL support if the SSL validation is enabled. This is confusing.

That means if SSL validation is not disabled then you can't use SSL at all? Is this the expected behaviour or is this a bug? If it's the expected behaviour then the parameter name 'disableSslValidation' is misleading.

ryanjbaxter:
@michael-wirth that is not my understanding of how that code worked.  SSL validation (meaning the Apache HTTP Client would validate SSL certificates against the ones it trusts) was enabled by default.  When `disableSslValidation` is `true` the `TrustManager` that is created does not validate the certificates.

michael-wirth:
@ryanjbaxter This is a **bug** then, as the `https` protocol is not supported if `disableSslValidation` is `false`.

The `https` protocol must always be registered independent of the `disableSslValidation` flag. Same as the `http` protocol.

ryanjbaxter:
Looks like you are right, my misunderstanding how how this works.  Thanks!

qcastel:
Thanks @michael-wirth :) I was debugging the issue I had and I reach the same conclusion. Happy to see that someone already fixed it!

qcastel:
@michael-wirth Any chance to get it backported to 1.3.x? I would be interested by this fix but not sure we want to be on a milestones:

My understanding is that we need spring cloud 2.0.0-M6 and until 2.0.0 is officially released, I'm bit concerned to go in production on a version of spring cloud not released yet.

PR #252, opened 2017-09-13
Add SSL support for connection factory with SSL validation

michael-wirth:
fixing https://github.com/spring-cloud/spring-cloud-commons/issues/251


michael-wirth:
@ryanjbaxter There are already 2 new tests which check if the SSL protocol is registered and also check the related TrustManger
- newConnectionManagerWithSSL
- newConnectionManagerWithDisabledSSLValidation

Issue #211, opened 2017-05-18
ClassLoader Leak when adding spring-cloud-context dependency


ArloL:
I am currently trying to create my own version of the spring-boot-admin. Since we are still using a standalone Tomcat on our servers I am creating a war package. To discover memory leaks with regard to ClassLoader leaks we have created an integration test: https://github.com/evosec/tomcat-classloader-leak-test

Sadly the test fails for the spring-boot-admin. I could boil it down to the spring-cloud-context 1.2.0 dependency. I have created a demo project that demonstrates the issue: https://github.com/ArloL/spring-cloud-context-classloader-leak-demo

Note that [master runs successfully](https://travis-ci.org/ArloL/spring-cloud-context-classloader-leak-demo/builds/233524235) and the [spring-cloud-context branch fails](https://travis-ci.org/ArloL/spring-cloud-context-classloader-leak-demo/builds/233524881).

The only difference is the added spring-cloud-context dependency: https://github.com/ArloL/spring-cloud-context-classloader-leak-demo/compare/master...spring-cloud-context

I have tried analyzing the heap dump with VisualVM as described in <https://cdivilly.wordpress.com/2012/04/23/permgen-memory-leak>. Be aware that the query is a bit different nowadays `select x from org.apache.catalina.loader.ParallelWebappClassLoader x`. This sadly did not give any helpful results.

Then I tried adding the [classloader-leak-prevention-servlet3](https://github.com/mjiderhamn/classloader-leak-prevention). Sadly this does not change the test result: <https://travis-ci.org/ArloL/spring-cloud-context-classloader-leak-demo/builds/233563888>. There is also no difference in the heap dump.

I do realize that deployment inside a standalone Tomcat is not in the focus of a spring-*cloud* dependency but maybe there is a bug (e.g. instantiated Servlet that is not properly destroyed) that is presenting itself due to this test. Maybe someone else has an idea about what's going on here. Thanks a lot :)

mouellet:
This is a major issue for anyone who wants to use spring-cloud features for traditional deployment on servlet containers. The bootstrap context adds itself as a parent to the root `ApplicationContext` but it's never closed, it's not a web context.

You can to add a listener that will close the root web application context and its parent. This can be done by overriding the `SpringBootServletInitializer.onStartup()` method somewhat like this:
```java
servletContext.addListener(new ContextLoaderListener(rootAppContext) {
    @Override
    public void contextInitialized(ServletContextEvent event) {
[...]

PR #267, opened 2017-10-23
Ensure that the bootstrap context is closed after refresh.

spencergibb:
see https://github.com/spring-cloud/spring-cloud-zookeeper/issues/123

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/BootstrapApplicationListener.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/BootstrapApplicationListener.java
@@ -299,13 +299,17 @@ public class BootstrapApplicationListener
 		return this.order;
 	}
 
-	private static class AncestorInitializer implements
+	public static class AncestorInitializer implements
 			ApplicationContextInitializer<ConfigurableApplicationContext>, Ordered {
 
 		private ConfigurableApplicationContext parent;
 
 		public AncestorInitializer(ConfigurableApplicationContext parent) {
 			this.parent = parent;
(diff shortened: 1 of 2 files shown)


spencergibb:
@dsyer I know this is a bit of a hack, but it works. Any suggestions on how it could be done better? Should be included in Edgware.

dsyer:
I think this would be better:

```
diff --git a/spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/ContextRefresher.java b/spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/ContextRefresher.java
index c57a25b..e4979db 100644
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/ContextRefresher.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/ContextRefresher.java
@@ -102,7 +102,20 @@ public class ContextRefresher {
                }
                finally {
                        ConfigurableApplicationContext closeable = capture;
-                       closeable.close();
+                       while (closeable != null) {
+                               try {
+                                       closeable.close();
+                               }
+                               catch (Exception e) {
+                                       // Ignore;
+                               }
+                               if (closeable.getParent() instanceof ConfigurableApplicationContext) {
+                                       closeable = (ConfigurableApplicationContext) closeable.getParent();
+                               }
+                               else {
+                                       break;
+                               }
+                       }
                }
 
        }
```

spencergibb:
Closed via abe5374c792094826bc7c96d36d44473757f4c91

zachmarshall:
I think this fixes #211 right?

PR #285, opened 2017-12-07
Increase the function, when users use spring retry spring retry listener can be used

holy12345:
Hello @spencergibb   @ryanjbaxter 
During retry, spring retry provides a listener mechanism. Within the listener (one or more listeners), users can do something based on the actual business needs (including the actual business need to decide whether to continue retry operation). I think this mechanism is good for the user. As the official spring-retry documentation says (`Often it is useful to be able to receive additional callbacks for cross cutting concerns across a number of different retries. For this purpose Spring Retry provides the RetryListener interface`). 
So I submitted this PR, hoping to add this feature and hopefully you can approve it. Looking forward to your reply.Best wish :)

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-commons.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-commons.adoc
@@ -414,7 +414,7 @@ you would like to use for a given service.
 @Configuration
 public class MyConfiguration {
     @Bean
-    LoadBalancedBackOffPolicyFactory backOffPolciyFactory() {
+    LoadBalancedBackOffPolicyFactory backOffPolicyFactory() {
         return new LoadBalancedBackOffPolicyFactory() {
             @Override
             public BackOffPolicy createBackOffPolicy(String service) {
@@ -429,7 +429,7 @@ NOTE: `client` in the above examples should be replaced with your Ribbon client'
 name.
 
 If you want to add one or more `RetryListener` to your retry you will need to
-create a bean of type `LoadBalancedRetryListenerFactory` and return the` RetryListener` array
+create a bean of type `LoadBalancedRetryListenerFactory` and return the `RetryListener` array
 you would like to use for a given service.
 
 [source,java,indent=0]


holy12345:
@ryanjbaxter Please help me review the code to see if this PR is suitable.thank you very much :)

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancedRetryListenerFactory.java
```diff
@@ -0,0 +1,37 @@
...
+
+package org.springframework.cloud.client.loadbalancer;
+
+import org.springframework.retry.RetryListener;
+
+/**
+ * Factory class to return the retry listeners.
+ * @author Gang Li
+ */
+public interface LoadBalancedRetryListenerFactory {
```
ryanjbaxter: I don't think there is any reason to create a factory.  Users can just define `RetryListener` beans and we can just pass them in `LoadBalancerAutoConfiguration`

holy12345:
Hi @ryanjbaxter 
First of all thank you very much for helping me check the code I submitted : )
The reason I define `LoadBalancedRetryListenerFactory` is because when the user defines multiple retryListener implementations,users can create multiple retryListeners via the `LoadBalancedRetryListenerFactory#createRetryListeners()` method while specifying the order in which they are to be executed (we know that the retryListener executes in the same order as the elements in the retryListener array).

Your advice is also great，according to your suggestion, I think that only need to change a small amount of code in `LoadBalancerAutoConfiguration `on it, which sounds really good.
The changes are as follows:
```
@Autowired(required = false)
private RetryListener[] retryListeners = new RetryListener[0];

@Bean
public RetryTemplate retryTemplate() {
    RetryTemplate template =  new RetryTemplate();
    if (retryListeners.length != 0) {
        retryTemplate.setListeners(retryListeners);
    }
    return template;
}

```
But there seems to be a problem here, that is we can not specify the order in which retryListener is executed.

To sum up, I think we should provide users with such a factory, to allow users to create retryListener at the same time be able to set the order of their implementation.
Maybe my understanding is not very correct, looking forward to your reply. Thanks again, best wishes!

ryanjbaxter:
Can you explain why you think the order is important?

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
```diff
@@ -87,6 +92,11 @@ public ClientHttpResponse intercept(final HttpRequest request, final byte[] body
 				!lbProperties.isEnabled() || retryPolicy == null ? new NeverRetryPolicy()
 						: new InterceptorRetryPolicy(request, retryPolicy, loadBalancer,
 						serviceName));
+		final RetryListener[] retryListeners = this.retryListenerFactory.createRetryListeners(serviceName);
```
ryanjbaxter: I dont think it is necessary to create the listeners on every intercept call, it is probably enough to create them once.

holy12345:
Hi @ryanjbaxter 
Thank you very much for your reply :)
I think there are some situations where users need to define multiple `retryListener`,it may be like this:
```
    @Bean
    public LoadBalancedRetryListenerFactory retryListenerFactory() {
        return new LoadBalancedRetryListenerFactory() {
            @Override
            public RetryListener[] createRetryListeners(String s) {
                RetryListener listenerA = new RetryListener() {
                    @Override
                    public <T, E extends Throwable> void onError(RetryContext context, RetryCallback<T, E> callback, Throwable throwable) {
                        /**
                         * Get the result of the previous listener setting
                         * and do something else based on this result.
                         */
                        Object object = context.getAttribute("key");
                    }
                };

                RetryListener listenerB = new RetryListener() {
                    @Override
                    public <T, E extends Throwable> void onError(RetryContext context, RetryCallback<T, E> callback, Throwable throwable) {
                        Object object = /**Do some business logic and get a result*/
                        context.setAttribute("key", object);
                    }
                };
                return new RetryListener[]{listenerA, listenerB};
            }
        };
       return new RetryListener[]{listenerA, listenerB};
    }
```
Through the above code, you can see the order of the two listeners is very important. As you can see from the above code, the order of the two listeners is important, and the value of `context.getAttribute ("key")` in the `onError ()` method in listenerA might be null if the order is not correct.

Some people may also say why we define two listeners? we can also combine all of the logic in the `onError()` method of these two listeners into one listener.I think this is not impossible but  this approach is not friendly, imagine if `onError ()` method contains a lot of business logic (I think we prefer to define multiple listeners and to differentiate between the functions of each listener based on the business while we specify their order)

To summarize, I hope you can consider my PR again, thank you very much. Looking forward to your reply, best wishes!

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
holy12345: @ryanjbaxter Thanks for helping me review the code.
Read `RetryLoadBalancerInterceptor#intercept()` source code:
`RetryTemplate template = this.retryTemplate == null ? new RetryTemplate() : this.retryTemplate;`
I found that each interception call will be re-creates a `RetryTemplate` (since `this.retryTemplate` is always null for every request coming in) so I will create listeners every time.
I understand so right? Hope to receive your reply. Thanks again.

holy12345:
@ryanjbaxter  ping...

spencergibb:
Please be patient. We all just returned from a weeks travel, it's holiday season in the US and the isn't a release imminent. Don't worry, we'll get to it.

holy12345:
@spencergibb Roger that. thank you for your reply. Sorry for the inconvenience caused. Please help me to see these two PRs when you are convenient, perhaps my idea is not mature, very willing to listen to your opinion. Thanks again

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
ryanjbaxter: Yeah the service name can change, this is fine then.

ryanjbaxter:
I think it is enough to create one listener class and separate business logic into methods within that class (if the logic is complicated like you say) rather than separate logic into separate listeners and pass state information in the context.  That seems much more complicated and not as strait forward.

holy12345:
Hi @ryanjbaxter
Your advice is very valuable :)
I still have a question i want to get your valuable suggestions.For the listener we need to do something for it?
And you discussed before, if we do not care about the order between the listener, I think only need to modify some code to increase this function 
The changes are as follows:
```
@Autowired(required = false)
private RetryListener[] retryListeners = new RetryListener[0];

@Bean
public RetryTemplate retryTemplate() {
    RetryTemplate template =  new RetryTemplate();
    if (retryListeners.length != 0) {
        retryTemplate.setListeners(retryListeners);
    }
    return template;
}

```

Or we do not do anything at all, because now we have allowed the customer to customize the `retryTemplate `(see my earlier fixes #286 ) Of course, custom `retryTemplate `can also customize the listener.(there may be a problem here is if customers want to add listener, then you must customize `retryTemplate `, this will not unfriendly?)

Which do you think is better? If it is the first one I will re-modify the PR in accordance with the above code.

Finally thank you again, really great.

ryanjbaxter:
I think this will suffice
```
@Bean
public RetryTemplate retryTemplate(RetryListener[] retryListeners) {
...
}

Would also need to make the changes in sc-netflix as well so we can leverage them there.

holy12345:
Hi @ryanjbaxter 
Thank you very much for your reply, then I will submit a PR to improve this function.

holy12345:
Hi @ryanjbaxter 
Thanks for your teaching, I re-created a PR # 293 and explained why I created it (hopefully you can help me check that PR's code). I closed this PR first.thanks again

PR #287, opened 2017-12-11
Increase recoveryCallback function

holy12345:
Hi @ryanjbaxter 
When the ribbon retries fail to return information usually is an exception,sometimes this may not be friendly. spring retry provides a callback mechanism (when the retry fails, the user can customize the return message instead of returning a exception).When we just use ribbon (without integrating Hystrix), I think this callback mechanism is useful.
So I submitted a PR, hoping to increase this function. I hope you can help me review the code, and give your advice, look forward to your reply. best wishes : )

### DIFF
--- a/spring-cloud-commons/src/test/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptorTest.java
+++ b/spring-cloud-commons/src/test/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptorTest.java
@@ -35,7 +35,6 @@ import static org.mockito.Mockito.when;
 
 /**
  * @author Ryan Baxter
- * @author Gang Li
  */
 @RunWith(MockitoJUnitRunner.class)
 public class RetryLoadBalancerInterceptorTest {


holy12345:
Hi @ryanjbaxter 
First of all thank you very much for responding to me :)

The reason why I added this feature is as follows:
- We did not provide a fallback mechanism when an exception occurred when the project only used ribbon (no hystrix was integrated). To implement the fallback mechanism, we must integrate the hystrix with the ribbon (to be able to use the hystrix fallback mechanism). I think this may result in a certain coupling (why just use the hystrix fallback feature (a part of the Hystrix feature), we need to integrate the entire Hystrix into our project).

- When we implement `RecoveryCallback,` we can get some information from the `RetryContext`(such as the number of retries, etc. ) I think this information is helpful to the user and we can not get that information from the hystrix fallback method .

- I think we should provide this fallback mechanism for using ribbon only which helps users to fallback even without using hystrix. As for the fallback mechanism using ribbon or hystrix, or both, I think this is user-determined.

This is my reason, it may not be very correct, I hope you can give your advice. Looking forward to your reply, best wishes！

holy12345:
@ryanjbaxter  ping...

spencergibb:
Please stop pinging

ryanjbaxter:
I am not convinced.  If users want to protect themselves from errors when making remote requests than they will need to add Hystrix to do that.  I have not seen anyone complain about having to do this before so I don't think it is an issue really.  As far as the additional information provided by the `RetryContext` I dont see that being particularly useful in any way.  In addition the user would have configured this anyways so they should know for example how many times the request was retried before it ultimately failed.  I also prefer Hystrix because it is not tied to the retry implementation, and can be used to protect the app from other types of remote requests (for example DB requests).

PR #293, opened 2017-12-15
Increase spring retry listener function

holy12345:
Hi @ryanjbaxter 
In response to the questions we discussed earlier(#285), I resubmitted a PR.I submit the code logic allow me to explain briefly:
Today I spent a lot of time thinking about this part of the code(feign and ribbon retry principle), in #285  we seem to come to a solution, the solution is as follows(in `LoadBalancerAutoConfiguration`)
```
@Autowired(required = false)
private RetryListener[] retryListeners = new RetryListener[0];

@Bean
public RetryTemplate retryTemplate() {
    RetryTemplate template =  new RetryTemplate();
    if (retryListeners.length != 0) {
        retryTemplate.setListeners(retryListeners);
    }
    return template;
}
```
But after my analysis, I think this seems to be not possible because `RetryTemplate ` will be recreated when the request is intercepted by the `RetryLoadBalancerInterceptor#intercept()` method.(Each request will re-create a `RetryTemplate`). `RetryTemplate `re-create is not get the `RetryListener ` whitch we created.
Based on the code ：(`this.retryTemplate` always `null`)
```
RetryTemplate template = this.retryTemplate == null ? new RetryTemplate() : this.retryTemplate;
BackOffPolicy backOffPolicy = backOffPolicyFactory.createBackOffPolicy(serviceName);
```

So I think we should still have a `LoadBalancedRetryListenerFactory ` that creates our retryListener each time we intercept the request, just as you did with the `BackOffPolicy ` feature (`BackOffPolicy`  is recreated for each request intercepted).
Following the principle of creating `BackOffPolicy` , I submitted this PR for creating a `RetryListener` .
I also made changes in sc-netflix so that we can use it.

Finally, I am very grateful to your patient teaching, hoping to pass this PR (increase `RetryListener ` function), looking forward to your reply. Thanks again, best wishes

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerAutoConfiguration.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerAutoConfiguration.java
@@ -106,6 +106,7 @@ public class LoadBalancerAutoConfiguration {
 	@ConditionalOnClass(RetryTemplate.class)
 	public static class RetryAutoConfiguration {
 		@Bean
+		@ConditionalOnMissingBean
 		public RetryTemplate retryTemplate() {
 			RetryTemplate template =  new RetryTemplate();
 			template.setThrowLastExceptionOnExhausted(true);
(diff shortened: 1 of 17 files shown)


lijiajia3515:
@OlgaMaciaszek 
Increase BlockingLoadBalancedRetryFactory backoff, because always NoBackOffPolicy!
`
public interface LoadBalancedRetryFactory {
   ...

	/**
	 * Creates a {@link BackOffPolicy} for a given service.
	 * @param service The service to create the {@link BackOffPolicy} for.
	 * @return The {@link BackOffPolicy}.
	 */
	default BackOffPolicy createBackOffPolicy(String service) {
		return new NoBackOffPolicy();
	}

}

OlgaMaciaszek:
@lijiajia3515 yes, the idea is that the user can implement the interface, instantiate and provide their own policy instead.

PR #309, opened 2018-01-25
Refactoring recovery callback implementation

ryanjbaxter:
These changes are mostly cosmetic in nature and simplifies the recovery callback implementation when retrying requests.  It helps with readability, but also allows for cleaner implementations in downstream projects.


ryanjbaxter:
@lly5044 would appreciate your review

lly5044:
I think It's ok. ^_^

PR #331, opened 2018-03-06
Refactor spring retry

ryanjbaxter:
Simplify the number of interfaces involved in configuring retry to one interface. 
Removing some deprecated constructors.


spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancedRetryFactory.java
```diff
@@ -0,0 +1,67 @@
...
+	public RetryListener[] createRetryListeners(String service);
+
+	/**
+	 * Creates a {@link BackOffPolicy} for a given service
+	 * @param service The service to create the {@link BackOffPolicy} for
+	 * @return The {@link BackOffPolicy}
+	 */
+	public BackOffPolicy createBackOffPolicy(String service);
+
+	static class DefaultRetryFactory implements LoadBalancedRetryFactory {
```
spencergibb: Could we use default methods rather than a static class?

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
```diff
@@ -99,47 +60,42 @@ public ClientHttpResponse intercept(final HttpRequest request, final byte[] body
...
+					serviceName, serviceInstance,
+					requestFactory.createRequest(request, body, execution));
+			int statusCode = response.getRawStatusCode();
+			if (retryPolicy != null && retryPolicy.retryableStatusCode(statusCode)) {
+				throw new ClientHttpResponseStatusCodeException(serviceName, response);
+			}
+			return response;
+		}, new LoadBalancedRecoveryCallback<ClientHttpResponse, ClientHttpResponse>() {
+			@Override
+			protected ClientHttpResponse createResponse(ClientHttpResponse response, URI uri) {
```
spencergibb: Looks like this could be pushed up as the default impl of the abstract method too. Simplify even more.

ryanjbaxter: I dont think so.  This is a special case where both parameters are the same type.  In every other case (in Netflix and Feign) the two type parameters are different.

spencergibb: Maybe a comment then

ryanjbaxter: done

holy12345:
`RetryTemplate ` related information may be placed alone in a method, to make it more concise?

ryanjbaxter:
@holy12345 can you be more specific?

holy12345:
Hi @ryanjbaxter 
Sorry, because I was on vacation these days, I did not reply to you.
I want to assemble all the information about RetryTemplate into one method：
```
private RetryTemplate createRetryTemplate(HttpRequest request, String serviceName, LoadBalancedRetryPolicy retryPolicy ) {
	RetryTemplate template = new RetryTemplate();
	BackOffPolicy backOffPolicy = lbRetryFactory.createBackOffPolicy(serviceName);
	template.setBackOffPolicy(backOffPolicy == null ? new NoBackOffPolicy() : backOffPolicy);
	RetryListener[] retryListeners = lbRetryFactory.createRetryListeners(serviceName);
	if (retryListeners != null && retryListeners.length != 0) {
		template.setListeners(retryListeners);
	}
	template.setRetryPolicy(!lbProperties.isEnabled() || retryPolicy == null ? new NeverRetryPolicy()
			: new InterceptorRetryPolicy(request, retryPolicy, loadBalancer, serviceName));
	return template;
}
```
Because all the above operations are about retryTemplate, so I think we should pull it out into a separate method so that it looks more concise

ryanjbaxter:
@holy12345 Within `RetryLoadBalancerInterceptor`?

holy12345:
Yes, what do you think?

ryanjbaxter:
Sure makes sense, thanks for the suggestion

PR #342, opened 2018-03-21
Move RSA properties to its own class and make it conditional

### LINKED ISSUE #334: KeyProperties has a hard dependency on RsaAlgorithm despite trying to make it optional
This was originally reported in https://github.com/spring-projects/spring-boot/issues/12408. 

[Here's a sample](https://github.com/spring-cloud/spring-cloud-commons/files/1794146/key-properties-bug.zip) that produces one variant of the problem. Reproducing the other variant (to do with the verifier during class loading) will require the use of a different JVM I believe. The problem's the same as [this one that we fixed in Spring Boot a while ago](https://github.com/spring-projects/spring-boot/commit/c0a2c88da10822cebdbc5e28ff48df1dc4b04295).

ryanjbaxter:
Avoids ClassNotFoundException when RSA is not on the classpath.  Fixes #334.

### DIFF
--- a/spring-cloud-context/src/test/java/org/springframework/cloud/bootstrap/encrypt/RsaDisabledTests.java
+++ b/spring-cloud-context/src/test/java/org/springframework/cloud/bootstrap/encrypt/RsaDisabledTests.java
@@ -1,0 +1,61 @@
+package org.springframework.cloud.bootstrap.encrypt;
+/*
+ * Copyright 2013-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


spring-cloud-context/src/test/java/org/springframework/cloud/bootstrap/encrypt/RsaDisabledTests.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+@RunWith(FilteredClassPathRunner.class)
+@ClassPathExclusions({"spring-security-rsa*.jar"})
+public class RsaDisabledTests {
+
+	private ConfigurableApplicationContext context;
+
+	@Before
+	public void setUp() {
+		context = new SpringApplicationBuilder().web(false)
```
spencergibb: The new WebApplicationContextRunner is good for this kind of stuff. Cleans itself up and has assertions for the context. (Not that you need to change this PR, just FYI).

Issue #352, opened 2018-04-09
Adding timer around refresh endpoint


JagmohanSharma:
We require to add timer around force refresh done. So We have overridden this endpoint for this purpose. Would it be possible to assign the same id "refresh" to this subclass. 

If this timer around refresh endpoint makes use for others too, I have created fork to submit pull request for this change.

PR #353, opened 2018-04-09
Adding timer around force refresh endpoint

JagmohanSharma:
Regarding issue - https://github.com/spring-cloud/spring-cloud-commons/issues/352


JagmohanSharma:
Yes it is with micrometer. :) 👍

JagmohanSharma:
@spencergibb Sorry for commenting on closed ticket but can we add metrics for any particular actuator endpoint as well?

PR #362, opened 2018-04-30
Polish RefreshScopeBeanDefinitionEnhancer.setRefreshable()

izeye:
If `this.refreshables` and `refreshables` are the same instance, invoking `addAll()` isn't necessary.


ryanjbaxter:
Why not just move the `addAll` call inside the if instead of changing the logic?

izeye:
@ryanjbaxter Thanks for the feedback! I usually handles exceptional cases first. I guess it's just my habit or style although it sometimes reduces nested block depths. In this case your suggestion looks better, so I updated as you suggested.

Issue #367, opened 2018-05-10
Could not have two extensions for the same endpoint


snicoll:
This was originally raised here: https://github.com/spring-projects/spring-boot/issues/13126

An endpoint operations is centrally managed and so is its tech specific extensions. 

See also #354

Issue #354, opened 2018-04-11
Standard actuator endpoints override should be qualified


snicoll:
Spring Cloud has an override of the standard `Environment` endpoint. To avoid clashes and deactivating standard Spring Boot features, bean names should at least be qualified. I'd actually recommend not to name the extension in such a generic fashion.

[`EnvironmentWebEndpointExtension`](https://github.com/spring-cloud/spring-cloud-commons/blob/78b11c35ff8ccc5782f2cb663ccec5bd19cc35fe/spring-cloud-context/src/main/java/org/springframework/cloud/context/environment/EnvironmentWebEndpointExtension.java#L35) is a `EndpointWebExtension` (that was renamed after this class was introduced, hence why the name pattern is probably inconsistent).

We do have an extension in Spring Boot that is named the same way. If you fix the naming of the class (and the bean name) you'll effectively override our behaviour with this one.

Actually, we forgot to rename the bean name when we renamed the base concept so this is the case right now, see https://github.com/spring-projects/spring-boot/issues/12827

I'd recommend renaming the class (and the bean name) to reflect what the extension does on top of the standard endpoint.

snicoll:
So this effectively breaks `Finchley.RC1` with Spring Boot `2.0.2.RELEASE` as we've renamed our bean for consistency and Spring Cloud no longer overrides it (by accident as described above).

You can't have two extensions for the same endpoint. IMO, this is consistent with the endpoint model that mandates that the operations of an endpoint are defined in one class. The fact Spring Cloud uses a web extension to add its operation feels suspicious to me.

I can see two ways out:

1. Override our extension and use the same name so that we have one extension in the context. If you got down that route, please add integration tests to validates that the standard Spring Boot behaviour is still available
2. Override our endpoint (which IMO would be more consistent) and adding your operation with the high level API. You can't do that with 2.0.2 because of https://github.com/spring-projects/spring-boot/issues/13082

My preference would be that this feature is in a dedicated endpoint, rather than extending from ours (and I know that's unrealistic at this point).

snicoll:
@sanderino666 a PR was merged in #368 and is available in `Finchley.BUILD-SNAPSHOT` already.

@dsyer the milestone information is missing so I've just added it.

sanderino666:
@snicoll Thanks!

illingtonFlex:
[...]

PR #368, opened 2018-05-10
Stop exposing a duplicate web extension for the env endpoint

snicoll:
An actuator endpoint is designed around a single class flagged with
`@Endpoint`. Such endpoint can have some of its operations refined with
a given tech (for instance via a `EndpointWebExtension` for web specific
concerns).

This design mandates that an endpoint (and its tech specific refinements
are managed centrally to form a unified contract).

As Spring Cloud is extending the standard Spring Boot's `env` endpoint,
this commit introduces a WritableEnvironmentEndpoint with a web
extension.

Closes gh-367
Closes gh-354


snicoll:
The update property operation should be moved there and refined in a JMX extension if necessary rather than having a separate `MBean` for this. This makes better use of the endpoint API.

dsyer:
Merged in bf5de108. Thank you. I'm not sure what the extra refactoring you suggested means, but at least we have something that works now.

snicoll:
You are effectively overriding the env endpoint. This changes is more consistent with that decision and will allow you to move some operations up so that they can be shared.

PR #378, opened 2018-06-05
Fix GenericScope so that multiple subclasses can live in the same con…

mbenson:
…text (non-static inner class cannot resolve parent bean)


mbenson:
broke the PR fixing the email

PR #379, opened 2018-06-06
Fix GenericScope so that multiple subclasses can live in the same con…

mbenson:
…text (non-static inner class cannot resolve parent bean)


spencergibb:
@dsyer thoughts?

dsyer:
LGTM. Why is the lock method package private though?

mbenson:
`private` didn't seem to be accessible to the inner class. I will double-check.

mbenson:
If accepted, I will also submit a PR to merge the commit to `master`.

spencergibb:
No need, we merge forward

mbenson:
My point was that there are conflicts which I was offering to resolve. ;) But either way, they're not significant.

spencergibb:
Closed via 3f9315dca1cde4720275d9768f883f60151bd792

PR #391, opened 2018-07-11
Extract target class code into utility class

ryanjbaxter:
This code will be useful throughout Spring Cloud

### DIFF
--- a/spring-cloud-context/pom.xml
+++ b/spring-cloud-context/pom.xml
@@ -54,10 +54,6 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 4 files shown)


spring-cloud-context/pom.xml
```diff
@@ -53,6 +53,10 @@
 			<artifactId>spring-boot-autoconfigure-processor</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<dependency>
```
spencergibb: Ug, maybe it should go the other way? `ProxyUtils` in context. I think it makes more sense here anyway.

Issue #318, opened 2018-02-04
Configuration of spring.datasource.hikari.* properties in Finchley.M5 causes Spring Boot 2.0.0.RC1 application to fail


pjverity:
Originally posted at [https://github.com/spring-projects/spring-boot/issues/11901](https://github.com/spring-projects/spring-boot/issues/11901), but problem appears to lie with spring-cloud-commons, so re-posting here.

----

Hi,

I see this problem in 2.0.0.RC1 but not in 2.0.0.M7. My application contains several `spring.datasource.hikari.*` properties. (`maximum-pool-size` or `schema` are examples that cause this problem)

This causes the application to fail to start with the exception:

```Caused by: java.lang.IllegalStateException: The configuration of the pool is sealed once started.  Use HikariConfigMXBean for runtime changes.```

It seems to be that the Datasource is configured on the initial refresh, and then again when this fires in `ConfigurationPropertiesRebinder`:

```java
@Override
public void onApplicationEvent(EnvironmentChangeEvent event) {
	rebind();
}
```

HikariCP V2.7.6 _seals_ the configuration preventing properties from being changed. It seems the above application event tries to 'reconfigure' the sealed datasource and throws the above exception, causing the application to fail to start.

The newer HikariCP dependency is introduced via `spring-boot-starter-jdbc` 2.0.0.RC1. The 2.0.0.M7 version pulls in HikariCP 2.7.4 which (on first glance) doesn't appear to seal the configuration in this way.

I can work around the issue for now by ensuring Hikari 2.7.4 appears first on the class path.

spencergibb:
Can you post the full stack trace?

pjverity:
Sure - here you go:

```
2018-02-04 22:11:02.236 DEBUG 6104 --- [           main] o.s.b.d.LoggingFailureAnalysisReporter   : Application failed to start due to an exception

org.springframework.boot.context.properties.bind.BindException: Failed to bind properties under 'spring.datasource.hikari' to com.zaxxer.hikari.HikariDataSource
	at org.springframework.boot.context.properties.bind.Binder.handleBindError(Binder.java:227) ~[spring-boot-2.0.0.RC1.jar:2.0.0.RC1]
	at org.springframework.boot.context.properties.bind.Binder.bind(Binder.java:203) ~[spring-boot-2.0.0.RC1.jar:2.0.0.RC1]
[...]

PR #395, opened 2018-07-25
Adds javax.sql.DataSource to refreshables

ileler:
Perfecting the processing method of [gh-318](https://github.com/spring-cloud/spring-cloud-commons/issues/318)

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/RefreshAutoConfiguration.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/RefreshAutoConfiguration.java
@@ -102,7 +102,7 @@ public class RefreshAutoConfiguration {
 		 * control the bean definition (e.g. it came from auto-configuration).
 		 */
 		private Set<String> refreshables = new HashSet<>(
-				Arrays.asList("com.zaxxer.hikari.HikariDataSource"));
+				Arrays.asList(new String[] { "javax.sql.DataSource", "com.zaxxer.hikari.HikariDataSource" }));
 
 		private Environment environment;
 


ileler:
Can you take a look at  https://github.com/spring-cloud/spring-cloud-commons/issues/318#issuecomment-407623361

Most of the time, we use interfaces, not subclasses, so I think 'javax.sql.DataSource' should in the default.

spencergibb:
Please try with boot 2.0.4 and Finchley.SR1

ileler:
I tried it, the problem is still the same, you can look at the [sample code](https://github.com/ileler/java/tree/hikari-exception/hikari-exception) provided in https://github.com/spring-cloud/spring-cloud-commons/issues/318#issuecomment-407623361.

spencergibb:
I can't run that code as I don't have mysql. Care to update it so there are no external dependencies (ie H2 or similar)?

ileler:
I have switched the database from mysql to h2, and the problem still exists. Please take a look.

spencergibb:
Adding the `Datasource` interface does indeed fix the issue, but is still only limited to Hikari. I'd rather not add it if Hikari is not used. Flagging for team discussion.

ileler:
Yes, but Hikari is used by default
😂

spencergibb:
What we're going to try and do is inspect the implementation type of the bean, that way we don't have to add `DataSource` which isn't needed if it is not Hikari.

spencergibb:
After trying to implement this, we are going to add a documentation note, but you should either set `spring.cloud.refresh.extra-refreshable=javax.sql.DataSource` or, more appropriately, strongly type your `DataSource` bean

```java
	@Primary
	@Bean(name = "dbDataSource")
	@ConfigurationProperties(prefix = "datasource.db")
	public HikariDataSource dbDataSource() {
		return DataSourceBuilder.create().type(HikariDataSource.class).build();
	}
```

Issue #411, opened 2018-09-07
Consider adopting @ConditonalOnRefreshScope from spring-cloud-netflix


brenuart:
`spring-cloud-netflix` defines two very useful annotations to test for the RefreshScope feature: `@ConditionalOnRefreshScope` and `@ConditionalOnMissingRefreshScope`.

These annotations are currently defined in `EurekaClientAutoConfiguration` at https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java#L320-L364

It would be nice if these annotations were incorporated into commons so they could easily be reused by others.

See https://github.com/spring-cloud/spring-cloud-sleuth/issues/1086 for a first use case.

brenuart:
Want a PR for this? And maybe another against `spring-cloud-netflix` ?

spencergibb:
@brenuart yes

spencergibb:
See comment here https://github.com/spring-cloud/spring-cloud-commons/pull/413#issuecomment-437901998

PR #413, opened 2018-09-08
Move @ConditionalOnRefreshScope from spring-cloud-netflix

brenuart:
This PR addresses issue https://github.com/spring-cloud/spring-cloud-commons/issues/411.

The first commit is a copy of the original the conditional classes from spring-cloud-netflix.
The second commit makes the conditions simpler due to the fact that they now belong to the same project/dependency as the `RefreshScope` class itself.
I made two distinct commits on purpose to highlight the changes and help reviewers determine if the proposed changes are acceptable or not.


spencergibb:
Why did you remove this part
```
@ConditionalOnMissingClass("org.springframework.cloud.context.scope.refresh.RefreshScope")
```
I'd think not having the class would be fine too.

spencergibb:
I guess we need to verify that s-c-netflix is indeed depedent on spring-cloud-context

spencergibb:
spring-cloud-netflix-eureka-client depends on spring-cloud-netflix-core which has spring-cloud-context as optional, though spring-cloud-starter-netflix-eureka-client includes spring-cloud-starter. Using this in spring-cloud-netflix-eureka-client will make this required

brenuart:
@spencergibb Oh, I see... using this condition instead of the one currently in spring-cloud-netflix-eureka-client would make the dependency on spring-cloud-context required. Is that what you mean?
Does it mean using `@ConditionalOnRefreshScope` would not work and would instead throw an exception at runtime if spring-cloud-context is not on the classpath?

If so, it is likely to be the case for any other similar use case... which makes the PR less useful I'm afraid :(

The idea behind this move into commons was to have the condition defined closer to where the feature lives and maintained by the guys who actually wrote it. I always feel uncomfortable when writing such conditions myself as I may not be aware of every ins and outs. See what I mean?

Can't we find another option ?

spencergibb:
I think the sentiment was right, but by virtue of having `@ConditionalOnRefreshScope` from spring-cloud-context, you'd also have `@RefreshScope`.

PR #415, opened 2018-09-19
Refactor String#replaceAll

durigon:
If we repeatedly call String#replaceAll, we internally repeatedly call the regular expression pattern compilation every time as following:
```java
    public String replaceAll(String regex, String replacement) {
        return Pattern.compile(regex).matcher(this).replaceAll(replacement);
    }
```
The modifications are to keep the compiled pattern.
Therefore, compiling a relatively expensive regular expression pattern does not have to be done every time.


ryanjbaxter:
Can you submit this against the 2.0.x branch?

durigon:
@ryanjbaxter I will submit my PR to the 2.0.x branch.

durigon:
I submitted new PR to the 2.0.x branch as following. 
https://github.com/spring-cloud/spring-cloud-commons/pull/417

So, I will close this PR.

PR #417, opened 2018-09-19
Refactor String#replaceAll

durigon:
If we repeatedly call String#replaceAll, we internally repeatedly call the regular expression pattern compilation every time as following:
```java
    public String replaceAll(String regex, String replacement) {
        return Pattern.compile(regex).matcher(this).replaceAll(replacement);
    }
```
The modifications are to keep the compiled pattern.
Therefore, compiling a relatively expensive regular expression pattern does not have to be done every time.


spencergibb:
While I don't see an issue with it, it's not really a performance issue since it's in a constructor.

durigon:
@spencergibb Since there are two calls to replaceAll in the constructor, it causes Pattern.compile to be called twice and iteration to be repeated twice.

durigon:
@ryanjbaxter @spencergibb  Thanks

Issue #425, opened 2018-10-15
Lost "mainApplicationClass" property while starting up using the "SpringBootServletInitializer" way


391861737:
**knock knock**

I recently found that my customed `EnvironmentPostProcessor#postProcessEnvironment` throws a `NullPointerException` when calling `application.getMainApplicationClass().getName()` when I deployed my web application in a tomcat server (in which application boots from `ServletContainerInitializer` under **Servlet 3.0** and invokes `SpringBootServletInitializer#onStartup` to initialize), but it works fine while I boot it using default springboot's main method.

As I traced, `org.springframework.cloud.bootstrap.BootstrapApplicationListener` listens on the `ApplicationEnvironmentPreparedEvent` and then build a different `SpringApplicationBuilder`(`#bootstrapServiceContext`) when creating the [Bootstrap Application Context](https://cloud.spring.io/spring-cloud-static/Finchley.SR1/single/spring-cloud.html#_the_bootstrap_application_context) where a new `SpringApplication` instance is created using it's constructor which deduces mainApplicationClass by searching stackTraces for the "main" method (`org.springframework.boot.SpringApplication#deduceMainApplicationClass`).
That "main" method is unfortunately absence while booting inside a tomcat server using `ServletContainerInitializer`. Hence my customed `EvnProcessor#postProcessEnvironment`'s second argument `SpringApplication`  which deduces none mainApplicationClass produces a `NullPointerException`.

After all, mainApplicationClass is really useful, and deploying in some web containers and booting from `ServletContainerInitializer` is somehow unavoidable.

Here is a demo which can reproduce this problem as you package it using different profile defined inside this demo, then run it [demo.zip](https://github.com/spring-cloud/spring-cloud-commons/files/2477535/demo.zip).
Different way to run different packages are as follows:
1. jar:   run it directly using `java -jar`.
2. war: run it in a web container (tomcat for instance). this will produce the NullPointerException I mentioned above.

ryanjbaxter:
Closed via bbc82b4

PR #426, opened 2018-10-15
Set MainApplicationClass if it is null

391861737:
MainApplicationClass will be null, if it is booted from SpringBootServletInitializer. So we need to set it properly.

Fixes gh-425

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/BootstrapApplicationListener.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/BootstrapApplicationListener.java
@@ -173,23 +173,23 @@ public class BootstrapApplicationListener
 				// Don't use the default properties in this builder
 				.registerShutdownHook(false).logStartupInfo(false)
 				.web(WebApplicationType.NONE);
-        final SpringApplication builderApplication = builder.application();
-        if(builderApplication.getMainApplicationClass() == null){
-            // gh-425:
-            // SpringApplication cannot deduce the MainApplicationClass here
-            // if it is booted from SpringBootServletInitializer due to the
-            // absense of the "main" method in stackTraces.
-            // But luckily this method's second parameter "application" here
-            // carries the real MainApplicationClass which has been explicitly
-            // set by SpringBootServletInitializer itself already.
-            builder.main(application.getMainApplicationClass());
-        }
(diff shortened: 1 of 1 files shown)


391861737:
@ryanjbaxter Looks like that `spring-boot-actuator-2.1.0.BUILD-20181014.221153-565.jar` does not satisfied with the dash mark "-", and throws IllegalArgumentException while `ServiceRegistryEndpointTests#testGet` is getting data from an endpoint of "service-registry".

Inside the `spring-boot-actuator-2.1.0.BUILD-20181014.221153-565.jar`, `org.springframework.boot.actuate.endpoint.EndpointId#of` asserts the argument to be as the pattern of `[a-zA-Z0-9]+`. 

Does this mean that we should change our endpoint's value, or ascribe this to the `spring-boot-actuator`?

391861737:
And I found this https://github.com/spring-projects/spring-boot/issues/14827

spencergibb:
Don't worry about that failure

ryanjbaxter:
Correct me if I am wrong, but the `spring-boot-actuator` issue is separate from the issue this PR is resolving, correct?

391861737:
That's correct that they are separate, but my tests' failure above shows me that -

image
which is thrown by EndpointID for the reason I mentioned above.

That's obstacle I'm facing right now, not until it is solved will my pull requests pass tests.

391861737:
How do I rerun the tests on CircleCI ? Do I need to add some space characters then re-push my code ?

391861737:
> Don't worry about that failure

What should do then... is there anything I can do such as changing the endpoint ids ?

391861737:
As the `spring-boot-actuator`  issue has been fixed, this PR is ok now.

ryanjbaxter:
You need to fix your branch now, there is a bunch of additional commits because your branch is still based on `master`

391861737:
Will that work ? It's late at night here amost midnight... I wake up and open my computer as soon as I saw the updates here...

391861737:
If I did not reply instantly, I must be in bed... Thanks a lot for your patience. Communication across different (actually opposite) time-zones must be the least efficient way and full of waiting...

ryanjbaxter:
Yeah time zones are rough....

Your branch is still not right though.

391861737:
I suppose that would work.

Issue #423, opened 2018-10-12
How to set BackOffPolicy since LoadBalancedBackOffPolicyFactory is no longer available 


ChrisDeBruyne:
According to the current documentation, http://cloud.spring.io/spring-cloud-static/Finchley.RELEASE/single/spring-cloud.html#_backoff_policies ,I should be able to use a LoadBalancedBackOffPolicyFactory.

This class was available through spring-cloud-commons 1.3.x (https://github.com/spring-cloud/spring-cloud-commons/blob/1.3.x/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancedBackOffPolicyFactory.java)

In the current version this class is no longer available and it seems there is no replacement for it.

What is the current solution to set the BackOffPolicy when retrying requests?

ryanjbaxter:
Nice catch, we need to update the documentation.  You want to create a bean of type `LoadBalancedRetryFactory`.  There is a method there to return a backoff policy.

https://github.com/spring-cloud/spring-cloud-commons/blob/2.0.x/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancedRetryFactory.java

ChrisDeBruyne:
Thanks Ryan for the clarification.

The bean is being picked up, but now I stumbled upon the next problem.

We are using OpenFeign through spring-cloud-starter-openfeign with Ribbon.
So RibbonAutoConfiguration sets up the RibbonLoadBalancedRetryFactory;
This RetryFactory returns null from the createBackOffPolicy method.
It is here I would want to add a BackOffPolicy.

It looks like the only way to achieve this is to copy the implementation of RibbonLoadBalancedRetryFactory to my own bean and add the BackOffPolicy in there.

But this feels a bit off.
Am I missing something here or is this really the way to go?

Wouldn't it be better if there was a way to inject a BackOffPolicy?

ryanjbaxter:
You can extend `RibbonLoadBalancedRetryFactory` and then just override `createBackOffPolicy`,

ryanjbaxter:
Fixed via 575fb31

PR #427, opened 2018-10-15
polish the document

holy12345:
See the issues #423


ryanjbaxter:
Fixed via 575fb31

holy12345:
Hi @ryanjbaxter 

Thanks for your update, one more thing I want to ask. I see this commit(it only include how to create `BackOffPolicy` but not say how to create `RetryListener` and `LoadBalancedRetryPolicy`)

On the other hand I found this section we also should modify

http://cloud.spring.io/spring-cloud-static/Finchley.RELEASE/single/spring-cloud.html#_retrying_failed_requests

It's make sense ?
Thanks again

ryanjbaxter:
Thanks.  We could add some documentation about how to create a `RetryListener` but that was not what this bug was about.

We also need to correct the documentation about backoff policies here http://cloud.spring.io/spring-cloud-static/Finchley.SR1/single/spring-cloud.html#_backoff_policies.  I submitted a quick fix for that here https://github.com/spring-cloud/spring-cloud-netflix/commit/70008f771733b8ad38465f69c6b877c15786176e

PR #429, opened 2018-10-17
Set MainApplicationClass if it is null

391861737:
MainApplicationClass will be null, if it is booted from SpringBootServletInitializer. So we need to set it properly.

Fixes gh-425


391861737:
Please delete me... I did it wrong...

PR #436, opened 2018-10-26
Check type before invoking bean factory 

ryanjbaxter:
Fixes https://github.com/spring-cloud/spring-cloud-config/issues/1167


dsyer:
LGTM. The thing that saves us is the `EnvironmentAware` I think.

dsyer:
See also c5138b5b for change merged to master

Issue #438, opened 2018-10-29
@ConditionalOnClass evaluation fails on missing classes on Java 11


mp911de:
Bootstap configurations that declare class-based conditions use `StandardAnnotationMetadata` . 
Absent classes cause a `ClassNotFoundException` because Java 11 changed slightly its behavior.

Bootstrap configuration should rather use ASM-based metadata parsing to avoid this issue.

Spring Cloud Commons version: 2.1 M1
Java version: 11

This ticket results from investigation in Spring Boot (see https://github.com/spring-projects/spring-boot/issues/14873) and Spring Cloud Vault (https://github.com/spring-cloud/spring-cloud-vault/issues/254#issuecomment-434064379).

spencergibb:
@mp911de in the two cases I see us using `AnnotationMetadata` it is with classes that implement `DeferredImportSelector` and `ImportBeanDefinitionRegistrar` and `AnnotationMetadata` is just passed in. How would we switch to ASM-based metadata parsing?

wilkinsona:
There's a stacktrace in [this comment on the Boot issue](
https://github.com/spring-projects/spring-boot/issues/14873#issuecomment-434072039) that shows where it's going wrong.

wilkinsona:
When the exception isn't thrown, the class is added directly as a source: https://github.com/spring-cloud/spring-cloud-commons/blob/5eb23a8c3c2c2669e6343720664ce2c656e4c88d/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/BootstrapApplicationListener.java#L207.

That's what causes `StandardAnnotationMetadata` to be used. You'd need to change that source registration mechanism to switch to ASM-based metadata.

wilkinsona:
I think you'll need to move the handling of the `BootstrapConfiguration` sources into an `ImportSelector` that returns their class names in the `String[]`. Switching from `Class` to `String` to represent the sources that should be used, will mean that Framework switches to ASM-based metadata.

spencergibb:
We already have the names as strings. Is there a different way to set sources on a `SpringApplicationBuilder`?

wilkinsona:
Not that will get Framework to go down the path of using ASM-based metadata. AFAIK, you need to use an `ImportSelector` for that.

PR #444, opened 2018-11-12
Moves BootstrapConfiguration classes to ImportSelector.

spencergibb:
This change moves a dynamic list of classes passed to
`SpringApplicationBuilder.sources` to a `DeferredImportSelector` loaded
by a simple `@Configuration` class. `BootstrapImportSelector` now loads
the classes and orders them and returns the class names so that
`@Conditional` annotations (such as `@ConditionalOnClass`) function
properly.

This was exposed while using java 11.

fixes gh-438

Still needs a test, but the code @mp911de posted here https://github.com/spring-projects/spring-boot/issues/14873#issuecomment-434051815 works properly.

/cc @wilkinsona

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/BootstrapImportSelector.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/BootstrapImportSelector.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2013-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.bootstrap;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.springframework.context.EnvironmentAware;
+import org.springframework.context.annotation.DeferredImportSelector;
+import org.springframework.core.annotation.AnnotationAwareOrderComparator;
+import org.springframework.core.env.Environment;
+import org.springframework.core.io.support.SpringFactoriesLoader;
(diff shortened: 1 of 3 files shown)


spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/BootstrapImportSelector.java
```diff
@@ -0,0 +1,78 @@
...
+		ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+		// Use names and ensure unique to protect against duplicates
+		List<String> names = new ArrayList<>(SpringFactoriesLoader
+				.loadFactoryNames(BootstrapConfiguration.class, classLoader));
+		names.addAll(Arrays.asList(StringUtils.commaDelimitedListToStringArray(
+				environment.getProperty("spring.cloud.bootstrap.sources", ""))));
+
+		List<Class<?>> sources = new ArrayList<>();
+		for (String name : names) {
+			Class<?> cls = ClassUtils.resolveClassName(name, null);
```
mp911de: Is class loading here still required? Seems as if Spring's `ConfigurationParser` considers ordering itself.

spencergibb: Our ordering tests fail without it.

spencergibb: @wilkinsona or @dsyer? I'm going to merge and we can revisit this particular nuance.

wilkinsona: I don't think the parser considers ordering of selected imports. Instead, the selector is expected to have done its own ordering. You could use the annotation metadata to get the `@Order` annotation and order things without class loading, but if this approach has worked up until now, it would appear to be good enough.

spencergibb: The supplied annotation metadata looked like it was only for the import selector class. Is there a way to get the annotation metadata for a class name?

wilkinsona: > The supplied annotation metadata looked like it was only for the import selector class

Doh. Silly me. 

>  Is there a way to get the annotation metadata for a class name?

`org.springframework.core.type.classreading.SimpleMetadataReaderFactory.getMetadataReader(String)`

spencergibb: This is what I came up with 1acb2b8b8bcaeac69284112b972215466190f03c

PR #445, opened 2018-11-13
Circuit Breaker Abstraction

ryanjbaxter:
The goal here is to come up with a simple API that would allow us to support different circuit breaker implementations.

We can extend this to initially support Hystrix in Spring Cloud Netflix, as well as Resiliency4J.


spring-cloud-commons/src/main/java/org/springframework/cloud/client/circuitbreaker/CircuitBreakerBuilder.java
```diff
@@ -0,0 +1,13 @@
+package org.springframework.cloud.client.circuitbreaker;
+
+/**
+ * Constructs circuit breakers.
+ *
+ * @author Ryan Baxter
+ */
+public interface CircuitBreakerBuilder {
+
+	public CircuitBreakerBuilder id(String id);
```
spencergibb: There are ways to make the return type generic so you can reference methods specific to the implementation.

ryanjbaxter: You mean on the `build` method?

ryanjbaxter: I'm kind of confused because you are working with the abstraction not the underlying implementation, thats the point

spencergibb: Your right. https://github.com/spring-cloud/spring-cloud-netflix/pull/3277/files#diff-c7ed37301464ccebbbb9c6efb0787d36R26 it works.

ryanjbaxter:
Code has been moved to incubator project https://github.com/spring-cloud-incubator/spring-cloud-circuitbreaker

Issue #442, opened 2018-11-12
Add autoconfiguration to assert Boot version


marcingrzejszczak:
We have a lot of questions about which version of Boot is supported with which version of the release train etc.

We could use an autoconfiguration to print a very nice message about this. Also, we could reuse this information in the releaser to print the matched versions.

PR #447, opened 2018-11-14
Spring Cloud CompatibilityVerifier

marcingrzejszczak:
Due to the fact that some users have problem with setting up Spring Cloud application, we've decided
to add a mismatch verification mechanism. It will break if your current setup is not compatible
with Spring Cloud requirements, together with a report, showing what exactly went wrong.

At the moment we verify which version of Spring Boot is added to your classpath.

In order to disable this feature, set `spring.cloud.util.mismatch-verifier.enabled` to `false`.
If you want to override the compatible Spring Boot versions, just set the
`spring.cloud.util.mismatch-verifier.compatible-boot-versions` property with a comma separated list
of compatible Spring Boot versions.

fixes gh-442


spencergibb:
Two things (reading on phone)
Where explicitly do we say this version of commons works with this version of boot?

Also, previously, multiple versions of boot were supported, I think this supports that, but want to verify. 

I think there is a SpringBootVersion class, did you look at that?

marcingrzejszczak:
> Where explicitly do we say this version of commons works with this version of boot?
> Also, previously, multiple versions of boot were supported, I think this supports that, but want to verify.

https://github.com/spring-cloud/spring-cloud-commons/pull/447/files#diff-abcf99993f9b3027eb877703965a5ad7R55 

The value of the `spring.cloud.util.mismatch-verifier.compatible-boot-versions` points to a comma separated list of compatible versions. That can be `1.5` but also `1.5.0.RELEASE` or `1.4,1.5`.

> I think there is a SpringBootVersion class, did you look at that?

Of course :) I start with checking it out and fallback to reflection if necessary - https://github.com/spring-cloud/spring-cloud-commons/pull/447/files#diff-abcf99993f9b3027eb877703965a5ad7R167

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/MismatchVerifierAutoConfiguration.java
```diff
@@ -0,0 +1,265 @@
...
+	@Bean
+	CompositeMismatchVerifier compositeMismatchVerifier(List<MismatchVerifier> verifiers) {
+		CompositeMismatchVerifier verifier = new CompositeMismatchVerifier(verifiers);
+		verifier.verifyDependencies();
+		return verifier;
+	}
+
+	@Bean
+	SpringBootVersionVerifier springBootVersionVerifier(
+			@Value("${spring.cloud.util.mismatch-verifier.compatible-boot-versions:1.5}") List<String> acceptedVersions) {
```
spencergibb: Let's make this `@ConfigurationProperties` rather than `@Value`. You lose functionality, such as relaxed binding, and it will make a more prominent place to change the compatible boot version.

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/MismatchVerifierAutoConfiguration.java
```diff
@@ -0,0 +1,265 @@
...
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} that fails the build fast for incompatible
+ * versions of dependencies (e.g. invalid version of Boot).
+ *
+ * @author Marcin Grzejszczak
+ * @since 1.3.6
+ */
+@Configuration
+@ConditionalOnProperty(value = "spring.cloud.util.mismatch-verifier.enabled", matchIfMissing = true)
```
spencergibb: Not sure I like the name. Maybe `spring.cloud.boot.version-verifier`? I you don't like the boot part since it's a composite, maybe just `spring.cloud.version-verifier`?

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/MismatchVerifierAutoConfiguration.java
```diff
@@ -0,0 +1,265 @@
...
+@AutoConfigureOrder(0)
+public class MismatchVerifierAutoConfiguration {
```
spencergibb: If we change the property, let's change the classes as well.

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/MismatchVerifierAutoConfiguration.java
marcingrzejszczak: I didn't want to actually mention the `version` in the name cause right now we're checking the version but maybe, in the future, we will be able to add more stuff there (I don't know what exactly)

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/MismatchVerifierAutoConfiguration.java
marcingrzejszczak: Makes sense!

spring-cloud-commons/src/test/java/org/springframework/cloud/commons/util/MismatchVerifierTests.java
```diff
@@ -0,0 +1,59 @@
+package org.springframework.cloud.commons.util;
```
ryanjbaxter: licesne

spring-cloud-commons/src/test/java/org/springframework/cloud/commons/util/SpringBootDependencyTests.java
```diff
@@ -0,0 +1,169 @@
+package org.springframework.cloud.commons.util;
```
ryanjbaxter: license

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/MismatchVerifierAutoConfiguration.java
spencergibb: Mismatch-verifier is awkward. Mismatch is a negative and verifier is a positive. I don't actually know what that means.

config verifier would work but config is overloaded.

Just looking at @ryanjbaxter's comment on docs.

`spring.cloud.compatibility-verifier`

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/MismatchVerifierAutoConfiguration.java
```diff
@@ -0,0 +1,265 @@
...
+		this.description = "";
+	}
+
+	// if not OK
+	VerificationResult(String errorDescription) {
+		this.description = errorDescription;
+	}
+}
+
+class DependencyMismatchReportMessage {
```
spencergibb: I wonder if we could throw a specific exception and then use boot's failure analyzer mechanism rather than creating our own reporting mechanism. https://www.baeldung.com/spring-boot-failure-analyzer

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/MismatchVerifierAutoConfiguration.java
marcingrzejszczak: Updated

Issue #440, opened 2018-10-31
Bootstrap context is not closed in case of main context startup failure


robotmrv:
Hi!
I've found that in case of the main context start failure its bootstrap context were not closed and disposable beans were not disposed.
This behavior is observed if bootstrap context has bean which creates its own non daemon thread.
I think this issue is similar to https://github.com/spring-cloud/spring-cloud-commons/issues/333 .

Here the example app
https://github.com/robotmrv/no-close-on-start-failure
com.example.nocloseonstartfailure.bootstrap.MyBootstrapConfig#myDisposableBean  - emulates such disposable bean.
com.example.nocloseonstartfailure.main.NoCloseOnStartFailureApplication#badBean - emulates bean which cannot be instantiated.

Test Case:
1. Try to start application 

Expected result: 
application start fails and java process is finished

Actual result: 
application start fails and java process is NOT finished

spencergibb:
This is very specific to non-daemon threads. Changing the thread in your sample to deamon and the java process ends.

spencergibb:
I think what you want to say in "Expected result" is: application start fails, bootstrap context is closed and java process is finished.

spencergibb:
@dsyer any thoughts on this?

spencergibb:
See https://github.com/spring-cloud/spring-cloud-commons/pull/450
Uses similar workaround but in the BootstrapApplicationListener.

zhanwh9:
When will this listener get invoked?
Before the CloseContextOnFailureApplicationListener is added to SpringApplication listeners, EventMulticaster has been created and has its copy of application listeners.
Thus the newly added CloseContextOnFailureApplicationListener will not be seen and invoked.

I'm using 1.2.5.RELEASE and trying to pick this fix.

PR #450, opened 2018-11-19
Closes bootstrap context on parent ApplicationFailedEvent.

spencergibb:
fixes gh-440


spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/BootstrapApplicationListener.java
```diff
@@ -103,7 +102,11 @@ public void onApplicationEvent(ApplicationEnvironmentPreparedEvent event) {
 		if (context == null) {
 			context = bootstrapServiceContext(environment, event.getSpringApplication(),
 					configName);
+			ConfigurableApplicationContext bootstrapContext = context;
+			event.getSpringApplication().addListeners((ApplicationListener<ApplicationFailedEvent>) afe ->
```
dsyer: This is very slow at runtime. It's better to have a `SmartApplicationListener` so the listener class doesn't have to be introspected.

spencergibb: ok, but is the approach the right one?

spencergibb: Ok, I've updated to use a `SmartApplicationListener`

PR #457, opened 2018-12-19
Allow overriding encrypted properties with unencrypted properties

jaarts:
Since the decrypted property source is taking precedence over all other property sources, it would not allow encrypted properties to be overridden by unencrypted properties. This commit removes the decrypted property from the overrides if it gets overridden by an unencrypted property.


ryanjbaxter:
I think this is OK but would like to know what @dsyer thinks.

Also if you would like this in Finchley, you will need to submit the PR against the 2.0.x branch.

dsyer:
I can't think of a reason not to do this, as long as the tests all pass.

jaarts:
@ryanjbaxter, Do I create a separate PR for Finchley or just change the base of this PR?

jaarts:
Alright, I've created PR #459 based on the 2.0.x branch.
Thanks!

PR #462, opened 2018-12-31
Only attempt to decrypt properties that are not overridden

jaarts:
Since decryption can be expensive (e.g. an http request needs to be made) or just not possible (in tests), it can be beneficial to not attempt to decrypt properties that get overridden anyway.

In this PR the decryption is postponed until after all the overrides are gathered.

Side effect is that `failOnError` will only fail if the property does not get overridden.

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializer.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializer.java
@@ -163,21 +163,25 @@ public class EnvironmentDecryptApplicationInitializer implements
 			sources.add(0, source);
 		}
 		for (PropertySource<?> source : sources) {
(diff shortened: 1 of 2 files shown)


skarzhevskyy:
The problem still observed when configuration is loaded from remote config server.
(This used to work in Finchley Release)

If you look at this test [54baded](https://github.com/skarzhevskyy/spring-cloud-commons/commit/54baded23cf6eed70fdacbdc963bb516157a708d)   it works fine when configuration loaded locally e.g. from application.yml

But when you attempt to load the same configuration from config server  as in [spring-cloud-config](https://github.com/spring-cloud/spring-cloud-config) you will get the error since system will try to decrypt value 'bar7-p1-clear' that is overridden. 

If you guys will point me out to integration test example in spring-cloud projects family I would be happy to create new test to demonstrate the problem using Config Server/ Config Client combination.

mpalourdio:
@skarzhevskyy I have submitted a potential fix. Please feel free to review : https://github.com/spring-cloud/spring-cloud-commons/pull/472

PR #469, opened 2019-01-17
Trim profile names in array

### LINKED ISSUE #460: String Array Type Property from yml is works differently.
It seems to use [ConfigFileApplicationListener](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigFileApplicationListener.java#L371
) to read spring's profile from file of application- "profile" .yml type except for the Spring cloud normally.
Bind the profile list to the String [] from the environment (using Binder)
Here, the best case is as follows.
```yml
spring.profiles.include:
      - far
      - boo
```
However, the following comma seperated forms work well.
```yml
spring.profiles.include: far,boo
```
Even comma seprated forms, including white space, work well.
```yml
spring.profiles.include: far, boo
```
However, in the case of comma seprated with white space, [PropertySourceBootstrapConfiguration](https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/config/PropertySourceBootstrapConfiguration.java#L235)
Does not seem to work normally when updating the profile at.
In this configuration, StringUtils is used to bind to Object to String array without using Binder.
Since we do not trim the white space here, the final result will be generated as follows.
```text
The following profiles are active:  far,boo,far
```
The format of yml, including white space, is not a regular method, but should not the behavior in PropertySourceBoottsrapConfiguration and ConfigFileApplicationListener work together?

ryanjbaxter:
Fixes #460

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/config/PropertySourceBootstrapConfiguration.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/config/PropertySourceBootstrapConfiguration.java
@@ -222,7 +222,12 @@ public class PropertySourceBootstrapConfiguration implements
 
 	private String[] getProfilesForValue(Object property) {
 		final String value = (property == null ? null : property.toString());
-		return StringUtils.commaDelimitedListToStringArray(value);
+		String[] values = StringUtils.commaDelimitedListToStringArray(value);
+		String[] trimmedValues = new String[values.length];
+		for(int i = 0; i < values.length; i++) {
+			trimmedValues[i] = values[i].trim();
+		}
+		return trimmedValues;
 	}
 
 }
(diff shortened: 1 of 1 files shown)


spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/config/PropertySourceBootstrapConfiguration.java
```diff
@@ -222,7 +222,12 @@ private void handleIncludedProfiles(ConfigurableEnvironment environment) {
 
 	private String[] getProfilesForValue(Object property) {
 		final String value = (property == null ? null : property.toString());
-		return StringUtils.commaDelimitedListToStringArray(value);
+		String[] values = StringUtils.commaDelimitedListToStringArray(value);
```
spencergibb: `StringUtils.tokenizeToStringArray()` does this automatically

Issue #471, opened 2019-01-23
{cipher} password not decrypted anymore


mpalourdio:
Hello,

I have just updated from SB 2.1.2 + Greenwich RC2 to Greenwich release.

I have a spring-cloud config server, and an eureka server. Eureka server fails at start when trying to connect to my local config server. It looks like the ciphered password in bootstrap.properties is not decrypted anymore. 

`ConfigServicePropertySourceLocator#addAuthorizationToken`, receives the {cipher} password in plain text, which leads to the stack trace below.

Reverting to Greenwich.RC2 fixes the problem. 

```
java.lang.IllegalStateException: Could not locate PropertySource and the fail fast property is set, failing
	at org.springframework.cloud.config.client.ConfigServicePropertySourceLocator.locate(ConfigServicePropertySourceLocator.java:136) ~[spring-cloud-config-client-2.1.0.RELEASE.jar:2.1.0.RELEASE]
	at org.springframework.cloud.config.client.ConfigServerHealthIndicator.getPropertySource(ConfigServerHealthIndicator.java:54) ~[spring-cloud-config-client-2.1.0.RELEASE.jar:2.1.0.RELEASE]
	at org.springframework.cloud.config.client.ConfigServerHealthIndicator.doHealthCheck(ConfigServerHealthIndicator.java:35) ~[spring-cloud-config-client-2.1.0.RELEASE.jar:2.1.0.RELEASE]
	at org.springframework.boot.actuate.health.AbstractHealthIndicator.health(AbstractHealthIndicator.java:84) ~[spring-boot-actuator-2.1.2.RELEASE.jar:2.1.2.RELEASE]
	at org.springframework.boot.actuate.health.CompositeHealthIndicator.health(CompositeHealthIndicator.java:98) [spring-boot-actuator-2.1.2.RELEASE.jar:2.1.2.RELEASE]
	at org.springframework.boot.actuate.health.HealthEndpoint.health(HealthEndpoint.java:50) [spring-boot-actuator-2.1.2.RELEASE.jar:2.1.2.RELEASE]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_191]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_191]
[...]

PR #358, opened 2018-04-17
Fix never decrypt CompositePropertySource

JoneWeng:
CompositePropertySource extends EnumerablePropertySource,

so never exec `else if` case

PR #473, opened 2019-01-28
Fix override behaviour when using CompositePropertySource.

jaarts:
The issue #471 is actually related to https://github.com/spring-cloud/spring-cloud-commons/pull/358 (which just activated an hidden issue). The propertySources from CompositePropertySource should be walked in reversed order for the overrides to behave properly. 

Fixes gh-471

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializer.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializer.java
@@ -15,7 +15,13 @@
  */
 package org.springframework.cloud.bootstrap.encrypt;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.regex.Pattern;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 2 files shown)


spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializer.java
```diff
@@ -15,12 +15,7 @@
  */
 package org.springframework.cloud.bootstrap.encrypt;
 
-import java.util.ArrayList;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
ryanjbaxter: We do not use '*' imports

jaarts: Alright, adjusted it.

ryanjbaxter:
This fix needs to go in the 2.0.x branch as well right?

jaarts:
#358 was not applied on the 2.0.x branch, so the code that is fixed is unreachable there.

mpalourdio:
Hello @ryanjbaxter @jaarts 

I have just tested this patch. Unfortunately, even if it fixes #471, it fails at decrypting properties that are fetched from configuration server, for example [this property](https://github.com/mpalourdio/configuration/blob/92ce67348fff3d8523ecd5fc2b5cc67f652cc402/application.properties#L12) in not decrypted at all on my test env.

Please let me know if you need more input.

ryanjbaxter:
Ugh ok, I reopened the issue

mpalourdio:
After freshly checking out master, it looks like what I said above was wrong. I musy have messed up applying the patch.
My property is well decrypted, but I still have exactly the same stacktrace that in #471

ryanjbaxter:
Can you move your comments to #471.  I dont understand it seems like the problem is fixed so why the stacktrace now?

PR #482, opened 2019-02-14
Add ConditionalOnDiscoveryEnabled.

OlgaMaciaszek:


spring-cloud-commons/src/main/java/org/springframework/cloud/client/ConditionalOnDiscoveryEnabled.java
```diff
@@ -0,0 +1,40 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+
+/**
+ * Provides a more succint conditional <code>spring.cloud.discovery.enbaled</code>.
```
PascalSchumacher: I guess this should be `enabled`.

PR #484, opened 2019-02-18
Fix overriding encrypted properties

jaarts:
This PR contains all the changes from PRs #459, #462 and #473, and includes fixes for #471.
It makes sure that:
1. Encrypted properties can be overridden by not encrypted properties (#459)
1. Properties are only decrypted if they are not overridden. (#462)
1. The property sources from `CompositePropertySource` are, similar to the sources from the environment, looped over in reversed order, so the override behaviour is correct. (#473 / #471)
1. Old decrypted property sources are removed from the environment before decrypting the environment again. (#471)

Besides the projects unit and integration tests, also the test projects from https://github.com/skarzhevskyy/playground-spring-cloud-integration-tests and https://github.com/mpalourdio/ConfigServer / https://github.com/mpalourdio/DiscoveryServer succeed.

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializer.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializer.java
@@ -167,13 +167,7 @@ public class EnvironmentDecryptApplicationInitializer implements
 
 	public Map<String, Object> decrypt(PropertySources propertySources) {
 		Map<String, Object> overrides = new LinkedHashMap<>();
-		List<PropertySource<?>> sources = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
I am unsure about 4.  Should we be removing it?  Doesn't having the decrypted properties around and not having to decrypt them again save us from having to do it everytime?

jaarts:
I guess we could be smarter about it, but currently there is no way to correlate the decrypted values to the corresponding encrypted value. So, when decrypting the second time, we cannot see if the encrypted value has changed or not, or even disappeared completely.

TYsewyn:
Thank you for your contribution! We couldn't merge this PR into master because of other changes.
We did however copied parts of your code from this PR over to https://github.com/spring-cloud/spring-cloud-commons/pull/582 which has been merged.
Therefor I'm also closing this one.

PR #565, opened 2019-05-13
refactor(spring-cloud-commons): Change the Event listened by Abstract…

chuntaojun:
…AutoServiceRegistration to Appl

Some services built by springboot have spring.main.web-application-type=none case, which will cause
the service to not be automatically registered.

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/serviceregistry/AbstractAutoServiceRegistration.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/serviceregistry/AbstractAutoServiceRegistration.java
@@ -25,8 +25,8 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.beans.BeansException;
+import org.springframework.boot.context.event.ApplicationReadyEvent;
 import org.springframework.boot.web.context.ConfigurableWebServerApplicationContext;
-import org.springframework.boot.web.context.WebServerInitializedEvent;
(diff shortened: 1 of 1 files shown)


spring-cloud-commons/src/main/java/org/springframework/cloud/client/serviceregistry/AbstractAutoServiceRegistration.java
```diff
@@ -84,20 +84,21 @@ protected ApplicationContext getContext() {
...
+	public void bind(ApplicationReadyEvent event) {
 		ApplicationContext context = event.getApplicationContext();
 		if (context instanceof ConfigurableWebServerApplicationContext) {
 			if ("management".equals(((ConfigurableWebServerApplicationContext) context)
 					.getServerNamespace())) {
 				return;
 			}
 		}
-		this.port.compareAndSet(0, event.getWebServer().getPort());
+		String port = context.getEnvironment().getProperty("server.port");
```
spencergibb: This will not work of it had not been set

chuntaojun: > This will not work of it had not been set

Enmm, I did not consider the reason that server.port is not set, I will change it. The reason why I am modifying the source code is because in some scenarios, the user sets the web container himself, but I hope that springcloud can automatically register the service if it listens. Is the WebServerInitializedEvent event, it is actually unable to automatically register the service.

chuntaojun:
> I'm afraid we can not accept these changes. The existing event needs to stay. If the service is not long-running and no port to register with discovery, what is the point?

:ok_hand:

PR #572, opened 2019-06-20
Extend ServiceInstanceChooser with new choose method

Aloren:
In case you need to choose instance based on some additional parameters there is no way to do that using `ServiceInstanceChooser`. It currently only gives ability to choose instance based only on service name.
`ServiceInstanceChooser` is implemented by `RibbonLoadBalancerClient` and already has this method `ServiceInstance choose(String serviceId, Object hint)` that is not exposed in the interface.

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/ServiceInstanceChooser.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/ServiceInstanceChooser.java
@@ -33,4 +33,14 @@ public interface ServiceInstanceChooser {
 	 */
 	ServiceInstance choose(String serviceId);
 
+	/**
+	 * Chooses a ServiceInstance from the LoadBalancer for the specified service and according to the specified hint.
+	 * @param serviceId The service ID to look up the LoadBalancer.
+	 * @param hint to specify the service instance
+	 * @return A ServiceInstance that matches the serviceId.
+	 */
+	default ServiceInstance choose(String serviceId, Object hint) {
+		return choose(serviceId);
+	}
+
 }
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
I am not sure if we can make this change in Hoxton since it changes the interface.  Maybe we can if we have a default implementation in the interface itself so it doesn't break anyone.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/ServiceInstanceChooser.java
```diff
@@ -33,4 +33,12 @@
 	 */
 	ServiceInstance choose(String serviceId);
 
+	/**
+	 * Chooses a ServiceInstance from the LoadBalancer for the specified service and according to the specified hint.
+	 * @param serviceId The service ID to look up the LoadBalancer.
+	 * @param hint to specify the service instance
+	 * @return A ServiceInstance that matches the serviceId.
+	 */
+	ServiceInstance choose(String serviceId, Object hint);
```
spencergibb: Change this to a `default` method that calls `choose(serviceId)`. You'll be able to remove the impls in `*Tests` that throw an exception.

Aloren: @spencergibb thank you for the suggestion. i've applied the change

spencergibb:
@ryanjbaxter @OlgaMaciaszek I wonder if we should do something similar to `ReactiveLoadBalancer` with a `Request` object. `Request` and `Response` aren't specific to the reactive lb and could be moved for common implementations.

ryanjbaxter:
@spencergibb I think that is worth while for sure

OlgaMaciaszek:
Makes sense to me. We could repackage it under `org.springframework.cloud.client.loadbalancer` and then they could be used in this PR.

OlgaMaciaszek:
Will get [it](https://github.com/spring-cloud/spring-cloud-commons/issues/772) done till tomorrow.

OlgaMaciaszek:
I'm actually going to add a method passing on the `Request` as part of the fix for: https://github.com/spring-cloud/spring-cloud-commons/issues/675; Working on it on this branch: https://github.com/spring-cloud/spring-cloud-commons/tree/lb-complete-lifecycle-new-gh-675.

Issue #576, opened 2019-07-04
Make @LoadBalanced WebClient.Builder use reactive Spring Cloud LoadBalancer


OlgaMaciaszek:
`@LoadBalanced WebClient.Builder` now requires a `LoadBalancerClient` implementation to work. All available implementations are non-reactive. Instead, it should work with `ReactiveLoadBalancer` implementations.

OlgaMaciaszek:
Backport to Greenwich for SR3.

PR #584, opened 2019-07-31
Gh 576 use reactive load balancer

OlgaMaciaszek:


spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerClient.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Represents a <a href="https://projectreactor.io/">Project-Reactor</a>-based client-side
+ * load balancer.
+ *
+ * @author Olga Maciaszek-Sharma
+ * @since 2.2.0
+ */
+
+public interface ReactorLoadBalancerClient {
```
spencergibb: I'm confused about this new class when we already have `ReactiveLoadBalancer`.

OlgaMaciaszek: that's the client that will retrieve an instance of the `ReactiveLoadBalancer` to use underneath based on the `serviceId` (will load either from default configuration or provided by `@LoadBalancerClient`. It also handles reconstructing the URI that is later used by the `ReactorLoadBalancerExchangeFunction`, but I can move this logic elsewhere. Please let me know where you think I should move it to.

OlgaMaciaszek:
Fixes gh-576

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerClient.java
OlgaMaciaszek: Changed.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerExchangeFilterFunction.java
```diff
@@ -0,0 +1,103 @@
...
+					originalUrl.toString());
+			LOG.warn(msg);
+			return Mono.just(
+					ClientResponse.create(HttpStatus.BAD_REQUEST).body(msg).build());
+		}
+		return choose(serviceId).handle((response, sink) -> {
+			ServiceInstance instance = response.getServer();
+			if (instance == null) {
+				String message = serviceInstanceUnavailableMessage(serviceId);
+				LOG.warn(message);
```
spencergibb: Add `if (LOG.isWarnAvailable()) { /* ... */ }`

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerExchangeFilterFunction.java
```diff
@@ -0,0 +1,103 @@
...
+				sink.error(new IllegalStateException(message));
+			}
+			else {
+				LOG.debug(String.format(
```
spencergibb: Add `if (LOG.isDebugAvailable()) { /* ... */ }`

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerExchangeFilterFunction.java
```diff
@@ -0,0 +1,103 @@
...
+
+	@Override
+	public Mono<ClientResponse> filter(ClientRequest request, ExchangeFunction next) {
+		URI originalUrl = request.url();
+		String serviceId = originalUrl.getHost();
+		if (serviceId == null) {
+			String msg = String.format(
+					"Request URI does not contain a valid hostname: %s",
+					originalUrl.toString());
+			LOG.warn(msg);
```
spencergibb: Add `if (LOG.isWarnAvailable()) { /* ... */ }`

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerExchangeFilterFunction.java
```diff
@@ -0,0 +1,103 @@
...
+			return Mono.just(
+					ClientResponse.create(HttpStatus.BAD_REQUEST).body(msg).build());
+		}
+		return choose(serviceId).handle((response, sink) -> {
```
spencergibb: Curious why `handle()` here rather than just `map()`?

spring-cloud-commons/src/test/java/org/springframework/cloud/client/loadbalancer/reactive/DefaultReactiveLoadBalancer.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.springframework.cloud.client.DefaultServiceInstance;
+import org.springframework.cloud.client.ServiceInstance;
+
+/**
+ * A sample implementation of {@link ReactiveLoadBalancer} used for tests.
+ *
+ * @author Olga Maciaszek-Sharma
+ */
+class DefaultReactiveLoadBalancer implements ReactiveLoadBalancer<ServiceInstance> {
```
spencergibb: How about `TestReactiveLoadbalancer`? `Default` makes me think it should not be in test :-)

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactiveLoadBalancer.java
```diff
@@ -42,4 +43,10 @@
 		return choose(REQUEST);
 	}
 
+	interface Factory<T> {
```
spencergibb: Probably should add `@FunctionalInterface` now that I think about it?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/annotation/LoadBalancerDiscoveryClientConfiguration.java
```diff
@@ -17,23 +17,32 @@
...
 /**
  * @author Spencer Gibb
+ * @author Olga Maciaszek-Sharma
  */
 @Configuration
 @EnableConfigurationProperties
-public class LoadBalancerClientConfiguration {
+@ConditionalOnBean(DiscoveryClient.class)
+@ConditionalOnDiscoveryEnabled
+public class LoadBalancerDiscoveryClientConfiguration {
```
spencergibb: Not sure I like the name change here. `@ConditionalOnBean(DiscoveryClient.class)` should probably move to the bean definition now.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/ReactorServiceInstanceLoadBalancer.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.cloud.client.ServiceInstance;
+
+/**
+ * A marker interface for {@link ReactorLoadBalancer} that allows selecting
+ * {@link ServiceInstance} objects.
+ *
+ * @author Olga Maciaszek-Sharma
+ * @since 2.2.0
+ */
+public interface ReactorServiceInstanceLoadBalancer
```
spencergibb: :+1: nice!

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerExchangeFilterFunction.java
OlgaMaciaszek: Wanted to be able to handle null `serviceId` and add more logs if it occurred informing the user why the operation failed, but I'm quite new to Reactor, so if you can think of a better way, I will be happy to change.

spring-cloud-commons/src/test/java/org/springframework/cloud/client/loadbalancer/reactive/DefaultReactiveLoadBalancer.java
OlgaMaciaszek: yes :) I think I had another `TestReactiveLoadBalancer` and I've named this one differently, but, yes, it does not make much sense, since they have different signatures anyway. Changing it.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerExchangeFilterFunction.java
OlgaMaciaszek: yes, so actually at first I wanted to return a different `onNext` already with the error response there, but then thought this would be easier handled by an exception. In this case, I can switch to `map`. Changing.

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -400,6 +400,15 @@ public class MyClass {
 The URI needs to use a virtual host name (that is, a service name, not a host name).
 The Ribbon client is used to create a full physical address.
 
+IMPORTANT: If you want to use a `@LoadBalanced WebClient.Builder` it is recommended that you add the `org.springframework.cloud:spring-cloud-loadbalancer` dependency to your project.
+Then, `ReactiveLoadBalancer` will be used underneath.
+Otherwise, the request will be handled by a non-reactive `LoadBalancerClient`.
+
+TIP: The `ReactorLoadBalancer` used underneath supports caching. If `cacheManager` is detected,
+cached version of `ServiceInstanceSupplier` will be used. If not, we will retrieve instances
```
ryanjbaxter: Possibly add info about what they need to do to enable this?  Point to the docs to enable caching?

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -514,7 +523,42 @@ TIP: If you see errors such as `java.lang.IllegalArgumentException: Can not set
 [[loadbalanced-webclient]]
 === Spring WebFlux WebClient as a Load Balancer Client
 
-`WebClient` can be configured to use the `LoadBalancerClient`. `LoadBalancerExchangeFilterFunction` is auto-configured if `spring-webflux` is on the classpath. The following example shows how to configure a `WebClient` to use load balancer:
+[[webflux-with-reactive-loadbalancer]]
+==== Spring WebFlux WebClient with Reactive Load Balancer
```
ryanjbaxter: It seems to make sense that we should make the reactive load balancer the default and allow them to optionally configure it to use a non reactive version if they want to.  Is there any reason why we wouldnt do that?

OlgaMaciaszek: If they add the `spring-cloud-loadbalancer` dependency (which I will be probably turning into a starter for M3), they will have the reactive one by default; but if somebody just has the previously used `commons` dependency in the classpath, it will just continue working as previously if they upgrade to new commons version (this is to avoid backward incompatibility); wdyt?

ryanjbaxter: I guess its the case where `org.springframework.cloud:spring-cloud-loadbalancer` is not on the classpath.  I think that should be reactive by default.  I understand it might be a change in behavior but is it noticeable at all?  I don't see any negatives to it being reactive by default or causing any problems.

OlgaMaciaszek: Updated the docs.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerExchangeFilterFunction.java
```diff
@@ -0,0 +1,110 @@
...
+			}
+			else {
+				if (LOG.isDebugEnabled()) {
+					LOG.debug(String.format(
+							"Load balancer has retrieved the instance for service %s: %s",
+							serviceId, instance.getUri()));
+				}
+				return instance;
+			}
+		}).flatMap(serviceInstance -> LoadBalancerUriTools.reconstructURI(serviceInstance,
```
smaldini: We shouldn't have to use a Mono for reconstructUri and therefore a flatmap here. Also I would try to merge a maximum of operator together (so do the reconstructURI in the map before)

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerExchangeFilterFunction.java
```diff
@@ -0,0 +1,110 @@
...
+				originalUrl)).map(uri -> buildClientRequest(request, uri))
```
smaldini: merge buildClientRequest with previous map()

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerExchangeFilterFunction.java
```diff
@@ -0,0 +1,110 @@
...
+				.flatMap(next::exchange)
+				.onErrorReturn(ClientResponse.create(HttpStatus.SERVICE_UNAVAILABLE)
```
smaldini: Why would you override the error with a client response here ? Shouldn't the error flow ?

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerExchangeFilterFunction.java
```diff
@@ -0,0 +1,110 @@
...
+				if (LOG.isDebugEnabled()) {
+					LOG.debug(String.format(
+							"Load balancer has retrieved the instance for service %s: %s",
+							serviceId, instance.getUri()));
+				}
+				return instance;
+			}
+		}).flatMap(serviceInstance -> LoadBalancerUriTools.reconstructURI(serviceInstance,
+				originalUrl)).map(uri -> buildClientRequest(request, uri))
+				.flatMap(next::exchange)
```
smaldini: merge flatMap with previous map

so it looks like :
```java
choose(serviceId).map(response -> {
			ServiceInstance instance = response.getServer();
			if (instance == null) {
				String message = serviceInstanceUnavailableMessage(serviceId);
				if (LOG.isWarnEnabled()) {
					LOG.warn(message);
				}
				throw new IllegalStateException(message);
			}

			if (LOG.isDebugEnabled()) {
				LOG.debug(String.format(
						"Load balancer has retrieved the instance for service %s: %s",
						serviceId, instance.getUri()));
			}
                           ClientRequest newRequest = buildClientRequest(request,
                                  LoadBalancerUriTools.reconstructURI(serviceInstance, originalUrl))
                        return next.exchange(newRequest);
		})
```

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerExchangeFilterFunction.java
OlgaMaciaszek: when we cannot get a valid instance, we want to get back to the user directly and respond with `SERVICE_UNAVAILABLE` so that they quickly know what has happened; we don't want to be going through with any other operations at this point.

PR #587, opened 2019-08-05
`ConfigurationProperties` changes can not be seen when `EnvironmentChangedEvent` happens

aftersss:
We use the following code and expect that after refresh, someProperties.getProperties() will be changed, but it does not. only when refresh is triggered one more time, the change can  be seen.

```
@Configuration
@EnableConfigurationProperties(SomeAutoConfiguration.SomeProperties.class)
public class SomeAutoConfiguration {
    
    @Resource
    SomeProperties someProperties;
    
    @EventListener(EnvironmentChangeEvent.class)
    public void onEnvChanged(){
        //after refresh, someProperties.getProperties() is not changed!
        System.out.println(someProperties.getProperties());
    }

    @ConfigurationProperties("some")
    public static class SomeProperties {

        private Map<String, String> properties = new LinkedHashMap<>();

        public Map<String, String> getProperties() {
            return properties;
        }

        public void setProperties(Map<String, String> properties) {
            this.properties = properties;
        }

    }
}
```

This PR improves this, give ConfigurationPropertiesRebinder a very high order so users can always see `ConfigurationProperties` changes when `EnvironmentChangedEvent` happens.

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/context/properties/ConfigurationPropertiesRebinder.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/context/properties/ConfigurationPropertiesRebinder.java
@@ -30,6 +30,7 @@ import org.springframework.cloud.util.ProxyUtils;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.ApplicationListener;
+import org.springframework.core.Ordered;
 import org.springframework.core.env.Environment;
 import org.springframework.jmx.export.annotation.ManagedAttribute;
 import org.springframework.jmx.export.annotation.ManagedOperation;
@@ -50,7 +51,7 @@ import org.springframework.stereotype.Component;
 @Component
 @ManagedResource
 public class ConfigurationPropertiesRebinder
-		implements ApplicationContextAware, ApplicationListener<EnvironmentChangeEvent> {
+		implements ApplicationContextAware, ApplicationListener<EnvironmentChangeEvent>, Ordered {
 
 	private ConfigurationPropertiesBeans beans;
 
@@ -127,6 +128,11 @@ public class ConfigurationPropertiesRebinder
 				|| event.getKeys().equals(event.getSource())) {
 			rebind();
 		}
+	}
+
+	@Override
+	public int getOrder() {
+		return Ordered.HIGHEST_PRECEDENCE + 10;
 	}
 
 }


ryanjbaxter:
Shouldn't the `@ConfigurationProperties` class above have an `@RefreshScope` annotation on it to make this work?

aftersss:
> Shouldn't the `@ConfigurationProperties` class above have an `@RefreshScope` annotation on it to make this work?

Maybe the creation of a bean is very expensive(for example, `DataSource`), and the properties of this bean don't change due to every refresh(most times `refresh` is to change other properties, not properties of this bean). If we use `@RefreshScope`, the bean will be recreated due to every refresh, but we only want to change some properties of this bean, without recreating, considering that recreating is too expensive.

spencergibb:
I still don't see what this fixes. Can you provide a test that fails prior to this change?

aftersss:
Here is a more meaningful example:

```
import java.util.LinkedHashMap;
import java.util.Map;
import javax.annotation.Resource;
import javax.sql.DataSource;

import org.apache.commons.dbcp2.BasicDataSource;

import org.springframework.boot.context.properties.ConfigurationProperties;
import org.springframework.boot.context.properties.EnableConfigurationProperties;
import org.springframework.cloud.context.environment.EnvironmentChangeEvent;
import org.springframework.context.ApplicationContext;
import org.springframework.context.annotation.Bean;
import org.springframework.context.annotation.Configuration;
import org.springframework.context.event.EventListener;

@Configuration
@EnableConfigurationProperties(MyDataSourceConfiguration.DataSourceDynamicProperties.class)
public class MyDataSourceConfiguration {
    
    @Resource
    DataSourceDynamicProperties dataSourceDynamicProperties;
    @Resource
    ApplicationContext applicationContext;

    @Bean("myDataSource")
    public DataSource myDataSource(){
        BasicDataSource dataSource = new BasicDataSource();
        dataSource.setUrl("***");
        dataSource.setUsername("***");
        dataSource.setPassword("***");
        dataSource.setMaxTotal(dataSourceDynamicProperties.getMaxTotal());
        return dataSource;
    }
    
    @EventListener(EnvironmentChangeEvent.class)
    public void onEnvChanged(){
        BasicDataSource dataSource = (BasicDataSource)applicationContext.getBean("myDataSource", DataSource.class);
        dataSource.setMaxTotal(dataSourceDynamicProperties.getMaxTotal());
        //dataSource.getMaxTotal() is still 10!
    }

    @ConfigurationProperties("my.datasource")
    public static class DataSourceDynamicProperties {

        private int maxTotal = 10;

        public int getMaxTotal() {
            return maxTotal;
        }

        public void setMaxTotal(int maxTotal) {
            this.maxTotal = maxTotal;
        }
    }
}
```

What I want to do is refresh the datasource's MaxTotal property dynamically using the following steps:

1. When this application is first started, maxTotal is `10`.(default value)

2. put key-value property pair `my.datasource.maxTotal=20` into configServer.

3. `curl -XPOST http://localhost/refresh` to refresh my application.

4. my expect is dataSource.getMaxTotal() being changed to 20, but this is still 10.

I found that this is because that `ConfigurationPropertiesRebinder` listens to `EnvironmentChangeEvent` too, and the order of `MyDataSourceConfiguration.onEnvChanged` and `ConfigurationPropertiesRebinder` is undefined,  so if `MyDataSourceConfiguration.onEnvChanged` is executed first, it can not see the change in `DataSourceDynamicProperties`. So I think `ConfigurationPropertiesRebinder` can be assigned with a very high order to let it be executed first, then `MyDataSourceConfiguration.onEnvChanged` can always see the change.

ryanjbaxter:
I still dont see why adding `@RefreshScope` to `DataSourceDynamicProperties` wouldnt do what you need

aftersss:
Adding @RefreshScope to DataSourceDynamicProperties still don't work, to make this work, another thing should be done too: change `@EventListener(EnvironmentChangeEvent.class)`  to `@EventListener(RefreshScopeRefreshedEvent.class)` . But If I do this, `@RefreshScope` is useless, it can be removed, and this still works, so this is confusing.

I think supporting `EnvironmentChangedEvent` in this example is better,  in my example, if I change the code to add order like the following, it still works:

```
 @EventListener(EnvironmentChangeEvent.class)
@Order(Ordered.LOWEST_PRECEDENCE)
    public void onEnvChanged(){
        BasicDataSource dataSource = (BasicDataSource)applicationContext.getBean("myDataSource", DataSource.class);
        dataSource.setMaxTotal(dataSourceDynamicProperties.getMaxTotal());
        .......
    }
```

This PR is to let users does not need to care about the order. just like what `OrderedCharactorEncoding` does.

PR #590, opened 2019-08-12
Fix gh 491 gh 553 non reactive loadbalancer client

OlgaMaciaszek:


OlgaMaciaszek:
Changes moved over from https://github.com/spring-cloud/spring-cloud-commons/pull/588 that was made against a different branch.

Issue #557, opened 2019-04-24
Provide A Reactive DiscoveryClient Implementation


ryanjbaxter:
Since the Spring Cloud LoadBalancer `LoadBalancer` implementation we will need to provide a reactive `ServiceDiscovery` client.

`DiscoveryClientServiceInstanceSupplier` assumes that the `DiscoveryClient` is reactive which may not be the case.  In fact we have no implementations that are currently reactive so the reactive load balancer implementation is going to block.

PR #593, opened 2019-08-22
Reactive service discovery client support

TYsewyn:
Closes gh-557


spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/composite/reactive/ReactiveCompositeDiscoveryClient.java
```diff
@@ -0,0 +1,74 @@
...
+	@Override
+	public Flux<ServiceInstance> getInstances(String serviceId) {
+		if (discoveryClients == null || discoveryClients.isEmpty()) {
+			return Flux.empty();
+		}
+		List<Flux<ServiceInstance>> serviceInstances = new ArrayList<>();
+		for (ReactiveDiscoveryClient discoveryClient : discoveryClients) {
+			serviceInstances.add(discoveryClient.getInstances(serviceId));
+		}
+		return Flux.first(serviceInstances);
```
ryanjbaxter: Why can't we merge all these together?

TYsewyn: Because of feature parity with the existing composite discovery client. Merging all streams would also make it really hard to filter out duplicates if a service has instances across multiple service registries.

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -251,6 +251,9 @@ This behavior can be disabled by setting `autoRegister=false` in `@EnableDiscove
 NOTE: `@EnableDiscoveryClient` is no longer required.
 You can put a `DiscoveryClient` implementation on the classpath to cause the Spring Boot application to register with the service discovery server.
 
+NOTE: Since 2.2.0 we also provide a `ReactiveDiscoveryClient` interface.
```
spencergibb: We tend not to mention versions in the doc. Maybe something like. "A `ReactiveDiscoveryClient` interface is now provided as well."

TYsewyn:
@spencergibb @ryanjbaxter Come to think of it, should we disable the `DiscoveryClient` configuration if the `ReactiveDiscoveryClient` is activated, or support both (which IMO is just overhead if you're running a non-blocking app)?

ryanjbaxter:
I think we should disable it, I cant think of a reason to have both.

ryanjbaxter:
How do we feel about getting this into Greenwhich at some point?  Probably too late for SR3 but maybe backport it for SR4?

TYsewyn:
Could make sense in order to make the transition from Greenwich to Hoxton more smooth. Let’s try to get it in M3 to get early feedback, fix/adapt things where/if necessary and backport it. 🙂 One of the things I would also try to do for M3 is adapt SC Gateway to benefit from this change too!

spring-cloud-commons/src/test/java/org/springframework/cloud/client/discovery/simple/reactive/SimpleReactiveDiscoveryClientAutoConfigurationTests.java
```diff
@@ -0,0 +1,107 @@
...
+		this.contextRunner.run((context) -> {
+			ReactiveDiscoveryClient client = context
+					.getBean(ReactiveDiscoveryClient.class);
+			assertThat(client).isNotNull();
+			assertThat(client.getOrder())
+					.isEqualTo(ReactiveDiscoveryClient.DEFAULT_ORDER);
+			InetUtils inet = context.getBean(InetUtils.class);
+			assertThat(inet).isNotNull();
+			SimpleDiscoveryProperties properties = context
+					.getBean(SimpleDiscoveryProperties.class);
```
OlgaMaciaszek: Maybe divide this test in smaller tests where each one has one setup block, one action block and one assertions block?

ryanjbaxter:
LGTM

PR #609, opened 2019-09-20
Use Ribbon by default.

OlgaMaciaszek:


spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerClientAutoConfiguration.java
```diff
@@ -41,40 +50,91 @@
...
-		this.webClientBuilders = webClientBuilders;
+	private static class ReactorLoadBalancerClientRibbonWarnLogger {
+
+		private static final Log LOG = LogFactory
+				.getLog(ReactorLoadBalancerClientRibbonWarnLogger.class);
+
+		@PostConstruct
+		void logWarning() {
+			if (LOG.isWarnEnabled()) {
+				LOG.warn("You have RibbonLoadBalancerClient on your classpath. "
```
ryanjbaxter: Shouldnt we state the Ribbon is in maintenance mode and they should switch to the Spring Cloud implementation?

PR #616, opened 2019-10-02
Support reactive discovery client for load balancing

TYsewyn:
Closes gh-614

### DIFF
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/DiscoveryClientServiceInstanceListSupplier.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/DiscoveryClientServiceInstanceListSupplier.java
@@ -19,9 +19,11 @@ package org.springframework.cloud.loadbalancer.core;
 import java.util.List;
 
 import reactor.core.publisher.Flux;
+import reactor.core.scheduler.Schedulers;
 
 import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.client.discovery.DiscoveryClient;
+import org.springframework.cloud.client.discovery.ReactiveDiscoveryClient;
 import org.springframework.core.env.Environment;
 
 import static org.springframework.cloud.loadbalancer.support.LoadBalancerClientFactory.PROPERTY_NAME;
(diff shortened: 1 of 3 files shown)


spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/DiscoveryClientServiceInstanceListSupplier.java
```diff
@@ -31,19 +33,28 @@
...
 public class DiscoveryClientServiceInstanceListSupplier
 		implements ServiceInstanceListSupplier {
 
-	private final DiscoveryClient delegate;
-
 	private final String serviceId;
 
+	private final Flux<ServiceInstance> serviceInstances;
+
 	public DiscoveryClientServiceInstanceListSupplier(DiscoveryClient delegate,
```
spencergibb: `@Deprecated`

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/DiscoveryClientServiceInstanceListSupplier.java
```diff
@@ -53,7 +64,7 @@ public String getServiceId() {
 
 	@Override
 	public Flux<List<ServiceInstance>> get() {
-		return Flux.defer(() -> Flux.just(delegate.getInstances(serviceId)));
+		return serviceInstances.collectList().flatMapMany(Flux::just);
```
spencergibb: Seems odd, but not sure I know any better. @smaldini or @simonbasle?

simonbasle: *if I'm reading that right*
if the `serviceInstances` is cold, ie it reflects the instances at the time it is subscribed to, then it is legit to remove `Flux.defer`. The original use of `defer` was to avoid eagerly capturing the `List` immediately on assembly.

The `flatMapMany` is overkill if the intent is to convert a `Mono<List>` to a `Flux<List>` (same content): use `.flux()`.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/DiscoveryClientServiceInstanceListSupplier.java
TYsewyn: I'm not sure we can deprecate this. Non-reactive applications need to have support for this too. At this moment the reactive discovery client will only be enabled if webflux is on the classpath and if `spring.cloud.discovery.reactive.enabled=true` has been set (which is also the default).

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/DiscoveryClientServiceInstanceListSupplier.java
TYsewyn: You're reading that right. Because we have a blocking and reactive service discovery now, we're keeping the reference of the flux. Both observables are cold (blocking via `Flux.defer`) so we're going to get the latest service instances on each subscribe.

Thanks for the `.flux()` hint! I'll adapt accordingly. 🙂

Issue #617, opened 2019-10-05
ReactiveDiscoveryClient.getInstances("...") returns an empty Flux


juergenzimmermann:
Using _Spring Cloud Hoxton.M3_ (spring-cloud-starter-config, spring-cloud-starter-consul-discovery which includes spring-cloud-starter-loadbalancer) and _Consul 1.6.1_ the following Kotlin fragment yields a Flux with 3 items:

- _consul_: i.e. Consul itself
- _config_: a registered config server with _Spring Cloud Config_
- _kunde_: a homegrown microservice _kunde_.

```
@Bean
fun discoveryClient(discoveryClient: ReactiveDiscoveryClient) = CommandLineRunner {
    val logger = ...
    discoveryClient.services
        .subscribe { service -> logger.warn("Registered Service: $service") }
}
```

However, the following Kotlin fragment results in an empty Flux when querying for the *existing* service _kunde_:

```
@Bean
fun discoveryClientKunde(discoveryClient: ReactiveDiscoveryClient) = CommandLineRunner {
    val logger = ...
    discoveryClient.getInstances("kunde")
        .collectList() // converts (empty) Flux to Mono with (empty) List
        .subscribe { instances -> logger.warn("Instances: $instances") }
}
```

When I use the injected non-reactive `DiscoveryClient` everything works fine.

Providing a testcase might be cumbersome (Consul, Config, microservice, ...). Please guide me to give you additional information.

juergenzimmermann:
Closed because the issue belongs to spring-cloud-consul: https://github.com/spring-cloud/spring-cloud-consul/issues/593

TYsewyn:
As mentioned in spring-cloud/spring-cloud-consul#593 the issue appears to reside in `ReactiveCompositeDiscoveryClient`, specifically in `return Flux.first(serviceInstances);`.
Therefor I'll reopen this issue and close the other one.

> `Flux.first` will return the first flux in the list of serviceInstances that emits any signal (onNext/onError/onComplete). Since the `SimpleReactiveDiscoveryClient` is being bootstrapped by default too and there is currently no way of en-/disabling this via properties, you will always end up having no instances (except if you have instances added to your properties file).

juergenzimmermann:
@TYsewyn ok, now I understand why I don't get an instance of a registered microservice.

[...]

PR #621, opened 2019-10-09
Take first non-empty flux into consideration for composite reactive service discovery

TYsewyn:
Fixes gh-617

### DIFF
--- a/spring-cloud-commons/src/main/java/reactor/core/publisher/CloudFlux.java
+++ b/spring-cloud-commons/src/main/java/reactor/core/publisher/CloudFlux.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2019-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package reactor.core.publisher;
+
+import org.reactivestreams.Publisher;
+
+/**
+ * @author Tim Ysewyn
+ */
+public abstract class CloudFlux<T> extends Flux<T> {
+
+	/**
+	 * Pick the first {@link Publisher} to emit an onNext signal and
+	 * replay all signals from that {@link Publisher}, effectively behaving like the
+	 * fastest of these competing sources.
+	 *
+	 * @param sources The competing source publishers
+	 * @param <I> The type of values in both source and output sequences
+	 *
+	 * @return a new {@link Flux} behaving like the fastest of its sources
+	 */
+	@SafeVarargs
+	public static <I> Flux<I> firstNonEmpty(Publisher<? extends I>... sources) {
+		return onAssembly(new FluxFirstNonEmptyEmitting<>(sources));
+	}
+
+	/**
+	 * Pick the first {@link Publisher} to emit an onNext signal and
+	 * replay all signals from that {@link Publisher}, effectively behaving like the
+	 * fastest of these competing sources.
+	 *
+	 * @param sources The competing source publishers
+	 * @param <I> The type of values in both source and output sequences
+	 *
+	 * @return a new {@link reactor.core.publisher.Flux} behaving like the fastest of its sources
+	 */
+	public static <I> Flux<I> firstNonEmpty(Iterable<? extends Publisher<? extends I>> sources) {
+		return onAssembly(new FluxFirstNonEmptyEmitting<>(sources));
+	}
+
+}
--- a/spring-cloud-commons/src/main/java/reactor/core/publisher/CloudFlux.java
+++ b/spring-cloud-commons/src/main/java/reactor/core/publisher/CloudFlux.java
@@ -24,13 +24,11 @@ import org.reactivestreams.Publisher;
 public abstract class CloudFlux<T> extends Flux<T> {
 
 	/**
-	 * Pick the first {@link Publisher} to emit an onNext signal and
-	 * replay all signals from that {@link Publisher}, effectively behaving like the
-	 * fastest of these competing sources.
-	 *
+	 * Pick the first {@link Publisher} to emit an onNext signal and replay all signals
+	 * from that {@link Publisher}, effectively behaving like the fastest of these
+	 * competing sources.
 	 * @param sources The competing source publishers
 	 * @param <I> The type of values in both source and output sequences
-	 *
 	 * @return a new {@link Flux} behaving like the fastest of its sources
 	 */
 	@SafeVarargs
@@ -39,16 +37,16 @@ public abstract class CloudFlux<T> extends Flux<T> {
 	}
 
 	/**
-	 * Pick the first {@link Publisher} to emit an onNext signal and
-	 * replay all signals from that {@link Publisher}, effectively behaving like the
-	 * fastest of these competing sources.
-	 *
+	 * Pick the first {@link Publisher} to emit an onNext signal and replay all signals
+	 * from that {@link Publisher}, effectively behaving like the fastest of these
+	 * competing sources.
 	 * @param sources The competing source publishers
 	 * @param <I> The type of values in both source and output sequences
-	 *
-	 * @return a new {@link reactor.core.publisher.Flux} behaving like the fastest of its sources
+	 * @return a new {@link reactor.core.publisher.Flux} behaving like the fastest of its
+	 * sources
 	 */
-	public static <I> Flux<I> firstNonEmpty(Iterable<? extends Publisher<? extends I>> sources) {
+	public static <I> Flux<I> firstNonEmpty(
+			Iterable<? extends Publisher<? extends I>> sources) {
 		return onAssembly(new FluxFirstNonEmptyEmitting<>(sources));
 	}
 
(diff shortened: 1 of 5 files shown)


spring-cloud-commons/src/main/java/reactor/core/publisher/FluxFirstNonEmptyEmitting.java
```diff
@@ -0,0 +1,347 @@
...
+	/**
+	 * Returns a new instance which has the additional source to be amb'd together with
+	 * the current array of sources.
+	 * <p>
+	 * This operation doesn't change the current FluxFirstEmitting instance.
+	 * @param source the new source to merge with the others
+	 * @return the new FluxFirstEmitting instance or null if the Amb runs with an Iterable
+	 */
+	@Nullable
+	FluxFirstEmitting<T> ambAdditionalSource(Publisher<? extends T> source) {
```
simonbasle: careful, you've copied that part without adapting it to the new operator. Note that since you only have a static operator, you don't need that part.

spring-cloud-commons/src/test/java/reactor/core/publisher/FluxFirstNonEmptyEmittingTests.java
```diff
@@ -0,0 +1,141 @@
...
+public class FluxFirstNonEmptyEmittingTests {
+
+	@Test(expected = NullPointerException.class)
+	public void arrayNull() {
+		CloudFlux.firstNonEmpty((Publisher<Integer>[]) null);
+	}
+
+	@Test(expected = NullPointerException.class)
+	public void iterableNull() {
+		new FluxFirstEmitting<>((Iterable<Publisher<Integer>>) null);
```
simonbasle: should be `new FluxFirstNonEmptyEmitting`

spring-cloud-commons/src/test/java/reactor/core/publisher/FluxFirstNonEmptyEmittingTests.java
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	@Test
+	public void iterableOneIsNullSource() {
+		StepVerifier
+				.create(CloudFlux.firstNonEmpty(Arrays.asList(Flux.never(),
+						(Publisher<Object>) null, Flux.never())))
+				.expectError(NullPointerException.class).verify();
+	}
+
```
simonbasle: this notably lacks any test that reflects what distinguish this operator from `Flux#first`: multiple tests that validate an early completing source like `Flux.empty()` vs other source(s) that emit an element. Testing the case where none of the sources emit any element would be interesting as well.

TYsewyn:
To clarify the functionality behind `firstNotEmpty` a little bit more: we want the first flux to emit an `onNext` or `onError` signal to win. Since `Flux.first` is also taking the first completion signal into account, this is not usable for us.
One might suggest to use `Flux.first(source1.switchIfEmpty(Flux.never()), first(source2.switchIfEmpty(Flux.never()), ...)`.
Unfortunately this has a downside to it: you need to be sure that *at least* one source is able to emit an `onNext` or `onError` signal.
Since we're contacting an external source for this use case (one or more service registries), we can't guarantee this.

spring-cloud-commons/src/main/java/reactor/core/publisher/CloudFlux.java
```diff
@@ -0,0 +1,58 @@
...
+ * in a next release.
+ *
+ * @author Tim Ysewyn
+ */
+public abstract class CloudFlux<T> extends Flux<T> {
+
+	/**
+	 * Pick the first {@link Publisher} to emit an onNext/onError signal and replay all
+	 * signals from that {@link Publisher}, effectively behaving like the fastest of these
+	 * competing sources. If none of the sources is emitting a signal, a completion signal
```
simonbasle: I'd rephrase a bit: "If all the source complete empty, a single completion signal is sent. Note that if all the sources are empty (never emit an element, ie. no onNext) AND at least one is also infinite (no onComplete/onError signal), the resulting `Flux` will be infinite and empty (like `Flux#never()`)."

spring-cloud-commons/src/main/java/reactor/core/publisher/FluxFirstNonEmptyEmitting.java
```diff
@@ -0,0 +1,344 @@
...
+					}
+
+					return true;
+				}
+			}
+			return false;
+		}
+
+		void resignFromRace() {
+			COMPETING_SUBSCRIBERS.decrementAndGet(this);
```
simonbasle: make this method return the result of the `decrementAndGet`, so that caller might directly compare it to `0` and propagate `onComplete`.

spring-cloud-commons/src/main/java/reactor/core/publisher/CloudFlux.java
simonbasle: to formalize the 3 cases I'm covering above:
```java
CloudFlux.firstNonEmpty(Flux.never(), Flux.empty(), Flux.just("foo")); //emits foo & completes
CloudFlux.firstNonEmpty(Flux.empty(), Flux.empty()); //completes
CloudFlux.firstNonEmpty(Flux.never(), Flux.empty()); //infinite nothing
```

spring-cloud-commons/src/main/java/reactor/core/publisher/FluxFirstNonEmptyEmitting.java
```diff
@@ -0,0 +1,344 @@
...
+			}
+			else if (parent.tryWin(index)) {
+				won = true;
+				actual.onError(t);
+			}
+		}
+
+		@Override
+		public void onComplete() {
+			parent.resignFromRace();
```
simonbasle: use the returned valued (see previous comment) to atomically decide to propagate complete

spring-cloud-commons/src/test/java/reactor/core/publisher/FluxFirstNonEmptyEmittingTests.java
```diff
@@ -0,0 +1,187 @@
...
+import static java.util.Collections.singletonList;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Tim Ysewyn
+ */
+public class FluxFirstNonEmptyEmittingTests {
+
+	@Test(expected = NullPointerException.class)
```
simonbasle: NB: when/if porting to Reactor, we'll prefer the `assertThatNullPointerException` style from AssertJ be used. If you can use AssertJ in spring-cloud tests, this will be one less review item when porting 😉

spring-cloud-commons/src/test/java/reactor/core/publisher/FluxFirstNonEmptyEmittingTests.java
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	@Test(expected = NullPointerException.class)
+	public void iterableNull() {
+		CloudFlux.firstNonEmpty((Iterable<Publisher<Integer>>) null);
+	}
+
+	@Test
+	public void firstWinner() {
+		StepVerifier
```
simonbasle: matter of preference but I find the `one StepVerifier method call per line` clearer to read in tests. Hopefully the code style lets you do that.

spring-cloud-commons/src/test/java/reactor/core/publisher/FluxFirstNonEmptyEmittingTests.java
```diff
@@ -0,0 +1,187 @@
...
+
+	@Test
+	public void bothEmpty() {
+		StepVerifier.create(CloudFlux.firstNonEmpty(Flux.empty(), Flux.empty()))
+				.expectComplete().verifyThenAssertThat().hasNotDiscardedElements()
+				.hasNotDroppedElements().hasNotDroppedErrors();
+	}
+
+	@Test
+	public void neverAndEmpty() {
```
simonbasle: one way to write this test and verify a longer period of time would be to use virtual time. You should also verify that no event occurs, like so:
```java
StepVerifier.withVirtualTime(CloudFlux.firstNonEmpty(Flux.never(), Flux.empty()))
    .expectSubscription()
    .expectNoEvent(Duration.ofDays(1))
    .thenCancel()
    .verifyThenAssertThat()
	.hasNotDiscardedElements().hasNotDroppedElements().hasNotDroppedErrors();
```

spring-cloud-commons/src/test/java/reactor/core/publisher/FluxFirstNonEmptyEmittingTests.java
```diff
@@ -0,0 +1,187 @@
...
+				.thenAwait(Duration.ofSeconds(2)).thenCancel().verifyThenAssertThat()
+				.hasNotDiscardedElements().hasNotDroppedElements().hasNotDroppedErrors();
+	}
+
+	@Test
+	public void firstEmitsError() {
+		RuntimeException ex = new RuntimeException("forced failure");
+		StepVerifier
+				.create(CloudFlux.firstNonEmpty(Flux.<Integer>error(ex), Flux.empty()))
+				.expectError(RuntimeException.class).verifyThenAssertThat()
```
simonbasle: ideally you would check the message at least? can use `expectErrorMessage("forced failure")` for that.

simonbasle: (same in the few tests below)

spring-cloud-commons/src/main/java/reactor/core/publisher/FluxFirstNonEmptyEmitting.java
```diff
@@ -0,0 +1,344 @@
...
+				}
+			}
+			return false;
+		}
+
+		void resignFromRace() {
+			COMPETING_SUBSCRIBERS.decrementAndGet(this);
+		}
+
+		boolean isLastRemainingFlux() {
```
simonbasle: with the above change, this method is not necessary anymore

spring-cloud-commons/src/test/java/reactor/core/publisher/FluxFirstNonEmptyEmittingTests.java
TYsewyn: Unfortunately the code is being reformatted during checkstyle. I will take it into account when this is being ported to `reactor-core`. 😉

PR #626, opened 2019-10-22
Add once method for Customizer of circuit breaker

making:
This PR adds `once` method to `org.springframework.cloud.client.circuitbreaker.Customizer` that guarantees `customize` method is called at most once per target.

Some `CircuitBreaker` implementations such as `Resilience4JCircuitBreaker` calls `customize` method every time `CircuitBreaker#run` is called.  It is sometimes inefficient and troublesome.
Using `once` method is useful where calling `customize` many times doesn't make sense,
for example, in case of [consuming Resilience4j's  events](https://resilience4j.readme.io/docs/circuitbreaker#section-consume-emitted-circuitbreakerevents).

Here is an example for each `io.github.resilience4j.circuitbreaker.CircuitBreaker` to consume events. 

```java
Customizer.once(circuitBreaker -> {
  circuitBreaker.getEventPublisher()
    .onStateTransition(event -> log.info("{}: {}", event.getCircuitBreakerName(), event.getStateTransition()));
}, CircuitBreaker::getName)
```


making:
@ryanjbaxter I've added documents and modified `once` logic.

spencergibb:
looks like you might need to rebase

making:
Fixed conflicts

Issue #619, opened 2019-10-07
SC LoadBalancer - can't build @LoadBalanced WebClient bean in constructor


OlgaMaciaszek:
For `@LoadBalanced WebClient.Builder` webClient instance is built in methods it works fine, but if it's built in constructor, the appropriate customizers seem not to have been executed on time.

PR #627, opened 2019-10-25
Fix gh 619 pass customisers in bean post processors

OlgaMaciaszek:
Fixes gh-619


OlgaMaciaszek:
> So if `WebClient.Builder.build()` is called in the constructor of a bean things fail, correct?
> 
> Is there a test that fails without this, that now passes?

Have verified it manually and it works - but, you're right, should be added there - will have a look at how to best check it there.

OlgaMaciaszek:
@spencergibb have added the test.

Issue #638, opened 2019-11-18
SC LoadBalancer: if fails to default `ConcurrentMapCacheManager` cache, does not evict entries


OlgaMaciaszek:
Add default and property-based TTL for the `ConcurrentMapCacheManager` default cache.

cashZhang27:
when i use spring-boot-starter-data-redis-reactive and spring-boot-starter-webflux,The second time call a service using webclient, it cannot be deserialized

PR #643, opened 2019-11-21
caffeine cache for loadbalancer Fix gh 638 

OlgaMaciaszek:
Fixes gh-638

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-commons.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-commons.adoc
@@ -759,6 +759,22 @@ in the following sections:
 * <<webclinet-loadbalancer-client, Spring WebClient as a Load Balancer Client>>
 * <<webflux-with-reactive-loadbalancer,Spring WebFlux WebClient with `ReactorLoadBalancerExchangeFilterFunction`>>
 
+=== Spring Cloud LoadBalancer Caching
+
+Apart from the basic `ServiceInstanceListSupplier` implementation that retrieves instances
+via `DiscoveryClient` each time it has to choose an instance, we provide a https://github.com/ben-manes/caffeine[Caffeine-backed]
+implementation. o make use of it, you need to have `com.github.ben-manes.caffeine:caffeine` in the classpath.
+The default setup includes `expireAfterWrite` set to 30 seconds and records set to soft references.
+
+You can override the default setup by passing your own https://static.javadoc.io/com.github.ben-manes.caffeine/caffeine/2.2.2/com/github/benmanes/caffeine/cache/CaffeineSpec.html[Caffeine Specification]
+in the `spring.cloud.loadbalancer.cache.caffeine.spec` property.
+
+You can also altogether disable loadBalancer caching by setting the value of `spring.cloud.loadbalancer.cache.enabled`
+to `false`.
+
+WARNING: Although the basic, non-cached, implementation is useful for prototyping and testing, it's much less efficient
+than the cached versions, so we recommend always using the cached version in production.
+
 [[spring-cloud-loadbalancer-starter]]
 === Spring Cloud LoadBalancer starter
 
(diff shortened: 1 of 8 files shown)


docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -759,18 +759,31 @@ in the following sections:
 * <<webclinet-loadbalancer-client, Spring WebClient as a Load Balancer Client>>
 * <<webflux-with-reactive-loadbalancer,Spring WebFlux WebClient with `ReactorLoadBalancerExchangeFilterFunction`>>
 
+=== Spring Cloud LoadBalancer Caching
+
+Apart from the basic `ServiceInstanceListSupplier` implementation that retrieves instances
+via `DiscoveryClient` each time it has to choose an instance, we provide a https://github.com/ben-manes/caffeine[Caffeine-backed]
+implementation. o make use of it, you need to have `com.github.ben-manes.caffeine:caffeine` in the classpath.
```
ryanjbaxter: typo, did you mean 
> To make use of it...

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/annotation/LoadBalancerClientConfiguration.java
```diff
@@ -73,12 +73,14 @@
 		@ConditionalOnMissingBean
 		public ServiceInstanceListSupplier discoveryClientServiceInstanceListSupplier(
 				ReactiveDiscoveryClient discoveryClient, Environment env,
-				ObjectProvider<CacheManager> cacheManager) {
+				ApplicationContext context) {
 			DiscoveryClientServiceInstanceListSupplier delegate = new DiscoveryClientServiceInstanceListSupplier(
 					discoveryClient, env);
-			if (cacheManager.getIfAvailable() != null) {
+			if (context.getBeanProvider(LoadBalancerCacheManager.class)
```
ryanjbaxter: is it more efficient to call `context.getBeanProvider` once?  regardless it would make the code a bit cleaner.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/cache/CaffeineBasedLoadBalancerCacheManager.java
```diff
@@ -0,0 +1,49 @@
...
+ * A Spring Cloud LoadBalancer specifi implementation of {@link CaffeineCacheManager} that
+ * implements the {@link LoadBalancerCacheManager} marker interface.
+ *
+ * @author Olga Maciaszek-Sharma
+ * @since 2.2.0
+ */
+public class CaffeineBasedLoadBalancerCacheManager extends CaffeineCacheManager
+		implements LoadBalancerCacheManager {
+
+	public CaffeineBasedLoadBalancerCacheManager(String specification) {
```
ryanjbaxter: I would be in favor of passing in the `LoadBalancerCacheProperties` object here for more flexibility in the future

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/cache/CaffeineBasedLoadBalancerCacheManager.java
```diff
@@ -0,0 +1,49 @@
...
+		super(SERVICE_INSTANCE_CACHE_NAME);
+		if (!StringUtils.isEmpty(specification)) {
+			setCacheSpecification(specification);
+		}
+		else {
+			setCaffeine(Caffeine.newBuilder().expireAfterWrite(30, TimeUnit.SECONDS)
```
ryanjbaxter: Would it be useful to make this a property itself so they dont have to provide their own specification to just adjust the time?  What about making the unit milliseconds for more fine grained control?

spencergibb: Make it a boot `Duration` and the user can use any format they want and convert it from there.

OlgaMaciaszek:
>  I think that we could be opinionated here and make caffeine required. Ribbons caching wasn't optional and we say you shouldn't go into production without it.

As we've later on discussed, though in principle I agree, the problem is that if we even only add the caffeine dependency to the starter, this can already mess up people's boot caching setup alone :/.

OlgaMaciaszek:
Fixes gh-638

PR #649, opened 2019-12-06
fix double checked locking with volatile

shenjianeng:
double checked locking with volatile

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/context/scope/GenericScope.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/context/scope/GenericScope.java
@@ -365,7 +365,7 @@ public class GenericScope implements Scope, BeanFactoryPostProcessor,
 
 		private final ObjectFactory<?> objectFactory;
 
-		private Object bean;
+		private volatile Object bean;
 
 		private Runnable callback;
 


pivotal-issuemaster:
@shenjianeng Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-commons&pullRequestId=649)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-commons&pullRequestId=649) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@shenjianeng Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-commons&pullRequestId=649)!

spencergibb:
Can you please explain the problem you're having and why this corrects it?

shenjianeng:
sorry, my English is poor。
> references:https://stackoverflow.com/questions/7855700/why-is-volatile-used-in-double-checked-locking
> The real problem is that Thread A may assign a memory space for instance before it is finished constructing instance. Thread B will see that assignment and try to use it. This results in Thread B failing because it is using a partially constructed version of instance.

spencergibb:
I understand the use of volatile. What problem are you seeing in spring cloud commons that prompted the change?

shenjianeng:
not yet but for preventing potential problems

>`org.springframework.cloud.context.scope.GenericScope.BeanLifecycleWrapper#getBean`
> double checked locking without volatile.

f4cklangzi:
If you don't add volatile, there may be deadlocks in getBean. I encountered

Issue #632, opened 2019-11-01
[Enhancement] consider zone aware loadbalancer


bilak:
Please consider implementing zone aware loadbalancer.
For example with eureka it's possible to define zones in which are services located (where they are running), however currently it's not possible to loadbalance based on zones (currently only roundrobin is supported).

If I can somehow help with implementation let me know.

Thanks

OlgaMaciaszek:
Thanks @bilak Will be working on it this week.

bilak:
Are you going to implement it based on discovery service or would it be possible to configure balancer with static IP addresses and zone metadata?

spencergibb:
It will be based on the discovery client, but it can be setup via configuration.

PR #652, opened 2019-12-10
Add each PropertySource from PropertySourceLocators individually

ryanjbaxter:
Fixes #611

PR #658, opened 2019-12-17
Zoned ServiceInstanceListSupplier

OlgaMaciaszek:
Fixes gh-632

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/ServiceInstance.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/ServiceInstance.java
@@ -24,14 +24,8 @@ import java.util.Map;
  *
  * @author Spencer Gibb
  * @author Tim Ysewyn
- * @author Olga Maciaszek-Sharma
  */
 public interface ServiceInstance {
-
-	/**
-	 * Metadata zone key.
-	 */
-	String ZONE = "zone";
 
 	/**
 	 * @return The unique instance ID as registered.
@@ -74,18 +68,6 @@ public interface ServiceInstance {
 	 * @return The scheme of the service instance.
 	 */
 	default String getScheme() {
-		return null;
-	}
-
-	/**
-	 * @return The {@link String} value from instance metadata corresponding to the
-	 * {@link ServiceInstance#ZONE} key.
-	 */
-	default String getZone() {
-		Map<String, String> metadata = getMetadata();
-		if (metadata != null) {
-			return metadata.get(ZONE);
-		}
 		return null;
 	}
 
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/ServiceInstance.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/ServiceInstance.java
@@ -24,9 +24,13 @@ import java.util.Map;
  *
  * @author Spencer Gibb
  * @author Tim Ysewyn
+ * @author Olga Maciaszek-Sharma
  */
 public interface ServiceInstance {
 
+	/**
+	 * Metadata zone key.
+	 */
 	String ZONE = "zone";
 
 	/**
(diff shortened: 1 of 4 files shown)


spring-cloud-commons/src/main/java/org/springframework/cloud/client/ServiceInstance.java
```diff
@@ -71,4 +77,16 @@ default String getScheme() {
 		return null;
 	}
 
+	/**
+	 * @return The {@link String} value from instance metadata corresponding to the
+	 * {@link ServiceInstance#ZONE} key.
+	 */
+	default String getZone() {
```
spencergibb: I'd rather not pollute `ServiceInstance` with shortcuts for metadata. I think this can go safely in the `ZonePreferenceServiceInstanceListSupplier`

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/ZonePreferenceServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,78 @@
...
+	public Flux<List<ServiceInstance>> get() {
+		return delegate.get().map(this::filteredByZone);
+	}
+
+	private List<ServiceInstance> filteredByZone(List<ServiceInstance> serviceInstances) {
+		if (zone == null) {
+			zone = environment.getProperty("spring.cloud.loadbalancer.zone");
+		}
+		if (zone != null) {
+			List<ServiceInstance> filteredInstances = serviceInstances.stream()
```
spencergibb: See @dsyer's comment https://github.com/spring-cloud/spring-cloud-commons/pull/652#discussion_r357203393 about `stream()`. Let's change to a loop here.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/ZonePreferenceServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,78 @@
...
+					.filter(serviceInstance -> serviceInstance.getZone() != null)
+					.filter(serviceInstance -> zone
+							.equalsIgnoreCase(serviceInstance.getZone()))
+					.collect(Collectors.toList());
+			if (filteredInstances.size() > 0) {
```
spencergibb: I assume this is here so that if everything is filtered out, we return the original list as something is better than nothing? If so can you make a small comment here about that.

OlgaMaciaszek: yes, that's how it is. Will add that.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/ZonePreferenceServiceInstanceListSupplier.java
OlgaMaciaszek: Didn't realise Stream API had performance issues. Hope they improve as it's so much more readable to me. Will switch it here, however.

OlgaMaciaszek:
> This LGTM, but I do have one question.
> 
> I might be confusing concepts here, but with Ribbon I believe we always got all instances across all zones but if there was an instance within the same zone Ribbon would select that instance above the others (and only try the others if it failed and we retried). With this implementation if we only return instances within the same zone we would loose the functionality of trying services in another zone when retrying wouldn't we?

That's probably something to think about. Ribbon has a whole logic of verifying whether the instances are up and running. We, however, re-query the SR to get the running instances. In the cached version, this might be slightly affected, but reducing the TTL should work.

We don't also really provide any retry support OOTB here, but it might be a good idea to add it. We could then do the first execution on a filtered subset and then retry on the entire set. I think we could provide sth like a `RetrySILS` that could have two different delegates - primary and fallback- (say, zoned and simple discovery service) and would retry on the second one (with a broader scope) when the request has failed with the first one. That seems like something for another issue, though; WDYT?

spencergibb:
Yes, it would be another issue

PR #665, opened 2020-01-06
some methods' name and comments replace ribbon by load balancer

fangjian0423:
Now Ribbon is in maintenance mode and will be replaced by Spring Cloud LoadBalancer.

I think the method names and comments in *LoadBalancerAutoConfiguration should remove the Ribbon-related texts because it depends on new client-side load balancer interface - `LoadBalancerClient`.


OlgaMaciaszek:
Hello @fangjian0423 . These comments and names refer to Ribbon, because these classes actually set Ribbon up. There are other config classes that set up SC LoadBalancer contained in this package: https://github.com/spring-cloud/spring-cloud-commons/tree/master/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive . When we remove Ribbon support altogether, we will simply delete the classes that auto-configure it.

PR #667, opened 2020-01-08
Trim scheme double slash

ArtyomGabeev:
Fixes https://github.com/spring-cloud/spring-cloud-kubernetes/issues/486


spencergibb:
Let's do this in s-k-k8s

Issue #647, opened 2019-12-03
Spring Cloud Loadbalancer: Expose API for `Request` or "hint"


andersenleo:
**Is your feature request related to a problem? Please describe.**

As described in a previous feature request (in the netflix-ribbon project): https://github.com/spring-cloud/spring-cloud-netflix/issues/3009 -- we have a need to implement a stateful loadbalancer based on various criterias. 

Basically - in order to pick the correct service-instance we need to pass along a "hint" (or, in Netflix Ribbon, "key") to the load-balancer implementation. As far as I can tell there's really no way of doing that in the new Loadbalancer feature in Spring Cloud?

**Describe the solution you'd like**

See the proposal in the original ticket. If it's also possible to pass along information about the source request to allow the load-balancer implementation to, for example - extract a "key" from a header or query parameter, it would be even more useful.

**Describe alternatives you've considered**

Hacking around the limitations using ThreadLocal - but that won't work with the Reactive LoadBalancer (which seems to be the core abstraction).
request here.

OlgaMaciaszek:
Thanks for reporting this. We should make this available, indeed. I don't think we want to modify the `ReactiveLoadBalancer` interface, but we could probably handle it via the `Publisher<Response<T>> chose(Request request)` method.

spencergibb:
That's exactly what the `choose(Request)` is for

PR #669, opened 2020-01-10
Add hints to API via the Request interface. Fixes gh-647.

OlgaMaciaszek:
Fixes gh-647

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/DefaultRequest.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/DefaultRequest.java
@@ -22,24 +22,24 @@ package org.springframework.cloud.client.loadbalancer.reactive;
  * @author Spencer Gibb
  * @author Olga Maciaszek-Sharma
  */
-public class DefaultRequest implements Request<DefaultRequestContext> {
+public class DefaultRequest<T> implements Request<T> {
 
-	private DefaultRequestContext context;
+	private T context;
 
 	public DefaultRequest() {
 		new DefaultRequestContext();
 	}
 
-	public DefaultRequest(DefaultRequestContext context) {
+	public DefaultRequest(T context) {
 		this.context = context;
 	}
 
 	@Override
-	public DefaultRequestContext getContext() {
+	public T getContext() {
 		return context;
 	}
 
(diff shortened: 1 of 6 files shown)


OlgaMaciaszek:
Incomplete. Will reopen after adding more changes.

spencergibb:
I was thinking of having some generics, so you could pass the http request to the loadbalancer for example.

```
public interface Request<C> {

   C getContext();

}
```

I don't want to limit ourselves to a string like ribbon did.

/cc @twicksell

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/DefaultRequest.java
```diff
@@ -17,8 +17,30 @@
 package org.springframework.cloud.client.loadbalancer.reactive;
 
 /**
+ * A default implementation of {@link Request}.
+ *
  * @author Spencer Gibb
+ * @author Olga Maciaszek-Sharma
  */
-public class DefaultRequest implements Request {
+public class DefaultRequest implements Request<DefaultRequestContext> {
```
spencergibb: I think this should be `<T>` still. And then when instantiated user can use `<DefaultRequestContext>`. Maybe?

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/Request.java
```diff
@@ -20,8 +20,23 @@
...
+	default C getContext() {
+		return null;
+	}
+
+	/**
+	 * A hint that can be used by {@link ReactiveLoadBalancer} implementations to pick the
+	 * correct instance.
+	 * @return a {@link String} representation of the hint
+	 */
+	default String getHint() {
```
spencergibb: Why have two methods?

OlgaMaciaszek: yes - that's a mistake.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/DefaultRequest.java
OlgaMaciaszek: ok

Issue #629, opened 2019-10-29
Add mechanism verifying servers are not down to SC LoadBalancer


OlgaMaciaszek:

OlgaMaciaszek:
Consider making it default when used with `SimpleDiscoveryClient`. See [this issue](https://github.com/spring-cloud/spring-cloud-gateway/issues/1482).

OlgaMaciaszek:
Done.

PR #682, opened 2020-01-30
Fix gh 629 healthcheck for loadbalancer

OlgaMaciaszek:
Fixes gh-629


spencergibb:
To not have my commits duplicated you should have force pushed.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/InstanceHealthChecker.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.springframework.cloud.client.ServiceInstance;
+
+/**
+ * Provides API for verifying service instance health.
+ *
+ * @author Olga Maciaszek-Sharma
+ * @since 2.2.0
+ */
+public interface InstanceHealthChecker {
```
spencergibb: We're getting closer to Ribbon with the addition of another interface. This is something I want to actively avoid.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,116 @@
...
+			healthyInstances.addAll(verifiedInstances);
+		});
+	}
+
+	private Flux<List<ServiceInstance>> healthCheckFlux() {
+		return Flux.create(emitter -> Schedulers
+				.newSingle("Health Check Verifier: " + getServiceId(), true)
+				.schedulePeriodically(() -> {
+					List<ServiceInstance> verifiedInstances = new ArrayList<>();
+					Flux.fromIterable(instances).filterWhen(healthChecker::isAlive)
```
spencergibb: Change this to a protected method here. Move the implementation from `PingHealthChecker`. If user wants a different implementation, let them extend this class and reuse the rest of the code.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/LoadBalancerProperties.java
```diff
@@ -41,4 +48,55 @@ public void setZone(String zone) {
...
+	public void setHealthCheck(HealthCheck healthCheck) {
+		this.healthCheck = healthCheck;
+	}
+
+	public static class HealthCheck {
+
+		/**
+		 * Initial delay value for the HealthCheck scheduler.
+		 */
+		private int initialDelay = 0;
```
spencergibb: These can be of type `Duration` so the user can say `30s` and there's no need for a separate time unit.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/config/LoadBalancerAutoConfiguration.java
```diff
@@ -49,6 +51,15 @@ public LoadBalancerAutoConfiguration(
 		this.configurations = configurations;
 	}
 
+	// Adding a @Primary bean to avoid NoUniqueBeanDefinition exception during constructor
```
spencergibb: Hmm. This shouldn't be needed. `LoadBalancerAutoConfiguration` should only be once.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,116 @@
...
+			.synchronizedList(new ArrayList<>());
+
+	private List<ServiceInstance> healthyInstances = Collections
+			.synchronizedList(new ArrayList<>());
+
+	private final InstanceHealthChecker healthChecker;
+
+	public HealthCheckServiceInstanceListSupplier(ServiceInstanceListSupplier delegate,
+			LoadBalancerProperties loadBalancerProperties,
+			InstanceHealthChecker healthChecker) {
```
spencergibb: Maybe only pass the `HealthCheck` properties?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/PingHealthChecker.java
```diff
@@ -0,0 +1,70 @@
...
+
+	public PingHealthChecker(Environment environment) {
+		webClient = WebClient.builder().build();
+		this.environment = environment;
+		defaultHealthCheckPath = environment
+				.getProperty(LOADBALANCER_PREFIX + HEALTH_CHECK_SUFFIX, ACTUATOR_HEALTH);
+	}
+
+	@Override
+	public Mono<Boolean> isAlive(ServiceInstance serviceInstance) {
```
spencergibb: Move to `HealthCheckServiceInstanceListSupplier`.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/PingHealthChecker.java
```diff
@@ -0,0 +1,70 @@
...
+		String healthCheckPropertyValue = environment.getProperty(LOADBALANCER_PREFIX
```
spencergibb: Yeah, not sure I like this ribbon style of property. I'd prefer a single default path property and then a true map that we can model as `@ConfigurationProperties` with serviceId as the key.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
OlgaMaciaszek: ok

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/InstanceHealthChecker.java
OlgaMaciaszek: Ok, thought that would be easier for people to customise, but can switch it.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/config/LoadBalancerAutoConfiguration.java
OlgaMaciaszek: There was a problem during the automated constructor argument resolution by framework when `@Primary` bean was not defined here. Was getting:

`"Error creating bean with name 'healthCheckServiceInstanceListSupplier' defined in io.github.olgamaciaszek.excluded.CustomLoadBalancerConfiguration: Unsatisfied dependency expressed through method 'healthCheckServiceInstanceListSupplier' parameter 2; nested exception is org.springframework.beans.factory.NoUniqueBeanDefinitionException: No qualifying bean of type 'org.springframework.cloud.client.loadbalancer.reactive.LoadBalancerProperties' available: expected single matching bean but found 2: spring.cloud.loadbalancer-org.springframework.cloud.client.loadbalancer.reactive.LoadBalancerProperties,loadBalancerProperties"` as part of the `ConstructorResolver` exectution.

I was not entirely sure what was causing that as wherever that bean is explicitly instantiated, it is always annotated with `@ConditionalOnMissingBean`.

Any thoughts @spencergibb ?

spencergibb: I see. It's already defined in `LoadBalancerClientConfiguration`. Can you eliminate this bean definition?

OlgaMaciaszek: Yes, that's right. Thought `@ConditionalOnMissingBean` would handle any such thing, but it seems it gets instantiated before and then the one generated from `@ConfigurationProperties` clashes.

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -923,6 +923,65 @@ public class CustomLoadBalancerConfiguration {
...
+requests against the `health` endpoint of the instance. You can also provide your own implementation
+of `InstanceHealthChecker` instead.
+
+`HealthCheckServiceInstanceListSupplier` uses properties prefixed with
+`spring.cloud.loadbalancer.healthcheck`. You can set the `initialDelay`, `period` and `unit`
+for the scheduler.
+
+For the `PingHealthChecker`, you can set the default path for the healthcheck URL by setting
+the value of the `spring.cloud.loadbalancer.healthcheck.path`. You can also set a specific value
+for any given service by settin the value of the `spring.cloud.loadbalancer.[SERVICE_ID].healthcheck.path`,
```
ryanjbaxter: typo "settin"

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -923,6 +923,65 @@ public class CustomLoadBalancerConfiguration {
...
+substituting the `[SERVICE_ID]` with the correct Id of your service. If the path is not set,
```
ryanjbaxter: "Id" should be all caps

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	@Override
+	public Flux<List<ServiceInstance>> get() {
+		if (!healthyInstances.isEmpty()) {
+			return Flux.defer(() -> Flux.fromIterable(healthyInstances).collectList());
+		}
+		// If there are no healthy instances, it might be better to still retry on all of
+		// them
+		if (LOG.isDebugEnabled()) {
```
ryanjbaxter: I think this should we a `WARN`

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/PingHealthChecker.java
```diff
@@ -0,0 +1,70 @@
...
+	private static final String ACTUATOR_HEALTH = "/actuator/health";
+
+	private final WebClient webClient;
+
+	private final Environment environment;
+
+	private final String defaultHealthCheckPath;
+
+	public PingHealthChecker(Environment environment) {
+		webClient = WebClient.builder().build();
```
ryanjbaxter: Should we make this customizable?  I could see someone wanting to maybe customize the behavior of the `WebClient`.  It might be nice if they didnt have to override this class to do so.  Maybe pass it in the constructor at least?

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -923,6 +923,65 @@ public class CustomLoadBalancerConfiguration {
...
 ----
 
+=== Instance Health-Check for LoadBalancer
+
+It is possible to enable a scheduled HealthCheck for the LoadBalancer. The `HealthCheckServiceInstanceListSupplier`
+is provided for that. It regularly verifies if the instances provided by a delegate
+`ServiceInstanceListSupplier` are still alive and only returns the healthy instances,
+unless there are none - then it returns all the retrieved instances.
+
+TIP: This mechanism is particularly helpful while using the `SimpleDiscoveryClient`. For the
```
ryanjbaxter: I think we also need to add some documentation about when the `SimpleDiscoveryClient` is used.  I dont see that in the current docs.

OlgaMaciaszek: It's a good idea, but I would add it in the `DiscoveryClient` section, not necessarily as a part of this PR.

OlgaMaciaszek:
Something has gone wrong during the rebase. I'm going to close this PR and submit a new one in order  to avoid messing up git log.

PR #683, opened 2020-02-03
Fix gh 629 new

OlgaMaciaszek:
Add healhtcheck-based loadBalancing. Fixes gh-629.


spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,137 @@
...
+		// them
+		if (LOG.isWarnEnabled()) {
+			LOG.warn(
+					"No verified healthy instances were found, returning all listed instances.");
+		}
+		return Flux.defer(() -> Flux.fromIterable(instances).collectList());
+	}
+
+	// Visible for tests
+	Mono<Boolean> isAlive(ServiceInstance serviceInstance) {
```
spencergibb: Protected, so users can override?

OlgaMaciaszek: They can override the method that calls it, but can make this protected, anyway.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,136 @@
...
+
+	protected Mono<Boolean> isAlive(ServiceInstance serviceInstance) {
+		String healthCheckPropertyValue = healthCheck.getPath()
+				.get(serviceInstance.getServiceId());
+		String healthCheckPath = healthCheckPropertyValue != null
+				? healthCheckPropertyValue : defaultHealthCheckPath;
+		return webClient.get()
+				.uri(UriComponentsBuilder.fromUri(serviceInstance.getUri())
+						.path(healthCheckPath).build().toUri())
+				.exchange()
```
robotmrv: @OlgaMaciaszek  @spencergibb  
this code uses `exchange()` but does not consume `ClientResponse` body  
see https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html#webflux-client-exchange  

> Unlike retrieve(), when using exchange(), it is the responsibility of the application to consume any response content regardless of the scenario (success, error, unexpected data, etc). Not doing so can cause a memory leak. The Javadoc for `ClientResponse lists all the available options for consuming the body. Generally prefer using retrieve() unless you have a good reason for using exchange() which does allow to check the response status and headers before deciding how to or if to consume the response.
--

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,136 @@
...
+		healthCheckFlux.subscribe(verifiedInstances -> {
+			healthyInstances.clear();
+			healthyInstances.addAll(verifiedInstances);
+		});
+	}
+
+	protected Flux<List<ServiceInstance>> healthCheckFlux() {
+		return Flux.create(emitter -> Schedulers
+				.newSingle("Health Check Verifier: " + getServiceId(), true)
+				.schedulePeriodically(() -> {
```
robotmrv: Task is never disposed  and cannot be disposed.  
What if one wants to refresh/recreate `HealthCheckServiceInstanceListSupplier`  bean. It could lead to leak and unnecessary pinging.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,136 @@
...
+	private final LoadBalancerProperties.HealthCheck healthCheck;
+
+	private final WebClient webClient;
+
+	private final String defaultHealthCheckPath;
+
+	private List<ServiceInstance> instances = Collections
+			.synchronizedList(new ArrayList<>());
+
+	private List<ServiceInstance> healthyInstances = Collections
```
robotmrv: `synchronizedList` blocks. It is not good to do so in reactive code.  
maybe it is better to replace it with `CopyOnWriteArrayList` or similar?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	@Override
+	public String getServiceId() {
+		return delegate.getServiceId();
+	}
+
+	@Override
+	public Flux<List<ServiceInstance>> get() {
+		if (!healthyInstances.isEmpty()) {
```
robotmrv: this check is performed on Assembly time  
probably it must be moved to the execution time  
e.g.   
```java
	return Flux.defer(() -> {
		List<ServiceInstance> it = healthyInstances;
		if (it.isEmpty()) {
			if (LOG.isWarnEnabled()) {
				LOG.warn("No verified healthy instances were found, returning all listed instances.");
			}
			it = instances;
		}
		return Flux.just(new ArrayList<>(it));//uses less operators.  
	});
}
```

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,136 @@
...
+
+		healthCheckFlux.subscribe(verifiedInstances -> {
+			healthyInstances.clear();
+			healthyInstances.addAll(verifiedInstances);
+		});
+	}
+
+	protected Flux<List<ServiceInstance>> healthCheckFlux() {
+		return Flux.create(emitter -> Schedulers
+				.newSingle("Health Check Verifier: " + getServiceId(), true)
```
robotmrv: for each instance of `HealthCheckServiceInstanceListSupplier` new `Scheduler` / `Thread` is created just for simple polling task  
it is not good if one wants to create several instances of `HealthCheckServiceInstanceListSupplier` as benefit of Reactive services is - reduced amount of `Thread`s.  
Maybe it is better to Schedule task on default `Schedulers` like `parallel()`, `single()` or just  create one static?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,136 @@
...
+			instances.clear();
+			instances.addAll(delegateInstances);
+		});
+
+		Flux<List<ServiceInstance>> healthCheckFlux = healthCheckFlux();
+
+		healthCheckFlux.subscribe(verifiedInstances -> {
+			healthyInstances.clear();
+			healthyInstances.addAll(verifiedInstances);
+		});
```
robotmrv: this is not atomic operation  
so check `if (!healthyInstances.isEmpty()) {` could pass and then return empty list in the Flux from `get()`  
healthyInstances better to change to `AtomicReference`

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	protected Flux<List<ServiceInstance>> healthCheckFlux() {
+		return Flux.create(emitter -> Schedulers
```
robotmrv: @OlgaMaciaszek @spencergibb  @ryanjbaxter 
IMO `Flux.create` does looks like WA  
This could be rewritten without usage of it and `Scheduler#schedulePeriodically` and more Thread safely 
```java
protected Flux<List<ServiceInstance>> healthCheckFlux() {
	return Flux.defer(() -> {
		List<ServiceInstance> result = new CopyOnWriteArrayList<>();
		List<Mono<ServiceInstance>> checks = new ArrayList<>();
		for (ServiceInstance instance : instances) {
			Mono<ServiceInstance> alive = isAlive(instance)
					.timeout(healthCheck.getInterval(), Mono.empty())//ensures that batch will be produced no longer than healthCheck.getInterval()
					.filter(it -> it)
					.map(check -> instance);
			checks.add(alive);
		}
		return Flux.merge(checks)//executes in parallel
				.map(alive -> {//will emit in on Each healthy instance with saving result of the current batch check 
					result.add(alive);
					return result;
				})
				.defaultIfEmpty(result);
	})
	.repeatWhen(restart -> restart.delayElements(healthCheck.getInterval()))//
	.delaySubscription(Duration.ofMillis(healthCheck.getInitialDelay()));
}

private void initInstances() {
	healthCheckDisposable = delegate.get()
			.doOnNext(delegateInstances -> {
				instances = Collections.unmodifiableList(new ArrayList<>(delegateInstances));
			})
			.thenMany(healthCheckFlux())//chains operations as they depends on each other
			.subscribeOn(Schedulers.parallel())//subscribes async 
			.subscribe(verifiedInstances -> healthyInstances.set(verifiedInstances));//atomic operation 
}

@Override
public void destroy() {
	this.healthCheckDisposable.dispose(); // disposes task so bean could be destroyed without leaks 
}
```

robotmrv:
@OlgaMaciaszek @ryanjbaxter @spencergibb   
I've created PR which addresses reported issues  
https://github.com/spring-cloud/spring-cloud-commons/pull/685  
please take a look

OlgaMaciaszek:
@robotmrv thanks for the PR. Sorry for not getting back to you earlier - was on vacation. I see it was already merged, so it will enter Hoxton.SR3.

robotmrv:
@OlgaMaciaszek no problem :smile:    
I've tried not to change your logic but since you implemented it initially it's probably worth taking a look at the PR before it is released

PR #684, opened 2020-02-04
Add info on SimpleDiscoveryClient in the docs. Remove redundant param.

OlgaMaciaszek:


docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -277,6 +277,22 @@ implementations provided by Spring Cloud, among others  `ConsulDiscoveryClient`,
...
+instance, that uses properties to get information on service and instances, will be used.
+
+The information about the available instances should be passed to via properties in the following format:
+`spring.cloud.discovery.client.simple.instances.service1[0].uri=http://s11:8080`, where
+`spring.cloud.discovery.client.simple.instances` is the common prefix, then `service1` stands
+for the ID of the service in question, while `[0]` indicates the index number of the instance
+(as visible in the example, indexes start with `0`), and then the value of `uri` is
+the actual URI under which the instance is available.
+
+WARN:: `SimpleDiscoveryClient` has been added for testing and demo purposes. We do not recommend
```
spencergibb: I don't agree with this. It's also always there in a composite with the lowest priority.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/simple/SimpleDiscoveryClientAutoConfiguration.java
```diff
@@ -68,7 +67,7 @@ public SimpleDiscoveryProperties simpleDiscoveryProperties() {
 	}
 
 	@Bean
-	@Order(Ordered.LOWEST_PRECEDENCE)
+	@Order
```
spencergibb: Why this change?

OlgaMaciaszek: This is not necessary as `Ordered.LOWEST_PRECEDENCE` is the default value.

docs/src/main/asciidoc/spring-cloud-commons.adoc
OlgaMaciaszek: Seemed to me that using a ServiceRegistry is way more production-appropriate than static input that does not track the state of the instances. However, can remove it.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/simple/SimpleDiscoveryClientAutoConfiguration.java
spencergibb: so it is!

docs/src/main/asciidoc/spring-cloud-commons.adoc
spencergibb: I think it has uses for external systems that can't register. Let's remove the warning. Other than that the docs are great!

OlgaMaciaszek: Done.

PR #685, opened 2020-02-07
Fixes issues in HealthCheckServiceInstanceListSupplier

robotmrv:
- WebClient response leaks
- potential Scheduled task leak
- rework polling to plain reactor operators
- remove non atomic operations on ServiceInstance lists

original issue gh-683
related to gh-629

### DIFF
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.loadbalancer.core;
 
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 
@@ -69,10 +70,10 @@ public class HealthCheckServiceInstanceListSupplier
 		this.defaultHealthCheckPath = healthCheck.getPath().getOrDefault("default",
 				"/actuator/health");
 		this.webClient = webClient;
-		this.aliveInstancesReplay = delegate.get()
+		this.aliveInstancesReplay = Flux.defer(delegate)
 				.delaySubscription(Duration.ofMillis(this.healthCheck.getInitialDelay()))
 				.switchMap(serviceInstances -> healthCheckFlux(serviceInstances)
-						.map(alive -> (List<ServiceInstance>) new ArrayList<>(alive))
+						.map(alive -> Collections.unmodifiableList(new ArrayList<>(alive)))
 				)
 				.replay(1)
 				.refCount(1);
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
@@ -27,7 +27,6 @@ import org.apache.commons.logging.LogFactory;
 import reactor.core.Disposable;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
-import reactor.core.scheduler.Schedulers;
 
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.cloud.client.ServiceInstance;
@@ -87,7 +86,6 @@ public class HealthCheckServiceInstanceListSupplier
 				}));
 
 		this.healthCheckDisposable = Flux.merge(instancesFlux.then(), verifiedInstancesFlux.then())
-				.subscribeOn(Schedulers.parallel())
 				.subscribe();
 	}
 
@@ -104,14 +102,13 @@ public class HealthCheckServiceInstanceListSupplier
 							}
 							return Mono.empty();
 						})
-						.timeout(this.healthCheck.getInterval(), Mono.fromSupplier(() -> {
+						.timeout(this.healthCheck.getInterval(), Mono.defer(() -> {
 							if (LOG.isDebugEnabled()) {
 								LOG.debug(String.format(
 										"The instance for service %s: %s did not respond for %s during health check",
 										instance.getServiceId(), instance.getUri(), this.healthCheck.getInterval()));
 							}
-							//returns Completion signal if null
-							return null;
+							return Mono.empty();
 						}))
 						.handle((isHealthy, sink) -> {
 							if (isHealthy) {
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
@@ -18,7 +18,6 @@ package org.springframework.cloud.loadbalancer.core;
 
 import java.time.Duration;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 
@@ -29,6 +28,7 @@ import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 import org.springframework.beans.factory.DisposableBean;
+import org.springframework.beans.factory.InitializingBean;
 import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.client.loadbalancer.reactive.LoadBalancerProperties;
 import org.springframework.http.HttpStatus;
@@ -45,7 +45,7 @@ import org.springframework.web.util.UriComponentsBuilder;
  * @since 2.2.0
  */
 public class HealthCheckServiceInstanceListSupplier
-		implements ServiceInstanceListSupplier, DisposableBean {
+		implements ServiceInstanceListSupplier, InitializingBean, DisposableBean {
 
 	private static final Log LOG = LogFactory
 			.getLog(HealthCheckServiceInstanceListSupplier.class);
@@ -58,11 +58,9 @@ public class HealthCheckServiceInstanceListSupplier
 
 	private final String defaultHealthCheckPath;
 
-	private volatile List<ServiceInstance> instances = Collections.emptyList();
-
-	private volatile List<ServiceInstance> healthyInstances = Collections.emptyList();
-
 	private Disposable healthCheckDisposable;
+
+	private Flux<List<ServiceInstance>> aliveInstancesReplay;
 
 	public HealthCheckServiceInstanceListSupplier(ServiceInstanceListSupplier delegate,
 			LoadBalancerProperties.HealthCheck healthCheck, WebClient webClient) {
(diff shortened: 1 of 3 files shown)


spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -74,32 +77,33 @@ public HealthCheckServiceInstanceListSupplier(ServiceInstanceListSupplier delega
...
-			instances.addAll(delegateInstances);
-		});
-
-		Flux<List<ServiceInstance>> healthCheckFlux = healthCheckFlux();
-
-		healthCheckFlux.subscribe(verifiedInstances -> {
-			healthyInstances.clear();
-			healthyInstances.addAll(verifiedInstances);
-		});
+		healthCheckDisposable = delegate.get().doOnNext(delegateInstances -> {
```
robotmrv: I do no see contract about `ServiceInstanceListSupplier#get()`   
`Flux` does it considered to be finite?

spencergibb: yes

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -74,32 +77,33 @@ public HealthCheckServiceInstanceListSupplier(ServiceInstanceListSupplier delega
...
-				}, healthCheck.getInitialDelay(), healthCheck.getInterval().toMillis(),
-						TimeUnit.MILLISECONDS),
-				FluxSink.OverflowStrategy.LATEST);
+		return Flux.defer(() -> {
+			List<ServiceInstance> result = new CopyOnWriteArrayList<>();
+
+			List<Mono<ServiceInstance>> checks = new ArrayList<>();
+			for (ServiceInstance instance : instances) {
+				Mono<ServiceInstance> alive = isAlive(instance)
+						.onErrorResume(throwable -> Mono.empty())
```
robotmrv: should some logs be added here?  
if so `debug` or `warn`?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -130,7 +134,14 @@ public String getServiceId() {
 				.uri(UriComponentsBuilder.fromUri(serviceInstance.getUri())
 						.path(healthCheckPath).build().toUri())
 				.exchange()
-				.map(clientResponse -> HttpStatus.OK.equals(clientResponse.statusCode()));
+				.flatMap(clientResponse -> clientResponse.releaseBody()
+						.thenReturn(HttpStatus.OK.equals(clientResponse.statusCode()))
```
robotmrv: should non standard codes be taken into account?  
if so - it is better to use `ClientResponse.rawStatusCode()`

spencergibb: sure

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
spencergibb: yes, debug is my first instinct.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
robotmrv: If it is finite (and assume that it has only one emit) what is the reason to return `Flux` instead of `Mono` from  `ServiceInstanceListSupplier#get`?  
If there is such contract maybe it is a good idea to document it, because it is not clear from current returning type

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -56,50 +59,77 @@
...
+								sink.next(instance);
+							}
+						});
+
+				checks.add(alive);
+			}
+			List<ServiceInstance> result = new CopyOnWriteArrayList<>();
+			return Flux.merge(checks).map(alive -> {
+				result.add(alive);
+				return result;
```
robotmrv: as far as `List` is mutated should it be emitted on any update or only on first?  
this could be rewritten to  
```java
List<ServiceInstance> result = new CopyOnWriteArrayList<>();
AtomicBoolean first = new AtomicBoolean(false);
return Flux.merge(checks).<List<ServiceInstance>>handle((alive, sink) -> {
	result.add(alive);
	if (first.compareAndSet(false, true)) {
		sink.next(result);
	}
})
.defaultIfEmpty(result);
```

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -40,10 +41,11 @@
  * {@link WebClient} to ping the <code>health</code> endpoint of the instances.
  *
  * @author Olga Maciaszek-Sharma
+ * @author Roman Matiushchenko
  * @since 2.2.0
  */
 public class HealthCheckServiceInstanceListSupplier
-		implements ServiceInstanceListSupplier {
+		implements ServiceInstanceListSupplier, DisposableBean {
```
robotmrv: I've tried to make minimal changes to the current logic but    
I've found testing of ping functionality a little bit inconvenient  
`HealthCheckServiceInstanceListSupplier` starts ping implicitly and eventually I test just another stream which uses common state...     
IMO It signals that this functionality must be either moved to a new class or just spitted into different methods.  
So I would like to make `HealthCheckServiceInstanceListSupplier` an `InitializingBean` , and move start of pings out of constructor   
and  
make this  
```java
Flux<List<ServiceInstance>> instancesFlux = this.delegate.get().doOnNext(delegateInstances -> {
	this.instances = Collections.unmodifiableList(new ArrayList<>(delegateInstances));
});
```
and 
```java 
Flux<List<ServiceInstance>> verifiedInstancesFlux = healthCheckFlux()
		.doOnNext((verifiedInstances -> {
			this.healthyInstances = verifiedInstances;
		}));
```
two distinct methods which are invoked from  `afterPropertiesSet()`  
does this make sense @spencergibb  @OlgaMaciaszek

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -56,50 +58,75 @@
...
 	private final String defaultHealthCheckPath;
 
-	private List<ServiceInstance> instances = Collections
-			.synchronizedList(new ArrayList<>());
+	private volatile List<ServiceInstance> instances = Collections.emptyList();
 
-	private List<ServiceInstance> healthyInstances = Collections
-			.synchronizedList(new ArrayList<>());
+	private volatile List<ServiceInstance> healthyInstances = Collections.emptyList();
+
```
robotmrv: could remove state with such changes  
```java
private Flux<List<ServiceInstance>> aliveInstancesReplay;
...
//in the constructor 
this.aliveInstancesReplay = this.delegate.get()
				.switchMap(serviceInstances -> {
					return healthCheckFlux(serviceInstances).map(alive -> alive.isEmpty() ? serviceInstances : alive);
				})
				.replay(1)
				.autoConnect()
				.onBackpressureLatest();
...
protected Flux<List<ServiceInstance>> healthCheckFlux(List<ServiceInstance> instances) {
...//the same logic but uses `instances` method arg instead of class field
}
...
public void afterPropertiesSet() {
    this.healthCheckDisposable = aliveInstancesReplay.subscribe();
}

//and `get()` become just 
@Override
	return aliveInstancesReplay;
}
``` 

less state and it is easier to test

robotmrv: downside - requires protected method signature change, but it was introduced recently so it is not a problem

robotmrv:
@spencergibb  @OlgaMaciaszek   
Changes are ready for review.  
Initial issue https://github.com/spring-cloud/spring-cloud-commons/issues/629 is closed  
Should it be reopened or should I create new issue?

spencergibb:
@robotmrv this PR is fine, no need to open anything else.

robotmrv:
@spencergibb @OlgaMaciaszek  
kindly reminder that this PR is ready for review

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -56,50 +58,72 @@
...
-					List<ServiceInstance> verifiedInstances = new ArrayList<>();
-					Flux.fromIterable(instances).filterWhen(this::isAlive)
-							.subscribe(serviceInstance -> {
-								verifiedInstances.add(serviceInstance);
-								emitter.next(verifiedInstances);
-							});
-				}, healthCheck.getInitialDelay(), healthCheck.getInterval().toMillis(),
-						TimeUnit.MILLISECONDS),
-				FluxSink.OverflowStrategy.LATEST);
+	protected Flux<List<ServiceInstance>> healthCheckFlux(List<ServiceInstance> instances) {
```
spencergibb: I'm unsure what to do here. This is a protected method that you have changed the signature. This is not backward compatible.

robotmrv: this class was not released yet, or am I missing something?

spencergibb: my bad you're right, class has since 2.2.0, but that is incorrect.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
bsideup: Why not just `Flux.merge(checks).collectList().defaultIfEmpty(emptyList())`?

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
```diff
@@ -56,50 +58,72 @@
...
+				.timeout(this.healthCheck.getInterval(), Mono.defer(() -> {
+					if (LOG.isDebugEnabled()) {
+						LOG.debug(String.format(
+								"The instance for service %s: %s did not respond for %s during health check",
+								instance.getServiceId(), instance.getUri(),
+								this.healthCheck.getInterval()));
+					}
+					return Mono.empty();
+				}))
+				.handle((isHealthy, sink) -> {
```
bsideup: can be replaced with `.filter(isHealthy -> isHealthy)`

robotmrv: @bsideup , not really.  
it can be replaced by  
```java
.filter(isHealthy -> isHealthy)
.map(it -> instance)
```
but is `filter` + `map` more efficient then `handle`? I thought `handle` is just for such cases

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java
robotmrv: this question was addressed to my intermediate implementation which mutated `healthyInstances` field.  
I was trying to preserve initial behavior.  
`.collectList()` will wait until all health checks are completed, in initial implementation `List` was emitted on the first healthy instance presence. As far as I understand it was made due to the fact that all health checks could take a long time (in current implementation it is `this.healthCheck.getInterval()` but in initial it could be `infinity`) but it should return healthy instances as soon as possible.

robotmrv:
@OlgaMaciaszek  
> Not sure why you've added the setup method in the test, though - AFAIK, works same with field instantiation.

it creates new instance for each test so previous test will not influence next one, also it is just to avoid instantiation in each test method.

>  Also, why are you adding a local variable assignment in `afterPropertiesSet()` and in `destroy()` methods? I assume it's for performance improvement?

It is not for performance reasons. I just dispose that was checked  

> Will, however, change some cosmetics things, such as unnecessary use of `this.`  

No problem it is redundant but just looks more clear to me, especially when I look at code on github. Recently `spring-security` team asked me to add `this.` in another PR, I thought you have similar conventions.

Issue #458, opened 2018-12-19
Endpoint ID 'service-registry' contains invalid characters


liuzhongkai:
2018-12-19 11:11:42.598  WARN 13196 --- [           main] o.s.boot.actuate.endpoint.EndpointId     : Endpoint ID 'service-registry' contains invalid characters, please migrate to a valid format.

liuzhongkai:
spring-boot-starter-actuator 2.1.1.RELEASE

spencergibb:
It is only a warning that can be ignored for now.

tanuja3088:
Facing same issue also unable to register my service to eureka it remains in STARTING status

aflaifel:
facing the same issue and can't inject any values from application.yml file

nintha:
if `EndpointId` includes '-' or '.', it prints this warning.

```
// org.springframework.boot.actuate.endpoint.EndpointId.java
private static final Pattern WARNING_PATTERN = Pattern.compile("[\\.\\-]+");
// ...
// other codes
// ...
if (WARNING_PATTERN.matcher(value).find()) {
    logWarning(value);
}
// ...
// lots of other codes
// ...
private static void logWarning(String value) {
    if (logger.isWarnEnabled() && loggedWarnings.add(value)) {
        logger.warn("Endpoint ID '" + value + "' contains invalid characters, please migrate to a valid format.");
    }
}
```

spencergibb:
> It is only a warning that can be ignored for now.

nintha:
> > It is only a warning that can be ignored for now.
Thanks for your reply. However, it is wired WARNING_PATTERN. Will it be fixed in the next version?

linux5396:
> > It is only a warning that can be ignored for now.
exc. why I see an Instance from eureka website after set register-with-eureka and fetch-registry into false?

antonkhvorov:
Any updates?

spencergibb:
No

ryanjbaxter:
There is an issue open in Boot to help address this problem https://github.com/spring-projects/spring-boot/issues/18148

ryanjbaxter:
We have communicated with the Boot team and they have decided that they will not remove support for ids with invalid characters until the next major release.  At that point we will migrate the ids to avoid any breaking changes in a minor release.

JWThewes:
Still not resolved with Hoxton...any plans on fixing this. I don't like warnings in my startup log ;)

[...]

PR #701, opened 2020-03-05
fix Endpoint ID 'service-registry' contains invalid characters, please migrate to a valid format

nkonev:
fix https://github.com/spring-cloud/spring-cloud-commons/issues/458


spring-cloud-commons/src/main/java/org/springframework/cloud/client/serviceregistry/endpoint/ServiceRegistryEndpoint.java
```diff
@@ -31,7 +31,7 @@
  * @author Spencer Gibb
  */
 @SuppressWarnings("unchecked")
-@Endpoint(id = "service-registry")
+@Endpoint(id = "serviceRegistry")
```
spencergibb: I think I want to go with just "serviceregistry" all lower case.

nkonev:
@spencergibb done

Issue #608, opened 2019-09-19
Rework PropertySourceBootstrapConfiguration to update the environment during environment post-processing and before ApplicationEnvironmentPreparedEvent is published


wilkinsona:
At the moment, `PropertySourceBootstrapConfiguration` is triggered by `BootstrapApplicationListener` on `ApplicationEnvironmentPreparedEvent` at which point it makes changes to the `Environment`. This breaks the expected lifecycle of the environment as no changes are expected after the environment has been prepared. It will hopefully be possible to correct this by switching to an `EnvironmentPostProcessor`. This should fix https://github.com/spring-projects/spring-boot/issues/18191. It should also allow the code that reinitialises the logging system – something that has caused numerous bugs over the years – to be removed.

If there's anything we can do in Boot or any assistance we can provide on the Cloud side to help to make this happen, please let us know.

spencergibb:
Ping @dsyer

dsyer:
Makes sense to look at it. I have a feeling that an `ApplicationListener` might need to be retained to do some sloshing of parent-child context relationships, but the bulk of the work with the `Environment` could go into an `EnvironmentPostProcessor`.

cschockaert:
Any update on this ticket? we are missing that feature a lot

spencergibb:
No

snicoll:
As part of fixing this issue, please document how a custom `EnvironmentPostProcessor` can run after the service configuration has been loaded from the config server. It should boil down to providing a sensible order value for it (perhaps Spring Boot itself would also have to clarify the ones that run by default). 

As far as I can see, it is impossible to write an `EnvironmentPostProcessor` that takes a decision based on what the user has configured when using the config server.

spencergibb:
This has been superseded by #703 which uses an EnvironmentPostProcessor. Bootstrap will now be opt in via `spring.config.use-legacy-processing=true`.

wilkinsona:
Awesome! Thanks, Spencer.

hnibbrig:
Hi. The linked issue [LoggerContext changes when using Spring Cloud, eliminating Logback metrics spring-projects/spring-boot#13077](https://github.com/spring-projects/spring-boot/issues/13077) was not resolved by the fix in #703 
Same for [LoggerContext changes when using Spring Cloud #564](https://github.com/micrometer-metrics/micrometer/issues/564).
@spencergibb should this be reopened or should I create a new issue?

[...]

PR #703, opened 2020-03-06
Use new boot ConfigData framework

spencergibb:
Bootstrap is now opt-in.

See #608 for original motivation.


spencergibb:
If you review post-merge, not a problem, I can address issues later since this is just a milestone.

dzou:
@spencergibb - Hey we have a library which provides a custom property source for bootstrap configuration when users add our `starter` dependency to their project.

Is bootstrap property sources being deprecated or is there an alternate way to access this feature? We are also wondering if we should instruct users that they should add `spring.config.use-legacy-processing=true` to their `application.properties` now or if we might be able to set the property to true in our starter dependency by default somehow.

ruben056:
Because bootstrapping is now, my client no longer connects to the config server (bootstrap properties with config server uri are no longer read).
Adding: "spring.cloud.bootstrap.enabled=true" to the application.properties does not help. When this property is evaluated (BootstrapApplicationListener.onApplicationEvent), using "Environment.getProperty()", it is always empty...
Same for property "spring.config.use-legacy-processing".
Putting these properties in bootstrap.properties does not make sense because it will not be read, but putting them in application.properties also does not work, what should be done here?
Also I see there is a way of bootstrapping if the  "org.springframework.cloud.bootstrap.marker.Marker" class is present, but it is not in the dependencies I have... (are you supposed to create it yourself?)

What is they new/correct way to have the client now about and connect to the config server after these changes?
I used the spring initializr to create the project with spring boot 2.4.0 (gradle). Only dependency was the "spring could config client", which was version "2020.0.0-M5".

ryanjbaxter:
@ruben056 see the documentation

https://docs.spring.io/spring-cloud-config/docs/3.0.0-M5/reference/html/#config-first-bootstrap

> It must be set as a System Property or environment variable.

ruben056:
@ryanjbaxter Thx for the info, I was looking at the "older" docs I guess (https://cloud.spring.io/spring-cloud-config/multi/multi__spring_cloud_config_client.html)

dzou:
@ruben056 - I think instead of setting the properties you can try using the new `spring-cloud-starter-bootstrap` to enable bootstrap; see: https://github.com/spring-cloud/spring-cloud-release/wiki/Spring-Cloud-2020.0-Release-Notes#breaking-changes

Adding this dependency to your project should have the equivalent effect of using `spring.config.use-legacy-processing`. I think for most applications its not practical to use the system property/environment variable route.

I think the "preferred" style is to migrate to the new `spring.config.import` style that is also discussed in the link. Think this should be in the new docs for config server.

PR #714, opened 2020-03-18
Adds support for calling Response.onComplete()

### LINKED ISSUE #675: Complete loadbalancer lifecycle
From https://github.com/spring-cloud/spring-cloud-commons/issues/674#issue-552954025

> I found `org.springframework.cloud.client.loadbalancer.reactive.Response.onComplete` method but it's never called and `CompletionContext` doesn't contain response object (which can be useful for extracting some additional stats)

Let's complete the lifecycle and then follow on with with server stats that would likely take advantage. of this.

spencergibb:
Fixes #675


spencergibb:
Superseded by #733

Issue #716, opened 2020-03-20
HealthCheckServiceInstanceListSupplierTests.shouldCancelSubscription() times out in CI


spencergibb:
```
Error Message
Assertion condition defined as a lambda expression in org.springframework.cloud.loadbalancer.core.HealthCheckServiceInstanceListSupplierTests that uses java.util.concurrent.atomic.AtomicInteger 
Expecting <AtomicInteger(0)> to have value:
  <1>
but did not within 10 seconds.
Stacktrace
org.awaitility.core.ConditionTimeoutException: 
Assertion condition defined as a lambda expression in org.springframework.cloud.loadbalancer.core.HealthCheckServiceInstanceListSupplierTests that uses java.util.concurrent.atomic.AtomicInteger 
Expecting <AtomicInteger(0)> to have value:
  <1>
but did not within 10 seconds.
	at org.springframework.cloud.loadbalancer.core.HealthCheckServiceInstanceListSupplierTests.shouldCancelSubscription(HealthCheckServiceInstanceListSupplierTests.java:488)
Caused by: java.lang.AssertionError: 

Expecting <AtomicInteger(0)> to have value:
  <1>
but did not.
	at org.springframework.cloud.loadbalancer.core.HealthCheckServiceInstanceListSupplierTests.lambda$shouldCancelSubscription$9(HealthCheckServiceInstanceListSupplierTests.java:489)
```
@robotmrv could you have a look?

spencergibb:
See d5dc4465b2291075cddc2fb5f5509a6af5febe4a for ignored test

PR #718, opened 2020-03-22
try to fix flaky test

robotmrv:
try to fix https://github.com/spring-cloud/spring-cloud-commons/issues/716  

cc @spencergibb


robotmrv:
Most likely the reason is that delegate is subscribed on parallel `Thread`. If main test `Thread` cancels health checks  before delegate was subscribed it could never be cancelled.

PR #719, opened 2020-03-25
Add a property source that produces a random value that is cached

ryanjbaxter:
This would allow the application to be refreshed and still get the same random value the next time the application needs one.

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/util/random/CachedRandomPropertySource.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/util/random/CachedRandomPropertySource.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2012-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.util.random;
+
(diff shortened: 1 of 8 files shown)


spring-cloud-context/src/main/java/org/springframework/cloud/util/random/CachedRandomPropertySource.java
```diff
@@ -0,0 +1,116 @@
...
+			randomValueCache.put(type, random);
+		}
+		return random;
+	}
+
+	private Map<String, Object> getCacheForKey(String key) {
+		if (logger.isDebugEnabled()) {
+			logger.debug("Looking in random cache for key: " + key);
+		}
+		Map<String, Object> randomValueCache = cache.get(key);
```
spencergibb: could be replaced by `computeIfAbsent()`

spring-cloud-context/src/main/java/org/springframework/cloud/util/random/CachedRandomPropertySource.java
```diff
@@ -0,0 +1,116 @@
...
+		extends PropertySource<RandomValuePropertySource> {
+
+	private static final String NAME = "cachedrandom";
+
+	private static final String PREFIX = NAME + ".";
+
+	private static Map<String, Map<String, Object>> cache = new ConcurrentHashMap<>();
+
+	public CachedRandomPropertySource(
+			RandomValuePropertySource randomValuePropertySource) {
```
spencergibb: Clever use, I was thinking of copying.

spring-cloud-context/src/main/java/org/springframework/cloud/util/random/CachedRandomPropertySource.java
```diff
@@ -0,0 +1,116 @@
...
+
+import org.springframework.boot.env.RandomValuePropertySource;
+import org.springframework.core.env.PropertySource;
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Ryan Baxter
+ */
+public class CachedRandomPropertySource
+		extends PropertySource<RandomValuePropertySource> {
```
spencergibb: `<CachedRandomPropertySource>`

PR #733, opened 2020-04-15
Lb complete lifecycle

OlgaMaciaszek:


adriancole:
was asked to take a look. I think there are some design considerations mostly around conflation of the callback with its parameters. It is easier to see if you look at the whole file here.

```java
public interface Response<T> {

	boolean hasServer();

	T getServer();

	/**
	 * Notification that the request completed.
	 * @param completionContext - completion context
	 */
	void onComplete(CompletionContext completionContext);

	/**
	 * Allows accessing the response's {@link CompletionContext}.
	 * @return completion context
	 */
	CompletionContext getCompletionContext();

}
```

### don't have a callback return itself

Putting "response" aside makes the callback related things easier to see. Typically a callback api won't include an accessor to the its callback parameters. 

Ex. instead of 
```java
public interface Callback{
	void onComplete(CompletionContext completionContext);
	CompletionContext getCompletionContext();
}
```

It would be..
```java
public interface Callback{
	void onComplete(CompletionContext completionContext);
}
```

Remember via composition or otherwise, you can catch and return what you wanted from the callback...
```
public class AwaitableCallback implements Callback {

  CompletionContext getCompletionContext() {
      this.countDown.await();
--snip--
```

### maybe onComplete should catch a response instead.
I don't know the full responsibilities here, but it looks like Response shouldn't also be a callback of its properties. Taken further, the callback should consider having the entire response, not just the completion context.

Ex. you could possibly clear everything up by separating the two.. they can still be in
the same type, just an inner type instead.

```java
public interface Response<T> {
       public interface Callback<T> {
           void onComplete(Response<T> response);
       }

	boolean hasServer();

	T getServer();

	CompletionContext getCompletionContext();
	@Nullable Throwable getThrowable();
}
```

Anyway this is drive-by but hope it helps

adriancole:
slightly unrelated, but in zipkin brave, we finally decided a response should also try to return its request. In LB, this is important as you can know the last request attempt which might not be what the user thought it was!

some hints:

https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpResponse.java#L40
https://github.com/openzipkin/brave/blob/master/instrumentation/http/RATIONALE.md#why-does-httphandlerhandlefinish-have-no-error-parameter

OlgaMaciaszek:
Thanks, @adriancole - this is helpful. I agree that we could have a separate `Callback` interface with `onComplete()` taking the `Response` parameter as in the last code snippet. Then, maybe we could set the value of a `completionContext` field (taken from `Response`) during the `onComplete()` call and then the stats processor could just call `getComletionContext()` to further process it. @spencergibb wdyt?

OlgaMaciaszek:
I also agree that being able to access the `Request` from the `Response` might add valuable information to whoever handles the data. We should probably discuss the possibility of doing it in LB as well.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/DefaultRequestContext.java
```diff
@@ -29,9 +31,18 @@
 	 */
 	private String hint = "default";
 
+	/**
+	 * The original load-balanced request.
+	 */
+	private Object originalRequest;
```
OlgaMaciaszek, line 37: In our usage via the `BlockingLoadBalancerClient` and `ReactorLoadBalancerExchangeFilterFunction` it's actually always `ClientResponse`, but on the interface level, it could be any object. Have tried using a generic value here, but due to similar issues as [in `CompletionContext`](https://github.com/spring-cloud/spring-cloud-commons/pull/733/files#r434712895), have settled for `Object`.

OlgaMaciaszek:
Closing in favour of https://github.com/spring-cloud/spring-cloud-commons/pull/783

PR #739, opened 2020-04-19
The configuration metadata default value for management.endpoint.env.post.enabled should be false instead of true

### LINKED ISSUE #738: Spring Cloud Configuration metadata for `management.endpoint.env.post.enabled` is true but should be false
Since #681 the default value of `management.endpoint.env.post.enabled` is false.

But the (additional) Spring Cloud Configuration Metadata describes the default value is `true`. This is confusing because so my IDEA 'tells' me the default value is `true` while it is actually `false`.

additional-spring-configuration-metadata.json: 

```
{
  "name": "management.endpoint.env.post.enabled",
  "type": "java.lang.Boolean",
  "description": "Enable changing the Environment through a POST to /env.",
  "defaultValue": true
}
```

j-tim:
Fixes #738

### DIFF
--- a/spring-cloud-commons/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-cloud-commons/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -43,7 +43,7 @@
 			"type": "java.lang.Boolean"
 		},
 		{
-			"defaultValue": true,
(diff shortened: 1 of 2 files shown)


spencergibb:
This should be made on the 2.2.x branch

j-tim:
Ok no problem I'll create a new pull request.

Issue #660, opened 2019-12-18
A simplified way of adjusting Spring Cloud LoadBalancer config


OlgaMaciaszek:
Think of an easier way of switching between `ServiceInstanceListSupplier` and `ReactorLoadBalancer` implementations, without the users having to create hierarchies of delegate beans in custom configs.

spencergibb:
A couple of ideas: a few predefined hierarchies enabled by a property, or a property that is a list of classes and we enable to hierarchy using reflection.

OlgaMaciaszek:
Adding a builder API and preselected configurations backed by properties.

OlgaMaciaszek:
Done.

PR #741, opened 2020-04-21
Lb custom configuration methods

OlgaMaciaszek:
Part of fix for gh-660.


spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/annotation/configutil/ServiceInstanceListSuppliers.java
```diff
@@ -0,0 +1,87 @@
...
+import org.springframework.web.reactive.function.client.WebClient;
+
+/**
+ * Utility class containing helper methods for setting up
+ * {@link ServiceInstanceListSupplier} configuration in a more concise way.
+ *
+ * @author Olga Maciaszek-Sharma
+ * @since 2.2.3
+ */
+public final class ServiceInstanceListSuppliers {
```
spencergibb, line 41: This one feels like a true builder is needed, since you will start with discovery client, then add zone preference or health chech and cache later.

OlgaMaciaszek:
Closing in favour of https://github.com/spring-cloud/spring-cloud-commons/pull/751.

Issue #611, opened 2019-09-21
Replace CompositePropertySource with one PropertySource per file


philwebb:
See https://github.com/spring-projects/spring-boot/issues/18174 for background and a sample application.

Raising an issue here as requested by @spencergibb

ewiegs4:
Any chance of someone looking into this soon? I could take a stab at it if not but haven't worked in this code base before so it could take some time to get my bearings.

ryanjbaxter:
@ewiegs4  if you want to take a shot at fixing the issue that would be great!  @TYsewyn can help if you have any questions.  Since the problem was in Greenwich as well, make sure you submit any PRs against the 2.1.x branch.

ryanjbaxter:
@philwebb can you point us to how this behavior is implemented in Boot?  We have multiple property sources due to the profile being enabled.  We would want to combine them into a composite, but how come `CompositeProperty` source does not implement the expected list behavior (overriding the list property)?

philwebb:
The [IndexedElementsBinder](https://github.com/spring-projects/spring-boot/blob/v2.2.1.RELEASE/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/IndexedElementsBinder.java#L70) class is the main area to look. Specifically, the `bindIndexed` method limits itself to one source at a time and returns as soon as binding works.

In Spring Boot 1.x we found that binding lists with profile specific files was quite confusing because it wasn't always clear when elements would be overridden (see [this comment for an example](https://github.com/spring-projects/spring-boot/issues/9137#issuecomment-300965005)).

In `ConfigFileApplicationListener` we register one `PropertySource` per file (see the [addLoadedPropertySources](https://github.com/spring-projects/spring-boot/blob/v2.2.1.RELEASE/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigFileApplicationListener.java#L674) method) which means they are adapted directly to individual `ConfigurationPropertySource` instances used by the `Binder`.

The `Binder` doesn't treat the `CompositePropertySource` in any special way, it's just a regular `PropertySource` as far as it's concerned. Because of this, the indexed binder will consider it as a single source and the early exit rule isn't being applied.

ryanjbaxter:
I have confirmed that the same problem happens with just a normal config client (as opposed to using spring.cloud.config.server.bootstrap=true).

[...]

PR #742, opened 2020-04-27
Spring Cloud Config client don't support removing PropertySource

obourgain:
We saw a regression in the Spring Cloud Config client.
We use it with Spring Cloud Config Server servicing files hosted in a
Git repository. We sometimes need to remove some configuration file from
the repository.
Since https://github.com/spring-cloud/spring-cloud-commons/issues/611
was merged, our applications can not detect when such a file is deleted.

The explanation is that before #611, the PropertySources (PS) were
wrapped in a CompositePropertySource, and then ContextRefresher.addConfigFilesToEnvironment() would
match the names and replace the old CompositePropertySource wrapping all
the PS including the one that was removed with the new CompositePropertySource that do not contains it, and all is fine.

After #611, we have a list of PS, and we would only update the existing
PS or add new ones but never detect that a PS was removed.

Here is a proposed fix, but any suggestion or modifications are welcome.
What is probably missing is a filter to only remove PropertySources that where loaded from the Config Server, as in the current state this PR causes integration test RefreshEndpointIntegrationTests to fail.

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/ContextRefresher.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/ContextRefresher.java
@@ -140,6 +140,19 @@ public class ContextRefresher {
 					}
 				}
 			}
+
+			MutablePropertySources newPropertySources = environment.getPropertySources();
+			List<String> toRemove = new ArrayList<>();
+			// this is a CopyOnWriteArrayList, it doesn't support Iterator.remove
+			for (PropertySource<?> propertySource : target) {
+					String name = propertySource.getName();
+					if (!newPropertySources.contains(name)) {
+						toRemove.add(name);
+					}
+			}
+			for (String name : toRemove) {
+				target.remove(name);
+			}
 		}
 		finally {
 			ConfigurableApplicationContext closeable = capture;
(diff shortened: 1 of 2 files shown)


spencergibb:
Yes, I think so.

obourgain:
Hello, I updated the PR to target branch 2.2.x.
I would also like to have this fix on 2.1.x as we are still using that in production and still have a few 1.5.x to update before going to 2.2.x

Also I would like to have some help about the failing test. It seems that my fix will remove too many configurations in some cases.

spencergibb:
2.1.x isn't an active branch

spencergibb:
Unfortunately this breaks some tests. You'll have to check them and and see why.

biergit:
Could someone take a look at this again please?

jawher:
@biergit We ended up copy/pasting the `ContextRefresher` in our codebase and applying this PR's patch (we have some projects still stuck with Spring Boot 2.1).

biergit:
The change doesn't work for me. Too many PropertySources are removed then (e,g, it seems all PropertySources included via @PropertySource)

biergit:
Instead of removing the last PropertySource on server side we just add an empty one. Works for our case.

obourgain:
we are not interested in this PR anympre, closing

Issue #595, opened 2019-08-29
Add ServiceInstanceListSupplier


elandau:
Would it be possible to add support for more advanced load balancing algorithms other than simple round robin.  Specifically, it would be great to see an implementation of Power of Two Choices.  There's a great description of the algorithm here, https://www.nginx.com/blog/nginx-power-of-two-choices-load-balancing-algorithm/.

To be able to support such algorithms we'd need to make a slight change to how the load balancer tracks ServiceInstances.  Instead of a ServiceInstanceSupplier I'd suggest creating a ServiceInstanceListSupplier (returning Flux<List<ServiceInstance>>).  This would allow the load balancer to reactively update its internal state (i.e add or remove ServerInstances from the pool) outside of the load balancer decision.  With the current Flux<ServiceInstance> this state could only be maintained as part of the load balancer decision which could be very costly.  This approach has the added benefit of not having to create a List<ServerInstance> for each load balancer decision.  The list of ServerInstance would only be updated when there was an actual change to the target cluster. 

Looking at the code I was also not able to find any places where Response#onComplete is actually invoked, other than unit tests.  This would also need to be fully wired up so that the algorithm can properly track the number of pending requests.

OlgaMaciaszek:
We will provide the `ServiceInstanceListSupplier` within this issue. Have created a new one for providing the Power of Two Choices implementation: https://github.com/spring-cloud/spring-cloud-commons/issues/601 .

PR #753, opened 2020-05-09
Remove deprecated ServiceInstanceSupplier

fitzoh:
While familiarizing myself with the codebase, I noticed that `ServiceInstanceSupplier` was deprecated in August 2019 (#595), and also noticed that the current version is 3.0.0-SNAPSHOT.

Sounds like a good time for a deprecation cleanup to me.


spring-cloud-loadbalancer/src/test/java/org/springframework/cloud/loadbalancer/core/LoadBalancerTests.java
```diff
@@ -123,11 +122,11 @@ public void emptyHosts() {
 	@Test
 	public void staticConfigurationWorks() {
```
m1ngyuan, line 123: Duplicate test case staticconfigurationworkswithserviceinstancelistsupplier

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/annotation/LoadBalancerClientConfiguration.java
```diff
@@ -76,23 +70,6 @@ public ServiceInstanceListSupplier discoveryClientServiceInstanceListSupplier(
 					.withCaching().build(context);
 		}
 
-		@Bean
```
m1ngyuan, line 79: I think it's just to replace the deprecated class, not to delete it completely

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/annotation/LoadBalancerClientConfiguration.java
```diff
@@ -109,23 +86,6 @@ public ServiceInstanceListSupplier discoveryClientServiceInstanceListSupplier(
...
```
m1ngyuan, line 112: I think it's just to replace the deprecated class, not to delete it completely

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/annotation/LoadBalancerClientConfiguration.java
m1ngyuan, line 79: I think it's just to replace the deprecated class, not to delete it completely

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/annotation/LoadBalancerClientConfiguration.java
m1ngyuan, line 112: I think it's just to replace the deprecated class, not to delete it completely

spencergibb: The class is already deprecated in 2.2.x and can be removed in master.

m1ngyuan: But we still need serviceinstancelistsupplier bean. We can modify it in this way:
@Bean
		@ConditionalOnBean(ReactiveDiscoveryClient.class)
		@ConditionalOnMissingBean
		public ServiceInstanceListSupplier discoveryClientServiceInstanceSupplier(
				ReactiveDiscoveryClient discoveryClient, Environment env,
				ApplicationContext context) {
			DiscoveryClientServiceInstanceListSupplier delegate = new DiscoveryClientServiceInstanceListSupplier(
					discoveryClient, env);
			ObjectProvider<LoadBalancerCacheManager> cacheManagerProvider = context
					.getBeanProvider(LoadBalancerCacheManager.class);
			if (cacheManagerProvider.getIfAvailable() != null) {
				return new CachingServiceInstanceListSupplier(delegate,
						cacheManagerProvider.getIfAvailable());
			}
			return delegate;
		}

spencergibb: That is already in this class

spencergibb: no

Issue #689, opened 2020-02-18
LoadBalancer: Add Sticky LB implementation.


OlgaMaciaszek:
To allow for functional backwards compatibility for SC Zookeeper.

fitzoh:
Any thoughts/directions on implementation on this?

Cookie/session based routing vs client IP hashing?
Some other approach?

neelimaj97:
Any inputs on when it will get implemented

OlgaMaciaszek:
@neelimaj97 planning to have it for the 2020.0.0 GA release. 

@fitzoh I was thinking about an implementation of `ReactorServiceInstanceLoadBalancer` that during the first invocation would choose an instance, set it in a variable and later redirects all the calls to that same instance, similarly to the [original Zookeeper implementation](https://github.com/spring-cloud/spring-cloud-zookeeper/blob/2.2.x/spring-cloud-zookeeper-discovery/src/main/java/org/springframework/cloud/zookeeper/discovery/dependency/StickyRule.java). It's a good time to discuss approaches, however, so if you think a different approach would be better, feel free to outline it here for discussion. Would you be interested in submitting a PR?

neelimaj97:
@OlgaMaciaszek Shouldnt only the requests in the same session go to the same instance? So if different users are accessing the same app, it would go to different instances rather than loading on one same instance

fitzoh:
That implementation does not appear to align with my assumptions about what sticky load balance behavior entails either.

The implementation above appears to say that all requests (regardless of the user making them) will end up at the same backend, with no requests hitting other backens until it rotates.

I was expecting an implementation that would ensure that a given user always ends on the same server, but that traffic is (somewhat) evenly distributed amongst the available backends.

andaziar:
Hi Olga! This kind of functionality is what we tried to achieve with our suggestion in #672 and our definition of "sticky" load balancing is in line with what @fitzoh said previously: "a given user always ends on the same server". However I would like to add that the user might be "re-balanced" to another service instance if the instance i.e. goes down. 

[...]

PR #764, opened 2020-05-23
POC for a session/cookie based sitcky load balancer implementation.

fitzoh:
Possible implementation for #689, meant as a discussion starter (no tests, not wired up to anything).

# How it works

The `StickySessionLoadBalancer` looks for a server id in the `scg-instance-id` cookie (not married to that name).

If a cookie value is found and that value is a valid service instance ID, the request is routed to that service instance.

If it is not found or if the value is invalid (perhaps the server is no longer active or unhealthy), then a delegate `ReactorServiceInstanceLoadBalancer` and the request is routed to that server instead.

Regardless of the path taken, the cookie is updated with the instance ID that the request was routed to.

# Assumptions/Requirements

This implementation assumes that caller will pass a `ServerWebExchange` as the `Request` `Context`.

It is also dependent on `ServiceInstance`s having valid IDs (the field is nullable, so it might not be available, right?)

# Rough spots

I wasn't seeing any existing context usage, so I just took a stab at it.  Casting from an `Object` feels a little gross

# Future work

Make the cookie name configurable
Make the way the cookie value is customized (people may want to encrypt/hash/do something to not expose server IDs)

### DIFF
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/StickySessionLoadBalancer.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/StickySessionLoadBalancer.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2012-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.loadbalancer.core;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import reactor.core.publisher.Mono;
+
+import org.springframework.beans.factory.ObjectProvider;
+import org.springframework.cloud.client.ServiceInstance;
+import org.springframework.cloud.client.loadbalancer.reactive.DefaultResponse;
+import org.springframework.cloud.client.loadbalancer.reactive.EmptyResponse;
+import org.springframework.cloud.client.loadbalancer.reactive.Request;
+import org.springframework.cloud.client.loadbalancer.reactive.Response;
+import org.springframework.http.HttpCookie;
+import org.springframework.http.ResponseCookie;
+import org.springframework.web.server.ServerWebExchange;
+
+/**
+ * A session cookie based implementation of {@link ReactorServiceInstanceLoadBalancer}
+ * that ensures requests from the same client are routed to the same server.
+ *
+ * @author Andrew Fitzgerald
+ * @author Spencer Gibb
+ * @author Olga Maciaszek-Sharma
+ */
+public class StickySessionLoadBalancer implements ReactorServiceInstanceLoadBalancer {
+
+	private static final Log log = LogFactory.getLog(StickySessionLoadBalancer.class);
+
+	private final ReactorServiceInstanceLoadBalancer delegate;
+
+	private final String serviceId;
+
+	// TODO make this configurable
+	private final String cookieName = "scg-instance-id";
+
+	private ObjectProvider<ServiceInstanceListSupplier> serviceInstanceListSupplierProvider;
+
+	/**
+	 * @param serviceInstanceListSupplierProvider a provider of
+	 * {@link ServiceInstanceListSupplier} that will be used to get available instances
+	 * @param serviceId id of the service for which to choose an instance
+	 */
+	public StickySessionLoadBalancer(ObjectProvider<ServiceInstanceListSupplier> serviceInstanceListSupplierProvider,
+			String serviceId) {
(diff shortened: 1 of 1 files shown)


andaziar:
Nice and tidy PoC @fitzoh . It was easy to follow 👍 . Some thoughts:

I can imagine that in some architectures other consumers will not be aware of other service instances, only the service id. I'm saying this as service instances are a bit ephemeral and consumers shouldn't need to know what service instance they should point to. Their only outlet should be the load balancer.

The strategy for checking what service instance to "stick to" should maybe be configurable so that any use-case can be plugged in? Some developers maybe have `user_id`'s or `affinity_key` in the incoming requests and will have to perform HTTP requests towards every service instance to check if they `own` that specific id or key.

Regarding the usage of context in this case, I think one needs to make changes to the `org.springframework.cloud.client.loadbalancer.reactive.ReactorLoadBalancerExchangeFilterFunction` which intercepts the incoming request. In there, you would have to extract the appropriate cookie/header/whatever from the incoming `ClientRequest` and put it inside the `Context`object. One way would be doing it like this:

```
        Request<?> lbRequest = (request) -> request.attribute("affinity_key")
                .map((Function<Object, DefaultRequest<?>>) DefaultRequest::new)
                .orElse(new DefaultRequest<>());
```
So if there is an attribute value mapped to key `affinity_key` it will put that value in Context and create a new Request<> with that context.

And then the `ReactorLoadBalancerExchangeFilterFunction` would have to first fetch the appropriate load balancer for the specified `serviceId`:

`ReactiveLoadBalancer<ServiceInstance> loadBalancer = loadBalancerFactory.getInstance(serviceId);`

and finally shoot your new `lbRequest` like this:

`loadBalancer.choose(request)`

which will invoke `StickySessionLoadBalancer#choose(Request)`

If you want I could make some changes to your PoC, but would like to hear your opinions on this first :)

fitzoh:
>Nice and tidy PoC @fitzoh . It was easy to follow 👍

Thanks @andaziar!

>I can imagine that in some architectures other consumers will not be aware of other service instances, only the service id. I'm saying this as service instances are a bit ephemeral and consumers shouldn't need to know what service instance they should point to. Their only outlet should be the load balancer.
>The strategy for checking what service instance to "stick to" should maybe be configurable so that any use-case can be plugged in? Some developers maybe have user_id's or affinity_key in the incoming requests and will have to perform HTTP requests towards every service instance to check if they own that specific id or key.

I'm struggling a little bit to follow this... For reference, I'm not super familiar with the library and haven't used it directly (I'm primarily here because I saw https://github.com/spring-cloud/spring-cloud-gateway/issues/1176#issuecomment-617648408 and have a use case for sticky session in spring-cloud-gateway)

>Regarding the usage of context in this case, I think one needs to make changes to the `org.springframework.cloud.client.loadbalancer.reactive.ReactorLoadBalancerExchangeFilterFunction` which intercepts the incoming request. In there, you would have to extract the appropriate cookie/header/whatever from the incoming `ClientRequest` and put it inside the `Context` object.

One caveat here... It wouldn't be sufficient to just grab the value out of the cookie/header, we also need to be able to propagate the selected header back to the user.

>If you want I could make some changes to your PoC, but would like to hear your opinions on this first :)

No objections here, but I'm definitely interested to hear what @OlgaMaciaszek has to say

OlgaMaciaszek:
Hi @fitzoh @andaziar, sorry for getting back to you with a delay.
Thanks for the first draft @fitzoh . It's a good start, however, I agree with some remarks pointed out by @andaziar:
- I think we should not assume that the client already knows the instance id/ key; so if they do, we should take it from the cookies, but if not, we should pick an instance and propagate the id to the cookies
- This will, indeed, require modifying `ReactorLoadBalancerExchangeFilterFunction` to pass this value both ways, however, rather than just passing along that cookie value, I think it makes sense to start propagating the entire `ClientRequest` and `ClientResponse` objects (this has also been suggested in https://github.com/spring-cloud/spring-cloud-commons/pull/733#issuecomment-622670619) 

Apart from this, I think we can handle the cookies using `ClientRequest` objects that get passed via the `LoadBalancerExchangeFilterFunction` rather than `ServerWebExchange`.

I have created the first draft of possible `LoadBalancerExchangeFunction` changes so that we can discuss it here: https://github.com/spring-cloud/spring-cloud-commons/pull/769. Let me know what you think?

Also, I'm going back to finishing up this PR now: https://github.com/spring-cloud/spring-cloud-commons/pull/733 - will include the changes with passing along the `ClientRequest` and `ClientResponse` there. After the final version of that is merged, you could resubmit your sticky session PR as well.

CC @spencergibb @andersenleo

OlgaMaciaszek:
@fitzoh you can also keep an eye on https://github.com/spring-cloud/spring-cloud-commons/pull/733 as the discussion about modifying which data is going to be passed along through requests and responses will probably take place there.

varna909:
Hi , 

Could you please let me know if the sticky session implementation is complete?Is there any docs on the usage?

OlgaMaciaszek:
Hi, @fitzoh Would you like to continue working on this? If yes, please merge the current `master` (that now passes the `ClientRequest` object to the LB via `ReactorLoadBalancerExchangeFilterFunction`, switch your implementation to using the requests and add some tests. If you don't have time to work on it now, please let me know - I will get it done.

fitzoh:
Interested in picking it back up, not sure on the timeline.
if it was something you were going to prioritize I would say go for it.

OlgaMaciaszek:
Ok @fitzoh, could you give me some kind of estimate? If it's up to a month, we could wait. If it's longer, I'd rather get it done earlier. Let me know.

fitzoh:
>Ok @fitzoh, could you give me some kind of estimate? If it's up to a month, we could wait. If it's longer, I'd rather get it done earlier. Let me know.

Sorry for the delay @OlgaMaciaszek , I'd so go ahead and implement it

Issue #760, opened 2020-05-21
HealthCheck causes calls to hang


OlgaMaciaszek:
Bug can be reproduced in this sample:

https://github.com/OlgaMaciaszek/spring-cloud-example

Started after changes from: 
https://github.com/spring-cloud/spring-cloud-commons/commit/6db6a81d82e4b8fea8e3678cbd3500a61bd37312

OlgaMaciaszek:
CC @robotmrv

robotmrv:
@OlgaMaciaszek  
Hi, reason of hanging is `CachingServiceInstanceListSupplier` it tries to cache infinite `Flux` from and collect it to  `Mono<List<Signal<List<ServiceInstance>>>>` see 
https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/CachingServiceInstanceListSupplier.java#L71

I think it is redundant as `HealthCheckServiceInstanceListSupplier` caches latest result

spencergibb:
The point of `CachingServiceInstanceListSupplier` is caching.

robotmrv:
to be  more correct, Flux is collected into List in https://github.com/reactor/reactor-addons/blob/master/reactor-extra/src/main/java/reactor/cache/CacheFlux.java#L103 

one more thing  
There is no call of `HealthCheckServiceInstanceListSupplier#afterPropertiesSet` in 
https://github.com/OlgaMaciaszek/spring-cloud-example/blob/fixed-config/microservice-b/src/main/java/com/example/microserviceb/MicroserviceBApplication.java#L47  
which will allow caching on `HealthCheckServiceInstanceListSupplier` side  

> The point of CachingServiceInstanceListSupplier is caching.

yes, but you can't cache infinite stream, you need to stop at some time.  
`Flux` is potentially infinite, so if you want to cache signals this way you you need to get first `List<ServiceInstance>` in cache.

robotmrv:
Btw, `Flux` infinity was discussed [here](https://github.com/spring-cloud/spring-cloud-commons/pull/685#discussion_r376490682). I understood that contract of `ServiceInstanceListSupplier`  allows infinite streams

spencergibb:
ok, so what is cached may need to be adjusted.

OlgaMaciaszek:
> one more thing
[...]

PR #765, opened 2020-05-26
Gh 760 health check with cache new

OlgaMaciaszek:
Fixes gh-760.

### DIFF
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/CachingServiceInstanceListSupplier.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/CachingServiceInstanceListSupplier.java
@@ -69,7 +69,7 @@ public class CachingServiceInstanceListSupplier
 				return Mono.empty();
 			}
 			return Flux.just(list).materialize().collectList();
-		}, delegate.getServiceId()).onCacheMissResume(delegate)
+		}, delegate.getServiceId()).onCacheMissResume(delegate.get().take(1))
 				.andWriteWith((key, signals) -> Flux.fromIterable(signals).dematerialize()
 						.doOnNext(instances -> {
 							Cache cache = cacheManager
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/DiscoveryClientServiceInstanceListSupplier.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/DiscoveryClientServiceInstanceListSupplier.java
@@ -41,20 +41,21 @@ public class DiscoveryClientServiceInstanceListSupplier
 
 	private final String serviceId;
 
-	private final Flux<ServiceInstance> serviceInstances;
+	private final Flux<List<ServiceInstance>> serviceInstances;
 
 	public DiscoveryClientServiceInstanceListSupplier(DiscoveryClient delegate,
 			Environment environment) {
 		this.serviceId = environment.getProperty(PROPERTY_NAME);
 		this.serviceInstances = Flux
 				.defer(() -> Flux.fromIterable(delegate.getInstances(serviceId))
-						.subscribeOn(Schedulers.boundedElastic()));
+						.collectList().flux().subscribeOn(Schedulers.boundedElastic()));
 	}
 
 	public DiscoveryClientServiceInstanceListSupplier(ReactiveDiscoveryClient delegate,
 			Environment environment) {
 		this.serviceId = environment.getProperty(PROPERTY_NAME);
-		this.serviceInstances = Flux.defer(() -> delegate.getInstances(serviceId));
+		this.serviceInstances = Flux
+				.defer(() -> delegate.getInstances(serviceId).collectList().flux());
 	}
 
 	@Override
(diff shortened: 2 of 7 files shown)


spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/DiscoveryClientServiceInstanceListSupplier.java
```diff
@@ -41,20 +41,21 @@
...
-	private final Flux<ServiceInstance> serviceInstances;
+	private final Flux<List<ServiceInstance>> serviceInstances;
 
 	public DiscoveryClientServiceInstanceListSupplier(DiscoveryClient delegate,
 			Environment environment) {
 		this.serviceId = environment.getProperty(PROPERTY_NAME);
 		this.serviceInstances = Flux
-				.defer(() -> Flux.fromIterable(delegate.getInstances(serviceId)))
-				.subscribeOn(Schedulers.boundedElastic());
+				.defer(() -> Flux.fromIterable(delegate.getInstances(serviceId))
```
robotmrv, line 50: @OlgaMaciaszek  
I see here two problems: 
1. Blocking `DiscoveryClient.getInstances` executes before switch to `Schedulers.boundedElastic()`.    `.subscribeOn(Schedulers.boundedElastic())` should be applied to `Flux.defer()` - IMO this is critical  
2. `Flux.fromIterable(delegate.getInstances(serviceId)).collectList().flux()` could be simplified to `Flux.defer(() -> Flux.just(delegate.getInstances(serviceId)))`  

so final result could look like 
```java
this.serviceInstances = Flux
		.defer(() -> Flux.just(delegate.getInstances(serviceId)))
		.subscribeOn(Schedulers.boundedElastic());
```

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/CachingServiceInstanceListSupplier.java
```diff
@@ -69,7 +69,7 @@ public CachingServiceInstanceListSupplier(ServiceInstanceListSupplier delegate,
 				return Mono.empty();
 			}
 			return Flux.just(list).materialize().collectList();
-		}, delegate.getServiceId()).onCacheMissResume(delegate)
+		}, delegate.getServiceId()).onCacheMissResume(delegate.get().take(1))
```
robotmrv, line 72: @OlgaMaciaszek   
I have to notice that `HealthCheckServiceInstanceListSupplier` + `CachingServiceInstanceListSupplier` interaction will produce behavior:   
`HealthCheckServiceInstanceListSupplier` first emit will always contain maximum one service instance (and just empty list in case of all services instances are unhealthy) as it emits on first confirmed healthy Instance.  
So `CachingServiceInstanceListSupplier` will return only one healthy instance for first cached value.  

Another possible problem. If we have 3 healthy services `HealthCheckServiceInstanceListSupplier` will emit:  
```
[instace1]  (1)
[instace1, instace2] (2)
[instace1, instace2, instace3] (3)
```
`HealthCheckServiceInstanceListSupplier` caches only last emit  
and repeats this sequence by default every `30 seconds` (see `LoadBalancerProperties.HealthCheck#getInterval`)  
but `CachingServiceInstanceListSupplier` by default caches for `30 seconds` (see `LoadBalancerCacheProperties#ttl`) 
and `HealthCheckServiceInstanceListSupplier` (1), (2) and (3) emits are distributed in time with some delay. 
So if `LoadBalancerProperties.HealthCheck#getInterval` and `LoadBalancerCacheProperties#ttl` are the same or similar there is a possibility that `CachingServiceInstanceListSupplier` could cache emit (1), but not (2) or (3) because its caching `phase` is aligned with `HealthCheckServiceInstanceListSupplier` repeat `phase`.  
This is why I pointed that `HealthCheckServiceInstanceListSupplier` + `CachingServiceInstanceListSupplier` is redundant. It is enough `HealthCheckServiceInstanceListSupplier` as it caches last result with healthy instances.  
In case of both usage it is better to  desynchronize their "phases"

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/DiscoveryClientServiceInstanceListSupplier.java
OlgaMaciaszek: You're right. Thanks. Fixing it.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/CachingServiceInstanceListSupplier.java
OlgaMaciaszek: @robotmrv 
> This is why I pointed that `HealthCheckServiceInstanceListSupplier` + `CachingServiceInstanceListSupplier` is redundant. It is enough `HealthCheckServiceInstanceListSupplier` as it caches last result with healthy instances.

The problem here is that we need to have a consistent solution. Even though it might be better to cache the instances in each of the suppliers similarly to what you've done (which is something that would have to be discussed, as users might prefer leveraging the caching libraries and their elastic config), we would have to then change it for all the suppliers and do it in a major release only. I can (and will) add information in docs about the fact that it can be used alone and has its own cache, but we cannot really expect the users to remember that they should use other suppliers with cache, but not this one.

> In case of both usage it is better to desynchronize their "phases"

Good idea. Will do that. Thanks.

PR #766, opened 2020-05-26
Gh 760 fix after review

OlgaMaciaszek:
CC @robotmrv


spencergibb:
Also @bsideup since he gave advice in this area originally

PR #783, opened 2020-07-07
Lb complete lifecycle

OlgaMaciaszek:


spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerLifecycle.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.springframework.cloud.client.loadbalancer;
+
+/**
+ * @author Olga Maciaszek-Sharma
+ */
+public interface LoadBalancerLifecycle {
+
+	default <RC, RES, T> boolean supports(Class<RC> reqClass, Class<RES> responseClass,
```
spencergibb, line 24: Move generics to interface

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/annotation/LoadBalancerClient.java
```diff
@@ -67,4 +67,6 @@
 	 */
 	Class<?>[] configuration() default {};
 
+	Class<?>[] lifecycleProcessors() default {};
```
spencergibb, line 70: Just get lifecycle processors from NamedContextFactory (ie from the `configuration` attribute) so this new field can be removed.

OlgaMaciaszek:
> Should it be `Class<X> clazz`?

It looks this way so that `NamedContextFactory` matches it. This has also puzzled me, but apparently that's the signature there.

OlgaMaciaszek:
Fixes gh-675.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
```diff
@@ -74,19 +88,35 @@ public ClientHttpResponse intercept(final HttpRequest request, final byte[] body
...
 							this.requestFactory.createRequest(request, body, execution));
 			int statusCode = response.getRawStatusCode();
 			if (retryPolicy != null && retryPolicy.retryableStatusCode(statusCode)) {
 				byte[] bodyCopy = StreamUtils.copyToByteArray(response.getBody());
 				response.close();
-				throw new ClientHttpResponseStatusCodeException(serviceName, response,
-						bodyCopy);
+				ClientHttpResponseStatusCodeException clientHttpResponseStatusCodeException = new ClientHttpResponseStatusCodeException(
+						serviceName, response, bodyCopy);
+				supportedLifecycleProcessors.forEach(lifecycle -> lifecycle.onComplete(
```
OlgaMaciaszek, line 111: I'm not entirely sure whether we should call `onComplete` here as lifecycle processors' `onComplete` from `BlockingLoadBalancerClient` will also be called during the `execute` method call in line 103. However, in `BlockingLoadBalancerClient` we pass `DefaultRequestContext` and `Object` for support check and here `HttpRequestContext` and `ClientHttpResponse`, so we can have more granular lifecycle processors. On the other hand, the ones that match with `DefaultRequestContext` and `Object` will be executed twice, which is not optimal.
@spencergibb wdyt?

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -947,6 +946,13 @@ public class CustomLoadBalancerConfiguration {
 
 NOTE:: `HealthCheckServiceInstanceListSupplier` has its own caching mechanism based on Reactor Flux `replay()`, therefore, if it's being used, you may want to skip wrapping that supplier with `CachingServiceInstanceListSupplier`.
 
+[[spring-cloud-loadbalancer-hints]]
+=== Spring Cloud LoadBalancer Hints
+
+Spring Cloud LoadBalancer allows you to set hints that will be passed to the LoadBalancer within the `Request` object and that can later be used in `ReactiveLoadBalancer` implementations that can handle them.
+
```
cah-andrew-fitzgerald, line 953: A hint is just a String, right?

OlgaMaciaszek: Yes, it's similar to how it's worked in the Ribbon-based implementation, but if you think it should be something else, we can discuss it here.

fitzoh: Was just verifying, wasn't 100% sure from the docs

OlgaMaciaszek: ok. will add info to docs.

docs/src/main/asciidoc/_configprops.adoc
```diff
@@ -34,6 +34,7 @@
 |spring.cloud.loadbalancer.health-check.initial-delay |  | Initial delay value for the HealthCheck scheduler.
 |spring.cloud.loadbalancer.health-check.interval | 25s | Interval for rerunning the HealthCheck scheduler.
 |spring.cloud.loadbalancer.health-check.path |  | 
+|spring.cloud.loadbalancer.hint |  | Allows setting the value of <code>hint</code> that is passed on to the LoadBalancer request and can subsequently be used in {@link ReactiveLoadBalancer} implementations.
```
Buzzardo, line 37: LoadBalancer and ReactiveLoadBalancer should be marked as code. You can use backticks to do it if you want something shorter than the code element. So `LoadBalancer` rather than <code>LoadBalancer</code>.

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -947,6 +946,13 @@ public class CustomLoadBalancerConfiguration {
 
 NOTE:: `HealthCheckServiceInstanceListSupplier` has its own caching mechanism based on Reactor Flux `replay()`, therefore, if it's being used, you may want to skip wrapping that supplier with `CachingServiceInstanceListSupplier`.
 
+[[spring-cloud-loadbalancer-hints]]
+=== Spring Cloud LoadBalancer Hints
+
+Spring Cloud LoadBalancer allows you to set `String` hints that will be passed to the LoadBalancer within the `Request` object and that can later be used in `ReactiveLoadBalancer` implementations that can handle them.
```
Buzzardo, line 952: Change to 

Spring Cloud LoadBalancer lets you set `String` hints that are passed to the LoadBalancer within the `Request` object and that can later be used in `ReactiveLoadBalancer` implementations that can handle them.

(Your sentence is correct, but we have a different editorial standard.)

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -1007,6 +1013,18 @@ public class MyConfiguration {
 ----
 ====
 
+[[loadbalancer-lifecycle]]
+=== Spring Cloud LoadBalancer Lifecycle
+
+One type of beans that it may be useful to register using <<custom-loadbalancer-configuration,Custom LoadBalancer configuration>> is `LoadBalancerLifecycle`.
```
Buzzardo, line 1019: Change "beans" to "bean".

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -1007,6 +1013,18 @@ public class MyConfiguration {
...
+
+The `LoadBalancerLifecycle` beans provide callback `onStart(Request<RC> request)` and `onComplete(CompletionContext<RES, T> completionContext)` methods that you should implement to specify what actions should take place before and after load-balancing.
```
Buzzardo, line 1021: Change to

The `LoadBalancerLifecycle` beans provide callback methods, named `onStart(Request<RC> request)` and `onComplete(CompletionContext<RES, T> completionContext)`, that you should implement to specify what actions should take place before and after load-balancing.

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -1007,6 +1013,18 @@ public class MyConfiguration {
...
+
+`onStart(Request<RC> request)` takes as a `Request` object as a parameter. It contains data that will be used to select an appropriate instance, including the downstream client request and <<spring-cloud-loadbalancer-hints,hint>>. To the `onComplete(CompletionContext<RES, T> completionContext)` method, on the other hand, a `CompletionContext` object is provided that contains the LoadBalancer `Response`, including the selected service instance, the `Status` of the request executed against that service instance and, if available, the response returned to the downstream client and, if an exception has occurred, the corresponding `Throwable`.
```
Buzzardo, line 1023: Change to

`onStart(Request<RC> request)` takes a `Request` object as a parameter. It contains data that is used to select an appropriate instance, including the downstream client request and <<spring-cloud-loadbalancer-hints,hint>>. On the other hand, a `CompletionContext` object is provided to the `onComplete(CompletionContext<RES, T> completionContext)` method. It contains the `LoadBalancer` `Response`, including the selected service instance, the `Status` of the request executed against that service instance and (if available) the response returned to the downstream client, and (if an exception has occurred) the corresponding `Throwable`.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/CompletionContext.java
```diff
@@ -19,22 +19,46 @@
 import org.springframework.core.style.ToStringCreator;
 
 /**
+ * Allows to propagate data related to load-balanced call completion status.
```
Buzzardo, line 22: Change to

Allows propagation of data related to load-balanced call completion status.

docs/src/main/asciidoc/_configprops.adoc
OlgaMaciaszek: This is file is automatically generated.

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -1007,6 +1013,18 @@ public class MyConfiguration {
...
 
+[[loadbalancer-lifecycle]]
+=== Spring Cloud LoadBalancer Lifecycle
+
+One type of bean that it may be useful to register using <<custom-loadbalancer-configuration,Custom LoadBalancer configuration>> is `LoadBalancerLifecycle`.
+
+The LoadBalancerLifecycle beans provide callback methods, named `onStart(Request<RC> request)` and `onComplete(CompletionContext<RES, T> completionContext)`, that you should implement to specify what actions should take place before and after load-balancing.
+
+`onStart(Request<RC> request)` takes a `Request` object as a parameter. It contains data that is used to select an appropriate instance, including the downstream client request and <<spring-cloud-loadbalancer-hints,hint>>. On the other hand, a `CompletionContext` object is provided to the `onComplete(CompletionContext<RES, T> completionContext)` method. It contains the LoadBalancer `Response`, including the selected service instance, the `Status` of the request executed against that service instance and (if available) the response returned to the downstream client, and (if an exception has occurred) the corresponding `Throwable`.
+
```
fitzoh, line 1024: Not sure where the right place to mention it is, but I was struggling to find descrptions of what `RC`, `RES`, and `T` represent/mean

Buzzardo: We could add a note after that paragraph to explain those three. To save Olga some typing, here's a structure for it:

NOTE: In the preceding method calls, `RC` means ..., `RES` means ..., and `T` means....

(Notice the extra period at the end, to close the sentence.)

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
```diff
@@ -113,10 +143,28 @@ private RetryTemplate createRetryTemplate(String serviceName, HttpRequest reques
...
 			template.setListeners(retryListeners);
 		}
-		template.setRetryPolicy(!this.lbProperties.isEnabled() || retryPolicy == null
+		template.setRetryPolicy(!this.retryProperties.isEnabled() || retryPolicy == null
 				? new NeverRetryPolicy() : new InterceptorRetryPolicy(request,
 						retryPolicy, this.loadBalancer, serviceName));
 		return template;
 	}
 
+	private Set<LoadBalancerLifecycle> supportedLifecycleProcessors(String serviceId) {
```
spencergibb, line 152: I suspect this will be a common thing to do. Maybe move to a common class after this PR is merged.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
spencergibb: shouldn't we call `loadBalancer.execute()`? Then that will take care of the lifecycle? We should test that lifecycle methods aren't called twice. Not required to merge, but should happen shortly after.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/ServiceInstanceChooser.java
```diff
@@ -33,4 +34,13 @@
...
 	ServiceInstance choose(String serviceId);
 
+	/**
+	 * Chooses a ServiceInstance from the LoadBalancer for the specified service and
+	 * LoadBalancer request.
+	 * @param serviceId The service ID to look up the LoadBalancer.
+	 * @param request The request to pass on to the LoadBalancer
+	 * @return A ServiceInstance that matches the serviceId.
+	 */
+	<T> ServiceInstance choose(String serviceId, Request<T> request);
```
spencergibb, line 44: I'm not sure this method is required. Should call execute?

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
OlgaMaciaszek: Will get this done now already.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/ServiceInstanceChooser.java
OlgaMaciaszek: From what I see, this entire interface seems to have been added specifically for the RetryInterceptor; I can have a closer look at that interceptor and see if it can be reworked in such a way that this interface can be removed, but I'd rather do it as a separate issue and PR.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
OlgaMaciaszek: Regarding using `execute()`, I have added [this comment](https://github.com/spring-cloud/spring-cloud-commons/pull/783#discussion_r458025632). However, actually, as a result of [this call](https://github.com/spring-cloud/spring-cloud-commons/blob/c8a57be0f1d82f9f381636ac530ad69fc0985a9e/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java#L80-L82), the Lifecycle `onComplete()` calls for `FAILED` and `SUCCESS` are being executed anyway. I thought that adding them here would allow for a better granularity since here we can filter the lifecycle processors by `ClientHttpResponse.class` as opposed to `Object.class` in `BlockingLoadBalancerClient`. However, as they would duplicate, I will remove it from here and add the test you've suggested. I will also add calling `onComplete(DISCARDED[...])` for when the retrieved service instance is `null`. Like this, all calls should be handled. Let me know what you think about this.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
OlgaMaciaszek: I am adding a static method to avoid instantiating new objects at every filter/ load-balancing calls since we only know all the data to filter against at method level and would not be able to instantiate it with appropriate field values at a singleton field initialisation level.

OlgaMaciaszek:
Fixes gh-675.

Issue #785, opened 2020-07-09
DiscoveryClientHealthIndicator should not query for all services


bono007:
**Enhancement**

The DiscoveryClientHealthIndicator invokes it's configured DiscoveryClient `getServices` method to determine its health. 

From [DiscoveryClientHealthIndicator](https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/health/DiscoveryClientHealthIndicator.java#L69)
```java
DiscoveryClient client = this.discoveryClient.getIfAvailable();
List<String> services = client.getServices();
String description = (this.properties.isIncludeDescription())
    ? client.description() : "";
builder.status(new Status("UP", description)).withDetail("services",
    services);
```
In our system, the services response from our ConsulDiscoveryClient is over 500KB long, so this operation is unnecessarily heavy. Consider allowing a lighter check. 

**NOTE**: This improvement is already added in `ConsulHealthIndicator` in [#651](https://github.com/spring-cloud/spring-cloud-consul/pull/651).

#### Suggested approach
A complexity of this is that the DiscoveryClientHealthIndicator codes to the DiscoveryClient interface and there is no other available method that could be "lighter". 

One approach could be to add a default method in DiscoveryClient (maybe `isHealthy` or something along those lines) that just delegates to `getServices()`. This would allow client implementations that want to opt-in to this behavior to
do so on their own timeline. 

Once the above is in place, then the ConsulDiscoveryClient could override that default method and do a lighter weight query.

bono007:
I am willing to submit the MR for this enhancement but I want to get buy-in from the team before I head out on it. Thanks.

PR #787, opened 2020-07-11
Add probe method to DiscoveryClient

bono007:
Adds a probe method to `Reactive/DiscoveryClient` and optionally use in `Reactive/DiscoveryClientHealthIndicator`. 

Fixes gh-785.

@spencergibb this is the follow on to the change in [ConsulHealthIndicator](https://github.com/spring-cloud/spring-cloud-consul/pull/651).


bono007:
Good point @ryanjbaxter . I meant to update this and target to 2.2.x w/ `useServicesQuery=true`. We could decide later what to do in master wrt which default value we want for  `useServicesQuery` going forward. I will do that shortly.

bono007:
Closed in favor of #789

Issue #481, opened 2019-02-13
Support A "Hedge" LoadBalancer


ryanjbaxter:
See this Spring Tip for the basic idea
https://spring.io/blog/2019/01/23/spring-tips-hedging-client-requests-with-the-reactive-webclient-and-a-service-registry

Here is some code 
https://github.com/spring-tips/hedging/blob/52a40255435eccb711808b4779aad8d639cc2264/client/src/main/java/com/example/client/ClientApplication.java#L40-L100

ryanjbaxter:
@spencergibb @OlgaMaciaszek would love your thoughts on an implementation.  Here are some options I have been thinking of

1.  We could add  a separate annotation, but this seems a bit confusing since it is really just a change in the algorithm we are using
2.  Add a property to `@LoadBalancer` something like `@LoadBalancer(algorithm="roundrobin")` this would be the default.  Then we could have `@LoadBalancer(algorithm="hedge")`.  This approach seems more flexible so if we have other algorithms in the future that would be easy.  The confusion with this approach is that we already have a way to configure the algorithm via ribbon that is different than this.

joshlong:
Hi! I would love to see something like this in the SCC. There was a really good comment on the YouTube video for my Spring Tip by a Ashok Koyi, which i reproduce here, that bears consideration.

"I understand its just an example, but this implementation has various problems

1. It does not handle errors properly. If any of the upstream servers always fail fast due to any issue at its end, the overall client call will fail even if another service returns the response with slightly more latency
2. You shared a HashMap across multiple threads (most commons Scheduler implementation), which is completely unsafe
3. This does not work with all load balancer algorithms. Many people prefer latency based next server selection (and not round robin). If this high value algo is selected, this solution becomes completely useless as loadBalancer always returns the best server according to the past statistics (the same server again and again) & only single call to the past best server is always selected
4. If by the time between you retrieve service ids from discover client &  the time you ask load balancer for next server, if the service registry info changes, say a bunch of servers leave due to network partition, your application does not behave the way you intend, i.e you might have less number of upstream servers than you intended (or) you may get no server returned if all he servers are invisible now

Overall, the next best server selection should be done inside the loadbalancer as he might have latency information. The concept is fine, but your implementation needs to be changed significantly to be anywhere near to be used in a production grade system due to the issues I mentioned above﻿"

spencergibb:
[...]

PR #806, opened 2020-08-10
Feature: Enable request hedging for WebClient

Anusien:
Hedging is a mechanism for idempotent requests to automatically start a new request after some time threshold has been reached. For example, automatically firing off a new request when the latency of an existing HTTP call is at the 95th percentile. In some kind of retry mechanism, you'd have to cancel the first request. With hedging, you instead let them all run and simply use the first one that responds. The goal is to trade off some extra network traffic for reduced tail latencies. This is something we built and have been using ourselves. We thought it would be useful to contribute back to the community.

There are two specific implementation issues that I need help from the community here to answer.

First, the implementation here uses `io.micrometer:micrometer.core` to track the histogram of response times (to calculate the percentile latency). spring-cloud-commons doesn't have a required dependency on that library; it's optional through `org.springframework.boot:spring-boot-start-actuator`. I'm at a loss whether the right thing to do is to factor out this class with the dependency, add the dependency as required, or any other compromise (use a different library; note the dependency is needed to use the class).

The second is ordering. This thing adds an `ExchangeFilterFunction`. There's no easy way to control where this gets added in the filter list relative to the filters added by, for example, load balancing. The solutions I came up with to control this require the end user to override both `LoadBalancerWebClientBuilderBeanPostProcessor` and this thing's BeanPostProcessor`.


spencergibb:
https://github.com/spring-cloud/spring-cloud-commons/issues/481

PR #809, opened 2020-08-13
Remove URL check that causes loading 401 responses from Spring Cloud …

rgoers:
Log4j 2 provides enhanced Spring Boot and Spring Cloud Config support with its log4j-spring-boot and log4j-spring-cloud-config-client modules. The Spring Cloud Config Client allows Log4j configuration files to be managed in Spring Cloud Config allowing for dynamic updating of the configuration, configurable SSL support, and support for credentials to access Spring Cloud Config server.

Unfortunately, Spring Cloud Commons' PropertySourceBootstrapConfiguration class causes a 401 error response when trying to access a logging configuration file from Spring Cloud Configuration server when credentials are required. This check really provides no value since the underlying LoggingSystem will throw an exception if the url cannot be accessed, and because it does not include credentials it results in a 401 response that prevents the Spring Boot application from starting if credentials are required to access the Spring Cloud Configuration server.

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/config/PropertySourceBootstrapConfiguration.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/config/PropertySourceBootstrapConfiguration.java
@@ -50,7 +50,6 @@ import org.springframework.core.env.Environment;
 import org.springframework.core.env.MutablePropertySources;
 import org.springframework.core.env.PropertySource;
 import org.springframework.core.env.StandardEnvironment;
-import org.springframework.util.ResourceUtils;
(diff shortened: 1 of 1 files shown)


spencergibb:
We'll follow the conversation in the boot PR

rgoers:
Thanks Spencer. I should note that the check in this piece of code should never fail. By the time this check happens the check in Spring Boot will already have occurred. Obviously if that fails this will never be reached. However, I do understand the desire to keep the logic the same.

rgoers:
The fix for Spring Boot is available in 2.2.10 and 2.3.4.  Please either apply this or modify the affected code to match Spring Boot.

rgoers:
Any updates on this?

PR #817, opened 2020-09-07
Compatibility check if Spring Boot version is unknown

dsyer:
It is better not to fail the compatibility check if the Spring Boot version is unknown.


spencergibb:
Cherry picked into 2.2.x and merged into master via f7360b05adc987ae2aaf9796d2216b147f199277

PR #838, opened 2020-10-12
Move onStart() call to choose method.

OlgaMaciaszek:


spencergibb:
Why the move?

OlgaMaciaszek:
So for instances where the `choose()` method is called alone (for example in Feign load-balanced clients), the suppliers will be called and whoever calls this method does not have to remember to trigger the suppliers, but it's done automatically by the LB.

OlgaMaciaszek:
Also, have renamed some of the request context classes to have `DefaultRequestContext` at the base of the hierarchy so that it's more intuitive for people that will be creating their `LoadBalancerLifecycle` classes.

spencergibb:
I think this complicates the idea of `LoadBalancerLifecycle`. We want folks to **not** use `choose()` directly. It should be documented that if they do, they are responsible for lifecycle method calls. Since we own the feign impl, we can do that. We should add tests there because it seems like this is a response to forgetting to make those call.

TLDR don't merge this and add some docs and javadocs.

OlgaMaciaszek:
Ok, you've convinced me about not moving it. Will close this.

Issue #859, opened 2020-12-03
LoadBalancer: Add Request-Based Sticky LB implementation.


OlgaMaciaszek:
See the discussions in https://github.com/spring-cloud/spring-cloud-commons/issues/689 and https://github.com/spring-cloud/spring-cloud-commons/pull/764.

PR #860, opened 2020-12-03
Request based sticky session

OlgaMaciaszek:
Fixes gh-859.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-commons.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-commons.adoc
@@ -972,9 +972,9 @@ WARNING: `HealthCheckServiceInstanceListSupplier` has its own caching mechanism 
 
 === Request-based Sticky Session for LoadBalancer
 
-It is possible to set up the LoadBalancer in such a way that it will prefer the instance with `instanceId` provided in a request cookie. We currently support this if the request is being passed to the LoadBalancer either via the `ClientRequestContext` or `ServerHttpRequestContext`, which are being used by the SC LoadBalancer exchange filter functions and filters.
+You can set up the LoadBalancer in such a way that it prefers the instance with `instanceId` provided in a request cookie. We currently support this if the request is being passed to the LoadBalancer through either `ClientRequestContext` or `ServerHttpRequestContext`, which are used by the SC LoadBalancer exchange filter functions and filters.
 
-For that, you will need to use the `RequestBasedStickySessionServiceInstanceListSupplier`. It can be configured either by setting the value of `spring.cloud.loadbalancer.configurations` to `request-based-sticky-session` or by providing your own `ServiceInstanceListSupplier` bean, for example:
+For that, you need to use the `RequestBasedStickySessionServiceInstanceListSupplier`. You can configure it either by setting the value of `spring.cloud.loadbalancer.configurations` to `request-based-sticky-session` or by providing your own `ServiceInstanceListSupplier` bean -- for example:
 
 [[health-check-based-custom-loadbalancer-configuration]]
 [source,java,indent=0]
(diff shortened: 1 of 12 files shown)


docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -970,6 +970,32 @@ public class CustomLoadBalancerConfiguration {
 
 WARNING: `HealthCheckServiceInstanceListSupplier` has its own caching mechanism based on Reactor Flux `replay()`. Therefore, if it's being used, you may want to skip wrapping that supplier with `CachingServiceInstanceListSupplier`.
 
+=== Request-based Sticky Session for LoadBalancer
+
+It is possible to set up the LoadBalancer in such a way that it will prefer the instance with `instanceId` provided in a request cookie. We currently support this if the request is being passed to the LoadBalancer either via the `ClientRequestContext` or `ServerHttpRequestContext`, which are being used by the SC LoadBalancer exchange filter functions and filters.
+
+For that, you will need to use the `RequestBasedStickySessionServiceInstanceListSupplier`. It can be configured either by setting the value of `spring.cloud.loadbalancer.configurations` to `request-based-sticky-session` or by providing your own `ServiceInstanceListSupplier` bean, for example:
```
Buzzardo, line 977: Change to:

For that, you need to use the `RequestBasedStickySessionServiceInstanceListSupplier`. You can configure it either by setting the value of `spring.cloud.loadbalancer.configurations` to `request-based-sticky-session` or by providing your own `ServiceInstanceListSupplier` bean -- for example:

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -970,6 +970,32 @@ public class CustomLoadBalancerConfiguration {
...
+			ConfigurableApplicationContext context) {
+		return ServiceInstanceListSupplier.builder()
+					.withDiscoveryClient()
+					.withRequestBasedStickySession()
+					.build(context);
+	    }
+	}
+----
+
+For that functionality, it will be useful to have the selected service instance (that can be different from the one in the original request cookie in case that one was not available) to be updated before sending the request forward. In order to do that set the value of `spring.cloud.loadbalancer.sticky-session.add-service-instance-cookie` to `true`.
```
Buzzardo, line 995: For that functionality, it is useful to have the selected service instance (which can be different from the one in the original request cookie if that one is not available) to be updated before sending the request forward. To do that, set the value of `spring.cloud.loadbalancer.sticky-session.add-service-instance-cookie` to `true`.

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -970,6 +970,32 @@ public class CustomLoadBalancerConfiguration {
...
+
+By default, the name of the cookie is `sc-lb-instance-id`. It can be modified by changing the value of `spring.cloud.loadbalancer.instance-id-cookie-name` property.
```
Buzzardo, line 997: Change to:

By default, the name of the cookie is `sc-lb-instance-id`. You can modify it by changing the value of the `spring.cloud.loadbalancer.instance-id-cookie-name` property.

Buzzardo (changes requested):

spencergibb (approved):

docs/src/main/asciidoc/spring-cloud-commons.adoc
SanthoshEnricher: Hi @Buzzardo 
Do we need to add spring.cloud.loadbalancer.sticky-session.add-service-instance-cookie in cloud gateway or in service which needs sticky session configuration?

OlgaMaciaszek: in GW, so that the selected instance is passed or in the LB client, if the instance is selected there directly (without routing)

PR #868, opened 2020-12-11
Add RandomLoadBalancer, along with tests and docs.

OlgaMaciaszek:


spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/RandomLoadBalancer.java
```diff
@@ -0,0 +1,121 @@
...
+		this.serviceInstanceSupplier = serviceInstanceSupplier;
+	}
+
+	@SuppressWarnings("rawtypes")
+	@Override
+	public Mono<org.springframework.cloud.client.loadbalancer.reactive.Response<ServiceInstance>> choose(
+			Request request) {
+		// TODO: move supplier to Request?
+		// Temporary conditional logic till deprecated members are removed.
+		if (serviceInstanceListSupplierProvider != null) {
```
OlgaMaciaszek, line 81: This is so convoluted in order to maintain deprecated approach. Later, in 3.0.0, it's cleaned up: https://github.com/spring-cloud/spring-cloud-commons/blob/random-load-balancer/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/RandomLoadBalancer.java

spencergibb (approved):

PR #870, opened 2020-12-16
Moves classes from spring-cloud-security to spring-cloud-commons

marcingrzejszczak:
Moves classes from spring-cloud-security to spring-cloud-commons

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/commons/security/ResourceServerTokenRelayAutoConfiguration.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/commons/security/ResourceServerTokenRelayAutoConfiguration.java
@@ -65,7 +65,7 @@ import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
 @ResourceServerTokenRelayAutoConfiguration.ConditionalOnOAuth2ClientInResourceServer
 @ConditionalOnClass(ResourceServerConfiguration.class)
 @ConditionalOnWebApplication
-@ConditionalOnProperty(value = "security.oauth2.resource.tokenRelay", matchIfMissing = true)
+@ConditionalOnProperty(value = "spring.cloud.mvc.token-relay.enabled=true", matchIfMissing = true)
 public class ResourceServerTokenRelayAutoConfiguration {
 
 	@Bean
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/commons/security/ResourceServerTokenRelayAutoConfiguration.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/commons/security/ResourceServerTokenRelayAutoConfiguration.java
@@ -1,0 +1,132 @@
+/*
+ * Copyright 2012-2015 the original author or authors.
+ *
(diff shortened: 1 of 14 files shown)


marcingrzejszczak:
Related https://github.com/spring-cloud/spring-cloud-security/issues/231

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/security/ResourceServerTokenRelayAutoConfiguration.java
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * @author Dave Syer
+ *
+ */
+@Configuration(proxyBeanMethods = false)
+@AutoConfigureAfter(OAuth2AutoConfiguration.class)
+@ResourceServerTokenRelayAutoConfiguration.ConditionalOnOAuth2ClientInResourceServer
+@ConditionalOnClass(ResourceServerConfiguration.class)
+@ConditionalOnWebApplication
+@ConditionalOnProperty(value = "security.oauth2.resource.tokenRelay", matchIfMissing = true)
```
spencergibb, line 68: we should reference properties via canonical form, so here `tokey-relay` (I know this was probably how it was).

I wonder if this property should be mvc specific. `spring.cloud.mvc.token-relay.enabled=true`? or something

marcingrzejszczak: `spring.cloud.mvc.token-relay.enabled=true` sounds like a good idea

marcingrzejszczak: that's even better cause those props will end up in the configprops.adoc since they start with spring.cloud

marcingrzejszczak:
Also instead of `security.oauth2.resource.loadBalanced` we could do sth like `spring.cloud.oauth2.load-balanced.enabled=true` ?

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-commons/commit/1f0f750707140bca77995f7aea0cd8b0d471e31a

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/security/ResourceServerTokenRelayAutoConfiguration.java
HaojunRen: In https://github.com/spring-cloud/spring-cloud-commons/blob/main/spring-cloud-commons/src/main/java/org/springframework/cloud/commons/security/ResourceServerTokenRelayAutoConfiguration.java

```
@ConditionalOnProperty(value = "spring.cloud.mvc.token-relay.enabled=true", matchIfMissing = true)
```

Is `=true` a mistake? the key includes `=true`?

spencergibb: yes, fixed in f1b3956

Issue #815, opened 2020-09-05
Property source decryption not working in Ilford


dsyer:
The existing tests all use legacy property source processing (bootstrap context), so there is nothing to indicate that it is meant to be working yet out of the box, but I think maybe we should make sure it does work for backwards compatibility.

dsyer:
I noticed a new feature in Spring Boot 2.4 - there's a `BootstrapRegistry` (kind of like a little DI framework) in the `EnvironmentPostProcessors`. Maybe that can be used to set up the decrypted properties? Could maybe even use a "legacy" Bootstrap application context to back the suppliers registered with the `BootstrapRegistry`.

spencergibb:
I've already used it here https://github.com/spring-cloud/spring-cloud-zookeeper/blob/master/spring-cloud-zookeeper-config/src/main/java/org/springframework/cloud/zookeeper/config/ZookeeperConfigDataLocationResolver.java#L84-L91

spencergibb:
This was definitely on my radar already in some TODO list somewhere, so thanks for creating the issue.

marcingrzejszczak:
Is this connected to https://github.com/spring-cloud/spring-cloud-commons/pull/703 ?

marcingrzejszczak:
Ah ok, I see that `EncryptionIntegrationTests` when disabling the legacy processing doesn't work

PR #872, opened 2020-12-18
Support TextEncryptor in new ConfigData framework

spencergibb:
This updates to allow ConfigData impls such as config client, consul and zk to decrypt credentials before making remote calls.

Still missing support for decryption during normal calls into `Environment`.

Fixes gh-815


spencergibb:
See https://github.com/spring-cloud/spring-cloud-config/pull/1773 for usage

Issue #887, opened 2021-01-20
split packages in spring-cloud-commons prevents usage of java modules when reactor core is used


amtineo:
**Describe the bug**
When using the java module system, there are split packages between the **reactor core** library and **spring-cloud-commons**. This prevents the system from loading the modules since it violates the package rule. 

The following error is thrown when attempting to start a spring boot application:

 ```
Error occurred during initialization of boot layer
java.lang.module.ResolutionException: Modules reactor.core and spring.cloud.commons export package reactor.core.publisher to module spring.boot.starter.actuator
```
**Sample**
Create a project with the following dependencies:

```
<dependency>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-starter-actuator</artifactId>
</dependency>
<dependency>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-starer-web</artifactId>
</dependency>
<dependency>
    <groupId>org.springframework.cloud</groupId>
    <artifactId>spring-cloud-starter-cloud-discovery</artifactId>
</dependency>
<dependency>
    <groupId>io.projectreactor</groupId>
    <artifactId>reactor-core</artifactId>
    <version>3.4.2</version>
</dependency>
```
Define the required modules in `module-info.java`:
```
requires spring.web;
requires spring.boot;
requires spring.boot.autoconfigure;
requires spring.boot.starter.web;
requires spring.boot.starter.tomcat;
requires spring.boot.starter.actuator;
requires org.apache.tomcat.embed.core;
requires org.apache.tomcat.embed.websocket;
requires com.fasterxml.jackson.databind;

requires reactor.core;

exports my.personal.beans;

opens my.personal.beans to spring.core;
```

Possible solution:

Rename the package `reactor.core.publisher` from `spring-cloud-commons` so it doesn't conflict with the same package name found in the `reactor core` library

spencergibb:
Unfortunately, it uses package protected class is `reactor.core.publisher`.

There's a [note](https://github.com/spring-cloud/spring-cloud-commons/commit/7033a9877f4f7ed879c6cd5dbb5dbfbe81118368#diff-cae98f54326b209e094628d030e65c7f1c6f5b0f2d9bebdf25e09daa4407a1d8R22) that this functionality will be moved to reactor and can then be removed here. @TYsewyn and @simonbasle would need to give us an update.

simonbasle:
[...]

PR #888, opened 2021-01-22
Moves CloudFlux to spring cloud package.

spencergibb:
FluxFirstNonEmptyEmitting and Tests used package private classes/interfaces from reactor. These were replaced with the public interfaces they were extending and any default methods copied to the implementation.

LambdaSubscriber was copied to the test where it was used and had similar treatment for its package protected interfaces.

Fixes gh-887


TYsewyn (approved):
LGTM!

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/publisher/FluxFirstNonEmptyEmitting.java
```diff
@@ -325,6 +344,31 @@ public void onComplete() {
 			}
 		}
 
+		@Override
+		public String stepName() {
```
simonbasle, line 348: this generic code has proven to be buggy. best approach is to override the method to return a constant. the goal of `stepName` is to give a human-friendly representation of the operator, which can be used eg. when printing additional backtraces in debug mode.

simonbasle: so here we could have simply `return "firstNonEmpty";`, or perhaps `CloudFlux.firstNonEmpty` to make it obvious this is a cloud-custom operator?

spring-cloud-commons/src/test/java/org/springframework/cloud/commons/publisher/FluxFirstNonEmptyEmittingTests.java
```diff
@@ -172,4 +181,191 @@ public void scanRaceCoordinator() {
 		assertThat(parent.scan(Scannable.Attr.CANCELLED)).isTrue();
 	}
 
+	static class LambdaSubscriber<T> implements CoreSubscriber<T>, Scannable, Disposable {
```
simonbasle, line 184: `LambdaSubscriber` is used in core tests because A) we need to pass a non-null `Subscriber` to the relevant constructors and B) we need something that is implementing `Scannable`.

Rather than copy the entire class, one solution would be to have a child class that extends `BaseSubscriber<T>` and make it implement `Scannable`. The `scanUnsafe` method needs to be overridden, but can simply `return null;`. This would greatly simplify this part, from 200 LoC to about 10.

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/publisher/FluxFirstNonEmptyEmitting.java
```diff
@@ -277,23 +293,26 @@ int resignFromRace() {
 		@Nullable
 		public Object scanUnsafe(Attr key) {
 			if (key == Attr.PARENT) {
-				return s;
+				return ReflectionUtils.getField(s, this);
```
simonbasle, line 296: this should actually work without reflection: resolve class-specific state first, then delegate to `super` for the rest. `DeferredSubscription` already answers `s` for the `Attr.PARENT` key.

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/publisher/FluxFirstNonEmptyEmitting.java
```diff
@@ -277,23 +293,26 @@ int resignFromRace() {
...
 
-			return InnerOperator.super.scanUnsafe(key);
+			return doScanUnsafe(key);
 		}
 
-		@Override
-		public void onSubscribe(Subscription s) {
-			set(s);
+		@Nullable
+		Object doScanUnsafe(Attr key) {
```
simonbasle, line 306: as commented above, this can be melded back into `scanUnsafe` to get:

```
@Nullable
public Object scanUnsafe(Attr key) {
    if (key == Attr.ACTUAL) return actual;
    if (key == Attr.CANCELLED) return parent.cancelled;
    return super.scanUnsafe(key);
}
```

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/publisher/FluxFirstNonEmptyEmitting.java
```diff
@@ -257,7 +267,13 @@ int resignFromRace() {
 	}
 
 	static final class FirstNonEmptyEmittingSubscriber<T> extends Operators.DeferredSubscription
-			implements InnerOperator<T, T> {
+			implements CoreSubscriber<T>, Scannable, Subscription {
+
+		private static final Field s = ReflectionUtils.findField(Operators.DeferredSubscription.class, "s",
```
simonbasle, line 272: the reflection can be eliminated, see below

simonbasle (changes requested):

simonbasle (approved):

PR #920, opened 2021-03-09
Move LoadBalancerServiceInstanceCookieTransformer to loadbalancer module

OlgaMaciaszek:
Keep the class with the autoconfiguration that uses it.

### DIFF
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/LoadBalancerServiceInstanceCookieTransformer.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/LoadBalancerServiceInstanceCookieTransformer.java
@@ -14,12 +14,15 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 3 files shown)


spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/LoadBalancerServiceInstanceCookieTransformer.java
```diff
@@ -14,12 +14,15 @@
  * limitations under the License.
  */
 
-package org.springframework.cloud.client.loadbalancer;
+package org.springframework.cloud.loadbalancer.core;
```
spencergibb, line 17: This is a breaking change, but was this just introduced for this release?

OlgaMaciaszek: This class has not been released yet :).

spencergibb (approved):

Issue #672, opened 2020-01-16
LoadBalancer: support hint-based instance selection


OlgaMaciaszek:
Related to: https://github.com/spring-cloud/spring-cloud-commons/issues/647

Possible implementation:
- a request comes with  `loadBalancerHint` header = `someVal`
- based on that header, we would pass the request with hint `loadBalancer.choose(serviceId, request)`, where `someVal` can be retrieved from `request.getContext()`
- then filter instances by `metadataMap.get("hint").equals("someVal")`

andersenleo:
Hi - 
we're currently migrating our code from the Ribbon load-balancer to Spring-Cloud. We've started hacking up a custom load-balancer (based on `ReactorServiceInstanceLoadBalancer` and fallback to the `RoundRobinLoadBalancer`). 

Our requirements is that we need to query service-instances to check if they own a specific "key" and select the one that holds the key. We're planning to create our own interceptor/filter for `RestTemplate` and `WebClient` to propagate the "key" from a http-header or request-attribute (and create a `Request` with a context that is the "key"). This is more or less a carbon-copy of the existing `ReactorLoadBalancerExchangeFilterFunction` (also for RestTemplate - the `LoadBalancerClient` abstraction isn't really suitable for us).

Now - we can probably work our way around the current Spring-Cloud LoadBalancer code, but it would be nice if we could either align with upcoming changes to the your code -- or even better, if we could contribute our code to Spring-Cloud :). 

So - to summarise: What's the current status on this issue? Are there any design docs or discussions that we could read up on to get an idea about what direction your heading?

OlgaMaciaszek:
@andersenleo would definitely be great if you could provide a PR -  we have this in the backlog for 2020-1 release train (Ilford), but it might be released faster if you contribute. Let me get back to you tomorrow here with some ideas on how we planned to implement it and let's discuss it then.

OlgaMaciaszek:
We've been thinking about the following features:
- A request would come with a specific header, for example `loadBalancerHint`
[...]

PR #923, opened 2021-03-10
Add hint-based instance filtering

OlgaMaciaszek:
Fixes gh-672.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-commons.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-commons.adoc
@@ -1056,17 +1056,17 @@ for any given service by setting the value of the `spring.cloud.loadbalancer.hin
 [[hints-based-loadbalancing]]
 === Hint-Based Load-Balancing
 
-We also provide a `HintBasedServiceInstanceListSupplier`, which is an out-of-the-box implementation for hint-based instance selection.
+We also provide a `HintBasedServiceInstanceListSupplier`, which is a `ServiceInstanceListSupplier` implementation for hint-based instance selection.
 
-`HintBasedServiceInstanceListSupplier` will check for a hint request header (the default header-name is `X-SC-LB-Hint`, but you can modify it by changing the value of the `spring.cloud.loadbalancer.hint-header-name` property) and, if it finds it, will use the hint value passed in the header  to filter service instances.
+`HintBasedServiceInstanceListSupplier` checks for a hint request header (the default header-name is `X-SC-LB-Hint`, but you can modify it by changing the value of the `spring.cloud.loadbalancer.hint-header-name` property) and, if it finds a hint request header, uses the hint value passed in the header  to filter service instances.
 
-If no hint header has been added, `HintBasedServiceInstanceListSupplier` will use <<spring-cloud-loadbalancer-hints,hint values from properties>> to filter service instances.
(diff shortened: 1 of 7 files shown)


spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HintBasedServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,97 @@
...
+				HttpHeaders headers = context.getClientRequest().getHeaders();
+				if (headers != null) {
+					String hintFromHeader = headers.getFirst(properties.getHintHeaderName());
+					if (StringUtils.hasText(hintFromHeader)) {
+						return get().map(instances -> filteredByHint(instances, hintFromHeader));
+					}
+				}
+			}
+		}
+		if (request.getContext() instanceof HintRequestContext) {
```
spencergibb, line 71: I wonder if this is two classes one extends the other with a `protected String getHint()` method, that in the base, it's (HintRequestContext) and the request header specific one you find the header? You'd have to explicitly say you wanted one or the other. Thoughts?

ryanjbaxter: IDK, I could maybe see a scenario where you set a hint via properties as the default, and then in some cases use the header to override that default.  Having two separate classes and having to chose one method or the other would make that use case impossible.  

At the very least I think the logic to find the hint to use should be moved to its own methods in this class to clean things up a bit and make the code easier to read.

OlgaMaciaszek: Yes, I agree with @ryanjbaxter  - the "default override" scenario was why I put this all together. Will extract to separate methods to make the code cleaner, though.

spencergibb (approved):

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -1053,6 +1053,38 @@ Spring Cloud LoadBalancer lets you set `String` hints that are passed to the Loa
...
+
+`HintBasedServiceInstanceListSupplier` will check for a hint request header (the default header-name is `X-SC-LB-Hint`, but you can modify it by changing the value of the `spring.cloud.loadbalancer.hint-header-name` property) and, if it finds it, will use the hint value passed in the header  to filter service instances.
+
+If no hint header has been added, `HintBasedServiceInstanceListSupplier` will use <<spring-cloud-loadbalancer-hints,hint values from properties>> to filter service instances.
+
+If there's no hint set either via the header or via properties, all service instances provided by the delegate will be returned.
+
+While filtering, `HintBasedServiceInstanceListSupplier` will look for service instances that have a matching value set under the `hint` key in their `metadataMap`. If no matching instances are found, all instances provided by the delegate will be returned.
+
+You could use this sample configuration to set it up:
```
Buzzardo, line 1069: Change "this" to "the following". (Otherwise, I think I should look up to what you just described. Then I see the colon and realize what you mean. By then, though, I've had to read the sentence twice.)

Buzzardo (changes requested):

Issue #880, opened 2021-01-07
Allow using secondary ServiceInstanceListSupplier by RetryAwareServiceInstanceListSupplier


OlgaMaciaszek:

agisbert:
@OlgaMaciaszek, I've seen this is what's left for the 3.0.1 milestone. We'd like the version released asap, are you already with this topic? Maybe I could submit a PR if you give me some insights on the issue. 

Regards.

OlgaMaciaszek:
Hello, @agisbert I'm really sorry - I have missed your comment. Working on it already.

OlgaMaciaszek:
@agisbert We've discussed the issue in the team and have decided not to implement it at this point and move it to  "waiting for votes" status instead. However, you can easily create your own `ServiceInstanceListSupplier` similar to the one in [this commit](https://github.com/spring-cloud/spring-cloud-commons/commit/897470abce70ed65f421aa3162e811e36cc8a210) and use it instead.

PR #928, opened 2021-03-16
Provide secondary delegate for RetryAwareServiceInstanceListSupplier.

OlgaMaciaszek:
Fixes gh-880.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-commons.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-commons.adoc
@@ -468,8 +468,6 @@ For the reactive implementation, you can additionally set:
 
 For the reactive implementation, you can also implement your own `LoadBalancerRetryPolicy` to have more detailed control over the load-balanced call retries.
 
-NOTE: For load-balanced retries, by default, we wrap the `ServiceInstanceListSupplier` bean with `RetryAwareServiceInstanceListSupplier` to select a different instance from the one previously chosen, if available. You can disable this behavior by setting the value of `spring.cloud.loadbalancer.retry.avoidPreviousInstance` to `false`.
-
 ====
 [source,java,indent=0]
 ----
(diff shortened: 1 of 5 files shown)


docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -526,6 +524,44 @@ public class MyConfiguration {
...
+	ServiceInstanceListSupplier serviceInstanceListSupplier(ConfigurableApplicationContext context) {
+		return ServiceInstanceListSupplier
+				.builder()
+				.withDiscoveryClient()
+				.withRetryAwareness()
+				.build(context);
+	}
+----
+
+WARNING: If you opt for instantiating `RetryAwareServiceInstanceListSupplier` on your own, make sure you set the value of `spring.cloud.loadbalancer.retry.avoid-previous-instance` to `false` so that we do not instantiate one for you automatically.
```
ryanjbaxter, line 563: Why do we do anything if they supply their own bean?

ryanjbaxter (commented):
I am kind of lost as to why we need a secondary instance supplier in order to ensure we don't return the same instance, can you elaborate?

ryanjbaxter:
I would think that just selecting another instance from the `ZonePreferenceServiceInstanceListSupplier` would be sufficient.  Maybe I am missing the use case.

spencergibb:
I've not seen someone ask for this before except the one person who commented in the issue you filed.

OlgaMaciaszek:
I don't remember whether I've seen anyone else, but it does make sense to me. 2 different suppliers can return completely different instance sets and there may be use-cases where it makes sense to fall back to a different set while retrying.

docs/src/main/asciidoc/spring-cloud-commons.adoc
OlgaMaciaszek: It would be better if they did not have to, however, the bean that we create automatically is created during the same configuration stage and `@ConditionalOnMissingBean` so, in order to avoid the programmatically created supplier getting wrapped by yet another retry supplier and at the same time, not change production behaviour, it could be set in this way. However, I don't like it a lot either, so if we can think of a better way, we could switch it up.

spencergibb:
I'm going to suggest we wait on this one and not try and get this in this service release

ryanjbaxter:
I agree

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -526,6 +524,44 @@ public class MyConfiguration {
 ----
 ====
 
+===== Retry-aware instance selection
+
+For load-balanced retries, by default, we wrap the `ServiceInstanceListSupplier` bean with `RetryAwareServiceInstanceListSupplier` to select a different instance from the one previously chosen, if available. You can disable this behavior by setting the value of `spring.cloud.loadbalancer.retry.avoidPreviousInstance` to `false`.
+You can also instantiate `RetryAwareServiceInstanceListSupplier` on your own. Then, apart from the constructor that allows for passing only one delegate, you can also use `RetryAwareServiceInstanceListSupplier(ServiceInstanceListSupplier delegate, ServiceInstanceListSupplier secondaryDelegate)`. If you do, the secondary delegate will be used for providing instances during retries.
```
Buzzardo, line 530: Change "will be" to "is".

Buzzardo (changes requested):

OlgaMaciaszek:
Closing as [we've decided to wait for votes](https://github.com/spring-cloud/spring-cloud-commons/issues/880#issuecomment-805102484) on this issue.

PR #936, opened 2021-03-30
Add abstract environment post processor for config data imports

ryanjbaxter:


spring-cloud-commons/src/main/java/org/springframework/cloud/commons/ConfigDataMissingEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+		if (!property.stream().anyMatch(impt -> impt.contains(getPrefix()))) {
+			throw new ImportException("spring.config.import missing " + getPrefix(), true);
+		}
+	}
+
+	private List<String> getConfigImports(ConfigurableEnvironment environment) {
+		List<String> property = environment.getProperty(CONFIG_IMPORT_PROPERTY, List.class);
+		if (property == null || property.isEmpty()) {
+			Binder binder = Binder.get(environment);
+			Bindable<String[]> CONFIG_DATA_LOCATION_ARRAY = Bindable.of(String[].class);
```
spencergibb, line 74: This can be a constant (it is in boot).

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/ConfigDataMissingEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+		if (property == null || property.isEmpty()) {
+			throw new ImportException("No spring.config.import set", false);
+		}
+		if (!property.stream().anyMatch(impt -> impt.contains(getPrefix()))) {
+			throw new ImportException("spring.config.import missing " + getPrefix(), true);
+		}
+	}
+
+	private List<String> getConfigImports(ConfigurableEnvironment environment) {
+		List<String> property = environment.getProperty(CONFIG_IMPORT_PROPERTY, List.class);
```
spencergibb, line 71: I assume this works even when there is just one property?

spencergibb (approved):

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/ConfigDataMissingEnvironmentPostProcessor.java
ryanjbaxter: Yes it does (tests are in the downstream projects)

spencergibb (approved):

PR #942, opened 2021-04-08
(PoC) Support configuration per client for BlockingLoadBalancedRetryFactory

Ferioney:
The main idea is to replace using of LoadBalancerProperties with the new LoadBalancerPropertiesFactory. If needed LoadBalancerProperties we should use LoadBalancerPropertiesFactory#getServiceProperties(serviceName).


spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerProperties.java
```diff
@@ -37,6 +38,16 @@
 @ConfigurationProperties("spring.cloud.loadbalancer")
 public class LoadBalancerProperties {
 
+	private Map<String, LoadBalancerProperties> instances = new HashMap<>();
```
OlgaMaciaszek, line 41: Let's rename this field to sth more meaningful (`instance` is a term we use in a very specific way), maybe `clientProperties`, `serviceProperties`, etc.?

OlgaMaciaszek (commented):
@Ferioney, thanks for submitting this. The PoC looks good to me. In fact, we were thinking of implementing a similar solution. @spencergibb @ryanjbaxter @marcingrzejszczak @Haybu - please let me know what you think about the PoC.

If we move forward with this, 2 things are important to keep in mind:
- we would need to implement the changes everywhere where we now use the LB properties, so also in the reactive implementations, such as `ReactorLoadBalancerExchangeFilterFunction`
- the most tricky thing: we would either need to move this to a major release train (which would be Kilburn, which, unfortunately, is not coming that soon) or make sure the changes are not backwards incompatible. As this is, indeed, an important feature, that probably should have been there from the beginning, I don't think we can wait till Kilburn and should implement it already in a non-breaking way. Some possible approaches would be just adding new constructors (and deprecating the old ones) and then handling the fields differently depending on what has been passed or using some kind of property flag that would allow choosing between the old and new approach (any other ideas also welcome)

@Ferioney please let us know what you think about the comments above and whether you would be interested in working on the full PR, keeping those things in mind.

ryanjbaxter:
Yeah I think that this should be doable in a backwards compatible fashion.

Ferioney:
@OlgaMaciaszek thanks for your feedback

- yes, it should be done in all places where properties are used.
- I agree that it should be done with backward compatibility. I changed `LoadBalancerPropertiesFactory` in PoC to use general and per-service properties configuration.

PR #947, opened 2021-04-22
Issue 914 properties solution

Ferioney:
Support properties configuration per client


spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerClientAutoConfiguration.java
```diff
@@ -61,16 +63,18 @@ public ReactorLoadBalancerExchangeFilterFunction loadBalancerExchangeFilterFunct
...
-			ObjectProvider<List<LoadBalancerClientRequestTransformer>> transformers) {
+			ObjectProvider<List<LoadBalancerClientRequestTransformer>> transformers,
+			LoadBalancerPropertiesFactory propertiesFactory) {
 		return new RetryableLoadBalancerExchangeFilterFunction(retryPolicy, loadBalancerFactory, properties,
-				transformers.getIfAvailable(Collections::emptyList));
+				transformers.getIfAvailable(Collections::emptyList), propertiesFactory);
 	}
 
 	@ConditionalOnMissingBean
 	@ConditionalOnProperty(value = "spring.cloud.loadbalancer.retry.enabled", havingValue = "true")
```
Ferioney, line 73: @OlgaMaciaszek @spencergibb 

I see an issue with enabling`LoadBalancerRetryPolicy` by `spring.cloud.loadbalancer.retry.enabled `.

I think we can use `spring.cloud.loadbalancer.retry.enabled` property for enabling retry in general. And then we should check in `policy` (and in other beans) that retry is enabled for a specific service (it has not implemented yet).

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerPropertiesFactory.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.client.loadbalancer;
+
```
OlgaMaciaszek, line 18: Please add javadocs for new classes. Please add your full name and surname with the `@author` tag in the javadocs of all changed/ added classes.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerServiceProperties.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package org.springframework.cloud.client.loadbalancer;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties("spring.cloud.loadbalancer")
```
OlgaMaciaszek, line 24: Let's change the prefix to `spring.cloud.loadbalancer.client` to indicate these are per-client properties and differentiate them from the other properties.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
```diff
@@ -58,6 +58,9 @@
 
 	private final ReactiveLoadBalancer.Factory<ServiceInstance> loadBalancerFactory;
 
+	private LoadBalancerPropertiesFactory factoryProperties;
+
+	@Deprecated
```
OlgaMaciaszek, line 63: Please add javadoc with `@deprecated` tag indicating the suggested replacement.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
```diff
@@ -58,6 +58,9 @@
 
 	private final ReactiveLoadBalancer.Factory<ServiceInstance> loadBalancerFactory;
 
+	private LoadBalancerPropertiesFactory factoryProperties;
```
OlgaMaciaszek, line 61: Please rename the field to `propertiesFactory`.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/LoadBalancerRetryPolicy.java
```diff
@@ -32,27 +32,71 @@
 	 * @param context the context for the retry operation
 	 * @return true to retry on the same service instance
 	 */
+	@Deprecated
```
OlgaMaciaszek, line 35: Please add JavaDoc with `@deprecated` tag indicating the suggested replacement.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/LoadBalancerRetryPolicy.java
```diff
@@ -32,27 +32,71 @@
...
+	default boolean retryableStatusCode(String serviceId, int statusCode) {
+		return retryableStatusCode(statusCode);
+	}
+
 	/**
 	 * Return <code>true</code> to retry on the provided HTTP method.
 	 * @param method the HTTP request method
 	 * @return true to retry on the provided HTTP method
 	 */
+	@Deprecated
```
OlgaMaciaszek, line 89: Please add javadoc with `@deprecated` tag indicating the suggested replacement.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerClientAutoConfiguration.java
OlgaMaciaszek: Yes, we definitely should keep `spring.cloud.loadbalancer.retry.enabled` to be used as a global switch, but we could probably also be able to set `spring.cloud.loadbalancer.client.clientName.retry.enabled` per client - wdyt?

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerClientAutoConfiguration.java
```diff
@@ -50,9 +51,10 @@
 	@Bean
 	public ReactorLoadBalancerExchangeFilterFunction loadBalancerExchangeFilterFunction(
 			ReactiveLoadBalancer.Factory<ServiceInstance> loadBalancerFactory, LoadBalancerProperties properties,
-			ObjectProvider<List<LoadBalancerClientRequestTransformer>> transformers) {
+			ObjectProvider<List<LoadBalancerClientRequestTransformer>> transformers,
+			LoadBalancerPropertiesFactory factoryProperties) {
```
OlgaMaciaszek, line 55: Please rename the argument to `propertiesFactory`.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerClientAutoConfiguration.java
```diff
@@ -50,9 +51,10 @@
...
 		return new ReactorLoadBalancerExchangeFilterFunction(loadBalancerFactory, properties,
-				transformers.getIfAvailable(Collections::emptyList));
+				transformers.getIfAvailable(Collections::emptyList), factoryProperties);
```
OlgaMaciaszek, line 57: Please rename the argument to `propertiesFactory`.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerExchangeFilterFunction.java
```diff
@@ -77,13 +80,21 @@ public ReactorLoadBalancerExchangeFilterFunction(ReactiveLoadBalancer.Factory<Se
 		this(loadBalancerFactory, properties, Collections.emptyList());
 	}
 
+	@Deprecated
```
OlgaMaciaszek, line 83: Please add javadoc with `@deprecated` tag indicating the suggested replacement.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/RetryableLoadBalancerExchangeFilterFunction.java
```diff
@@ -87,6 +90,7 @@ public RetryableLoadBalancerExchangeFilterFunction(LoadBalancerRetryPolicy retry
 		this(retryPolicy, loadBalancerFactory, properties, Collections.emptyList());
 	}
 
+	@Deprecated
```
OlgaMaciaszek, line 93: Please add javadoc with `@deprecated` tag indicating the suggested replacement.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/blocking/client/BlockingLoadBalancerClient.java
```diff
@@ -58,11 +59,19 @@
 
 	private final LoadBalancerProperties properties;
 
+	private LoadBalancerPropertiesFactory propertiesFactory;
+
+	@Deprecated
```
OlgaMaciaszek, line 64: Please add javadoc with `@deprecated` tag indicating the suggested replacement.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/blocking/retry/BlockingLoadBalancedRetryFactory.java
```diff
@@ -33,13 +34,31 @@
 
 	private final LoadBalancerProperties loadBalancerProperties;
 
+	private LoadBalancerPropertiesFactory factoryProperties;
```
OlgaMaciaszek, line 37: Please rename the field to `propertiesFactory`.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/blocking/retry/BlockingLoadBalancedRetryFactory.java
```diff
@@ -33,13 +34,31 @@
...
+
+	@Deprecated
```
OlgaMaciaszek, line 39: Please add javadoc with `@deprecated` tag indicating the suggested replacement.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/LoadBalancerServiceInstanceCookieTransformer.java
```diff
@@ -39,16 +40,25 @@
 
 	private final LoadBalancerProperties.StickySession stickySessionProperties;
 
+	private LoadBalancerPropertiesFactory propertiesFactory;
+
+	@Deprecated
```
OlgaMaciaszek, line 45: Please add javadoc with `@deprecated` tag indicating the suggested replacement.

OlgaMaciaszek (changes requested):
@Ferioney thanks for submitting this. In general, this looks good. Have added some comments, though - please address. Also, please add some tests verifying specifically the new property logic. Please add information in the docs regarding this feature (in `spring-cloud-commons.adoc`, within the "Spring Cloud LoadBalancer" section.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerPropertiesFactory.java
Ferioney: Added

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerServiceProperties.java
Ferioney: Changed

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
Ferioney: Added

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
Ferioney: Changed

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/LoadBalancerRetryPolicy.java
Ferioney: Added

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/LoadBalancerRetryPolicy.java
Ferioney: Added

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerClientAutoConfiguration.java
Ferioney: Changed

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerClientAutoConfiguration.java
Ferioney: Changed

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerExchangeFilterFunction.java
Ferioney: Added

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/RetryableLoadBalancerExchangeFilterFunction.java
Ferioney: Added

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/blocking/client/BlockingLoadBalancerClient.java
Ferioney: Added

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/blocking/retry/BlockingLoadBalancedRetryFactory.java
Ferioney: Changed

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/blocking/retry/BlockingLoadBalancedRetryFactory.java
Ferioney: Added

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/LoadBalancerServiceInstanceCookieTransformer.java
Ferioney: Added

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerClientAutoConfiguration.java
Ferioney: But I'm not sure how to implement switching by `spring.cloud.loadbalancer.client.clientName.retry.enabled` for reactive loadBalancer. Reactive loadBalance has two functions - with (`RetryableLoadBalancerExchangeFilterFunction`) and without (`ReactorLoadBalancerExchangeFilterFunction`) retry. and as I see there no possibility to create pipeline for a few filters (we can but we don't have an order guaranty). so I tried to make it in retry filter - [RetryableLoadBalancerExchangeFilterFunction.java](https://github.com/spring-cloud/spring-cloud-commons/pull/947/files#diff-a8e7d22f94822da32e047a7bac0fd8c6cc36fdffb667bd66682121a06b2e636fR128)

OlgaMaciaszek: That looks fine, but I'm thinking we also need a way to handle global properties; I imagine we would do so with retries, but also possibly with caching (we should allow disabling it globally - could you add this?) and possibly other functionalities in the future, so apart from the method `getLoadBalancerProperties(String serviceName)` in the factory, we could also have sth like `getGlobalLoadBalancerProperties()` (that would just return the non-`client`-prefixed properties - then for the major we would have to move most of the properties to the `spring.cloud.loadbalancer.client.clientName` - prefixed, and only leave the global ones prefixed with just `spring.cloud.loadbalancer`)  and in cases where we want to offer a global off switch (or some other global property) we could verify first the global properties. wdyt?

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -1228,6 +1228,25 @@ NOTE: The meters are registered in the registry when at least one record is adde
 
 TIP: You can further configure the behavior of those metrics (for example, add https://micrometer.io/docs/concepts#_histograms_and_percentiles[publishing percentiles and histograms]) by https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#production-ready-metrics-per-meter-properties[adding `MeterFilters`].
 
+=== Spring Cloud LoadBalancer per client configuration
+You can configure multiple loadBalancer clients by using `spring.cloud.loadbalancer.client.services[]` property:
```
OlgaMaciaszek, line 1232: Please add a few more words of explanation indicating that the properties can have different values per client, that there's a switch flag to enable this function and that, by default, the global properties only are being used to avoid backwards incompatibility.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerServiceProperties.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * A {@link ConfigurationProperties} bean for client LoadBalancer.
+ *
+ * @author Andrii Bohutskyi
+ */
+@ConfigurationProperties("spring.cloud.loadbalancer.client")
+public class LoadBalancerServiceProperties {
+
+	private Map<String, LoadBalancerProperties> services = new HashMap<>();
```
OlgaMaciaszek, line 32: Could you modify the implementation, so that the property is  `spring.cloud.loadbalancer.client.foo1.healthcheck...` as opposed to `spring.cloud.loadbalancer.client.services.foo1.healthcheck...`?

OlgaMaciaszek (changes requested):
@Ferioney thanks for submitting the changes. Handling global properties needs to be further discussed and potentially some implementation added there (see comments). I have also requested a change in the way the client properties are prefixed and adding more docs (also please, rebuild the project with `-Pdocs` flag before pushing the changes - some documentation will be regenerated). 

As we are getting closer to merging these changes, there's one more thing to consider - there are other projects that use `LoadBalancerProperties` directly, mainly SC Gateway and SC OpenFeign, so this PR will need to be accompanied with corresponding PRs in those projects  - will you have time to work on it or do you want us to handle it once this one is merged?

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerAutoConfiguration.java
```diff
@@ -47,13 +48,22 @@
...
 @ConditionalOnBean(LoadBalancerClient.class)
-@EnableConfigurationProperties(LoadBalancerProperties.class)
+@EnableConfigurationProperties({ LoadBalancerProperties.class, LoadBalancerServiceProperties.class })
 public class LoadBalancerAutoConfiguration {
 
+	@Bean
+	@ConditionalOnMissingBean
+	public LoadBalancerPropertiesFactory loadBalancerPropertiesFactory(LoadBalancerProperties properties,
+			LoadBalancerServiceProperties servicesProperties,
+			@Value("${spring.cloud.loadbalancer.service.configuration.enabled:false}") boolean isServiceProperties) {
```
OlgaMaciaszek, line 63: Let's also add this property to `additional-spring-configuration-metadata.json`.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerServiceProperties.java
spencergibb: I agree

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
```diff
@@ -159,15 +175,24 @@ private RetryTemplate createRetryTemplate(String serviceName, HttpRequest reques
...
-		String hintPropertyValue = properties.getHint().get(serviceId);
+		String defaultHint = getLoadBalancerProperties(serviceId).getHint().getOrDefault("default", "default");
+		String hintPropertyValue = getLoadBalancerProperties(serviceId).getHint().get(serviceId);
 		return hintPropertyValue != null ? hintPropertyValue : defaultHint;
 	}
 
+	@Deprecated
+	private LoadBalancerProperties getLoadBalancerProperties(String serviceId) {
+		if (propertiesFactory != null) {
+			return propertiesFactory.getLoadBalancerProperties(serviceId);
```
spencergibb, line 192: I think this isn't quite right. If there isn't properties for a service, the properties object should be returned.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/LoadBalancerRetryPolicy.java
```diff
@@ -31,28 +32,78 @@
...
 	boolean canRetrySameServiceInstance(LoadBalancerRetryContext context);
 
+	/**
+	 * Return <code>true</code> to retry on the same service instance.
+	 * @param serviceId the serviceId for the retry operation
+	 * @param context the context for the retry operation
+	 * @return true to retry on the same service instance
+	 */
+	default boolean canRetrySameServiceInstance(String serviceId, LoadBalancerRetryContext context) {
```
spencergibb, line 47: Why couldn't `LoadBalancerRetryContext` have the serviceId so all these methods won't need to change

spencergibb (changes requested):
Some thoughts, not totally reviewed

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerClientAutoConfiguration.java
Ferioney: I'll add `getGlobalLoadBalancerProperties()` method to the factory. But I don't see where we can use it.

For caching we use a different property class `LoadBalancerCacheProperties.java`. Change cache classes to use the new way of configuration not enough, functionality also should be changed (`DefaultLoadBalancerCacheManager` and `CaffeineBasedLoadBalancerCacheManager`). I think it should be done in a separate thread.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java
Ferioney: Factory method returns `default` properties when service-specific properties were not provided.
```
private LoadBalancerProperties getLoadBalancerServiceProperties(String serviceName) {
	return servicesProperties.getClient().getOrDefault(serviceName, servicesProperties.getDefaultClientProperties());
}
```

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/LoadBalancerRetryPolicy.java
Ferioney: `LoadBalancerRetryContext` is created in `ReactorLoadBalancerClientAutoConfiguration`. At that time we don't have information about service names. So I decided to add new methods.

Ferioney:
@OlgaMaciaszek thank you for your answer. 
yes, it should be done in SC Gateway and SC OpenFeign. I can try to help with it.

OlgaMaciaszek:
@Ferioney upon consideration, I think @spencergibb idea makes more sense. It would also resolve the `global` properties issue - we would automatically use `per-client`, but if there's none then use the global one OOTB - we can then remove the `getGlobalProperties` method and add information to the docs, that per-client properties will override global properties, but we will fall back to global if per-client is not set. Let's modify in this way - sorry for making you change that part one more time.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/LoadBalancerRetryPolicy.java
OlgaMaciaszek: Yes, we don't have it there.

OlgaMaciaszek:
Regarding https://github.com/spring-cloud/spring-cloud-commons/pull/947#issuecomment-830117657, thanks a lot - that would be great :). Let's finish working on this one first, and then we could switch to the other ones.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/ReactorLoadBalancerClientAutoConfiguration.java
OlgaMaciaszek: Another option would be to rework the way the `ServiceInstanceListSupplier` beans are being created and, in place of using `@ConditionalOnProperty` during bean creation, use the properties bean/ factory passed in constructor and then create the bean, for example, `new RetryAwareServiceInstanceListSupplier(delegate)` (see [usage from current code](https://github.com/spring-cloud/spring-cloud-commons/blob/6ed50fcbc7ab9e41eac5eb4641cde952fa81823e/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/annotation/LoadBalancerClientConfiguration.java#L197-L199)) if the flag from properties is `true` and just return the delegate bean if it's `false`. Then we would be able to always give the highest priority for `spring.cloud.loadbalancer.client.xxx.property` and fall back to `spring.cloud.loadbalancer.property` consistently for all the features. @Ferioney @spencergibb please let me know what you think about the various possible approaches/ propose others.

OlgaMaciaszek:
@Ferioney - thanks a lot for the work that you've put with this PR. We've had a team discussion on this and we need some time to consider design decisions and also investigate some alternative approaches. We will get back to you once that's done.

spencergibb:
@Ferioney thanks again for the work. It led us in the right direction. @olegz from spring cloud stream gave us the missing piece. 

Closing in favor of #1028

Issue #953, opened 2021-05-18
Property type is ignored when loading encryption keystore


bberto:
Property `encrypt.key-store.type` is ignored when loading encryption keystore. The keystore type is retrieved from file extension (eg: .pkcs12).

Can be easily reproduced using these settings:
```
encrypt:
  key-store:
    type: PKCS12
    alias: encrypt
    location: classpath:encrypt.p12
    password: 123456
```

That leads to this exception:
```
java.lang.IllegalStateException: Cannot load keys from store: class path resource [encrypt.p12]
...
Caused by: java.security.KeyStoreException: p12 not found
...
Caused by: java.security.NoSuchAlgorithmException: p12 KeyStore not available
```

I think the root cause is here:
https://github.com/spring-cloud/spring-cloud-commons/blob/6ed50fcbc7ab9e41eac5eb4641cde952fa81823e/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/TextEncryptorUtils.java#L123

and can be solved using a different `KeyStoreKeyFactory` constructor (including type as parameter):
https://github.com/dsyer/spring-security-rsa/blob/d194f8969083af6987904f05eaf12b2dbae642a7/src/main/java/org/springframework/security/rsa/crypto/KeyStoreKeyFactory.java#L52

ryanjbaxter:
What version of Spring Cloud?

If you have a fix, would you be interested in contributing the fix?

bberto:
Tester with 2020.0.2 release train.

PR provided

PR #954, opened 2021-05-19
Support to PKCS12 keystores through encryption.keyStore.type property

bberto:
fixes gh-953


spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/KeyProperties.java
```diff
@@ -113,9 +113,9 @@ public void setKeyStore(KeyProperties.KeyStore keyStore) {
 		private String secret;
 
 		/**
-		 * The KeyStore type. Defaults to jks.
+		 * The KeyStore type (key store file extension used as default).
 		 */
-		private String type = "jks";
+		private String type;
```
ryanjbaxter, line 118: Why did you remove the default value?

ryanjbaxter (changes requested):

spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/KeyProperties.java
bberto: Currently this property is not used in `TextEncryptorUtils.createTextEncryptor` and the default behaviour is to infer the keystore type from the file extension.

Removing the default value this behaviour is preserved, avoiding a breaking change for who is using, for example, `encryption.key-store.location=classpath:test.jceks`  with no `encryption.key-store.type` set.

However, while writing this comment, I noticed a default value is expected here:
https://github.com/spring-cloud/spring-cloud-config/blob/e645c802157caa88a8ed50ecb9067335f0c7522f/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/config/EncryptionAutoConfiguration.java#L104

What do you suggest?

ryanjbaxter: Could you enable your added code below only if it is not the default value?

bberto: I think it's better to rollback to default value. The current behaviour (infer type from extension) is not documented anywhere and the error in case of keystore type mismatch is pretty clear.

bberto:
Sorry, I made a mess with this PR. I opened a new one:
https://github.com/spring-cloud/spring-cloud-commons/pull/963

Issue #956, opened 2021-05-19
Not retry request if service unavailable


CrazyTosser46:
Spring cloud version 3.0.1 have a bug, connected with retry request. According to document (https://docs.spring.io/spring-cloud-commons/docs/current/reference/html/#retrying-failed-requests) for retry request for specific http codes, need add codes to `spring.cloud.loadbalancer.retry.retryableStatusCodes`. I added 500 and 503 codes this block, but i got 503 http code from webclient.

I have a simple application-A with eureka client, which connect to eureka server. During work application try call other applicaction-B by webclient with loadBalancerClient, but application-B not a registred to eureka server. I expected retrying request, because set `spring.cloud.loadbalancer.retry.retryableStatusCodes` with 503, but got 503 http code.

OlgaMaciaszek:
Please provide a [minimal, complete, verifiable example](https://stackoverflow.com/help/mcve) that reproduces the issue.

CrazyTosser46:
Hi @OlgaMaciaszek, thank you for you attention.
I found mistake in config file, during create example.

PR #957, opened 2021-05-19
fix bug connected with 503 http code

CrazyTosser46:
#956


spencergibb:
See https://github.com/spring-cloud/spring-cloud-commons/issues/956#issuecomment-848199800

PR #974, opened 2021-06-30
Add a TaskSchedulerWrapper for downstream projects to use

ryanjbaxter:
See https://github.com/spring-cloud/spring-cloud-kubernetes/pull/817#pullrequestreview-686941108 or need for this class


spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/TaskSchedulerWrapper.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Wrapper that downstream projects can use to keep {@link ThreadPoolTaskScheduler} local.
+ *
+ * Implementation taken from Spring Vault.
+ *
+ */
+public class TaskSchedulerWrapper implements InitializingBean, DisposableBean {
+
+	private final ThreadPoolTaskScheduler taskScheduler;
```
spencergibb, line 32: Maybe make it either generic <? extends TaskScheduler> or use TaskScheduler interface

spencergibb (approved):

Issue #973, opened 2021-06-28
Switch remaining jUnit 4 tests to jUnit jupiter


OlgaMaciaszek:

spencergibb:
Assigning to @robertmcnees

deepakchethan:
Hello @spencergibb, Can I take this up if nobody is working on it already? Thanks in advance!

spencergibb:
Someone is assigned already

robertmcnees:
Moved back to open so more modules can be converted.

OlgaMaciaszek:
@ManasviGoyal is working on reducing method and class visibility after the conversion to JUnitJupiter.

timtebeek:
Gave it a quick try to convert more classes over to JUnit 5; outcome here:
https://github.com/spring-cloud/spring-cloud-commons/compare/main...timtebeek:increase-junit-platform-usage
This is mostly the result of applying the [JUnit4 to 5 conversion from OpenRewrite](https://docs.openrewrite.org/tutorials/migrate-from-junit-4-to-junit-5).

I had not yet created a merge request (although willing) as it appeared @ManasviGoyal was also working on this previously, and I don't know the current state of that effort.

I had dropped the Adhoc suites for now, as they are ignored currently, and converting them to disabled suites [would require additional depencies](https://junit.org/junit5/docs/current/user-guide/#junit-platform-suite-engine-example). Please let me know if these should be restored as they were or converted with the added dependencies.

There's quite a few tests left over that rely on the `ModifiedClassPathRunner`; is anyone aware of a similar mechanism for JUnit Jupiter? I'm aware of the [FilteredClassLoader](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.developing-auto-configuration.testing.overriding-classpath) when testing Auto-configuration, but I don't know it that would be applicable here.

timtebeek:
Any replacement for the `ModifiedClassPathRunner` could also help complete the Jupiter migration in https://github.com/spring-cloud/spring-cloud-gateway/pull/2387, so mentioning that here to have a feedback loop back to that issue as well.

timtebeek:
[...]

PR #975, opened 2021-07-07
Upgrade JUnit version (#973)

robertmcnees:
Module spring-cloud-context-integration-tests

### DIFF
--- a/spring-cloud-context-integration-tests/pom.xml
+++ b/spring-cloud-context-integration-tests/pom.xml
@@ -61,8 +61,8 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
-			<groupId>org.junit.vintage</groupId>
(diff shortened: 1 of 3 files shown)


spring-cloud-context-integration-tests/pom.xml
```diff
@@ -61,8 +61,8 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
-			<groupId>org.junit.vintage</groupId>
-			<artifactId>junit-vintage-engine</artifactId>
+			<groupId>org.junit.jupiter</groupId>
```
spencergibb, line 64: This should be removed instead of changed. I believe it is part of spring-boot-starter-test.

spencergibb (changes requested):
LGTM with the dependency note

spencergibb (approved):

Issue #906, opened 2020-09-24
Add support for query parameters in loadbalancer health-check


wpstan:
`spring.cloud.loadbalancer.health-check.path.default=/api/test?view=monitor`
When I configured the path with a special character `?` 。it was automatically converted to a URI encoding %3F,  causing the request to fail.

How to solve this problem?
thanks

spencergibb:
It is probably because of how we construct the full URL in `ServiceInstanceListSupplierBuilder` assumes the path is just that, a path, not a path and request parameters (which are very distinct from a URI perspective).

njoysubho:
Hello,
Is it fine if I take this issue and try to solve it ?

OlgaMaciaszek:
@njoysubho yes, please go ahead and submit a PR.

njoysubho:
[PR](https://github.com/spring-cloud/spring-cloud-commons/pull/982) submitted. Thanks.

OlgaMaciaszek:
Fixed with https://github.com/spring-cloud/spring-cloud-commons/commit/de32d2b68c6f980c44855406a4c4fdd6e20f2fb6.

PR #982, opened 2021-07-18
Issue-906 Adding support for query param in lb health-check

njoysubho:
This PR is to fix issue https://github.com/spring-cloud/spring-cloud-commons/issues/906
where loadbalancer healthcheck has query param and
character "?" was automatically
converted to a URI encoding %3F, causing the request to fail.

What has changed
* Using `UriComponentsBuilder.fromUriString` this method parses the string and handles url
  and query param path.

### DIFF
--- a/spring-cloud-loadbalancer/src/test/java/org/springframework/cloud/loadbalancer/core/ServiceInstanceListSuppliersTestUtils.java
+++ b/spring-cloud-loadbalancer/src/test/java/org/springframework/cloud/loadbalancer/core/ServiceInstanceListSuppliersTestUtils.java
@@ -23,6 +23,7 @@ import reactor.core.publisher.Mono;
 
 import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.http.HttpStatus;
+import org.springframework.util.StringUtils;
 import org.springframework.web.client.RestTemplate;
 import org.springframework.web.reactive.function.client.WebClient;
 import org.springframework.web.util.UriComponentsBuilder;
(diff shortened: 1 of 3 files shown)


spring-cloud-loadbalancer/src/test/java/org/springframework/cloud/loadbalancer/core/ServiceInstanceListSuppliersTestUtils.java
```diff
@@ -58,4 +59,12 @@ private ServiceInstanceListSuppliersTestUtils() {
 		});
 	}
 
+	private static String getUri(ServiceInstance serviceInstance, String healthCheckPath) {
```
OlgaMaciaszek, line 62: Let's make the method `getUri` in `ServiceInstanceListSupplierBuilder` package-private and reuse it in this test rather than duplicating the code in the test.

OlgaMaciaszek (changes requested):
@njoysubho Thanks for submitting this PR. Looks good in general, however, there are some minor changes to make:
- have added a comment - please address
- please add `@author` tag with your full name to the javadocs of the classes you've changed.

OlgaMaciaszek:
Fixes gh-906.

OlgaMaciaszek:
@njoysubho Since we are doing a `2021.0.x` release today and I would like to get your commits in before that, I'm going to cherry-pick your commit directly into the `main` branch adding the changes described in the comments.

OlgaMaciaszek:
Merged within https://github.com/spring-cloud/spring-cloud-commons/commit/de32d2b68c6f980c44855406a4c4fdd6e20f2fb6.

Issue #775, opened 2020-06-16
Clarify whether to use cache for Spring Cloud LoadBalancer with Eureka Discovery Client


stefanocke:
The documentation says at https://cloud.spring.io/spring-cloud-static/spring-cloud-commons/2.2.2.RELEASE/reference/html/#loadbalancer-cache-configuration:

> 

Although the basic, non-cached, implementation is useful for prototyping and testing, it’s much less efficient than the cached versions, so we recommend always using the cached version in production.
--

While this might be true in general, the cache is probably not necessary at all, if LoadBalancer is used together with Eureka Discovery Client, since it has its own cache. For details, see for example https://github.com/spring-cloud/spring-cloud-netflix/issues/373#issuecomment-110331739. The description there is for Ribbon + Eureka, but is IMHO also applicable for Spring Cloud LB.

Since Eureka is still the mainly supported case, it should be clarified whether to use cache is necessary. Cache might increase size / footprint of the application, especially since we get a log warning to include Caffeine instead of default one. 

Another point besides footprint is ZonePreferenceServiceInstanceListSupplier. AFAIU, this filter happens before caching. So, caching might increase the time until it is recognized that a zone has no healthy instances anymore and other zones are considered for routing.

spencergibb:
PRs welcome

OlgaMaciaszek:
thanks, @stefanocke  `Since Eureka is still the mainly supported case,` - this is not exactly the case (we treat Eureka equally as all supported Service Registries), however, I agree it makes sense to add these considerations to docs. Would you like to submit a PR?

stefanocke:
@OlgaMaciaszek , yes I would like too. 
But could you give me an hint what is your opinion on this matter? Would you suggest to disable caching in Spring Cloud LB in this case?

Even if Eureka is not the mainly supported case, it is certainly an important one.
So I guess you would have some suggestion.

OlgaMaciaszek:
I would actually still suggest using cache on LB side, just to avoid sending a request to Eureka each time and the associated waits. But we should add information on Eureka caching as a `NOTE` and let the users take it into account (they might choose to opt out of, reduce the cache times it or adjust eureka cache settings).

antechrestos:
@OlgaMaciaszek What is the expected change about that? Some upgrade of the documentation telling that applications using eureka discovery does not need adding cafeine as eureka discovery client has its own cache ? Maybe upgrade the start up warning not to be logged in eureka context?

[...]

PR #991, opened 2021-07-26
Added load-balancer cashe configuration warnings #775

Tiffany-Chien:
It was 

> Although the basic, non-cached, implementation is useful for prototyping and testing, it’s much less efficient than the cached versions, so we recommend always using the cached version in production.

Added load-balancer cashe configuration warnings about double caching and changed to:

> Although the basic, non-cached, implementation is useful for prototyping and testing, it’s much less efficient than the cached versions, so we recommend always using the cached version in production. If the caching is already done by the `DiscoveryClient ` implementation, for example `EurekaDiscoveryClient`, the load-balancer caching should be disabled to prevent double caching.


OlgaMaciaszek:
@Tiffany-Chien Have changed base branch for the PR to `3.0.x`, cause that is the branch that you started working from.

OlgaMaciaszek (approved):
@Tiffany-Chien Thanks for submitting the PR. LGTM.

OlgaMaciaszek:
Fixes gh-775.

Issue #1009, opened 2021-09-16
Provide possibility to configure retry properties per client


Ferioney:
We have a service that uses a few feign clients and we need to have a possibility to configure retry policy for each of them.
But now `Spring Cloud LoadBalancer` doesn't support configuration per client [Issue 914](https://github.com/spring-cloud/spring-cloud-commons/issues/914).

I propose to add possibility to configure retry properties per client.
It can be done by implementing a new `LoadBalancedRetryFactory`:
```
public class InfraBlockingLoadBalancedRetryFactory implements LoadBalancedRetryFactory {
    @Override
    public LoadBalancedRetryPolicy createRetryPolicy(String service, ServiceInstanceChooser serviceInstanceChooser) {
        final LoadBalancerProperties properties = here we can get properties specific for client as we have a client name

        return new BlockingLoadBalancedRetryPolicy(properties);
    }

}
```

Issue #914, opened 2021-03-04
No way to configure loadbalancer properties per client


Aloren:
**Is your feature request related to a problem? Please describe.**
When migrating our integrations from ribbon to loadbalancer, we faced with an issue there is not possibility to configure `spring.cloud.loadbalancer` properties per client basis. Current `LoadBalancerProperties` give an ability to configure loadbalancing for the whole application. In case application is integrated with multiple client -- this is an issue.

**Describe the solution you'd like**
We would like to have an ability to configure `spring.cloud.loadbalancer` per client as it was previously done in ribbon.

**Describe alternatives you've considered**
There is no way to set specific properties for specific client.

shsrams:
We are in the same boat, where we are migrating from Ribbon to LoadBalancer but there is no way to configure properties per client. More importantly, we want to turn on healthcheck only on certain clients but we could not find an option with 2020.0.1 version of spring cloud loadbalancer

HashZhang:
In same boat too. Well, I have migrated from Ribbon to Loadbalancer last year. The way I migrated is that use `@LoadBalancerClient`, for example:

```
@LoadBalancerClient(value = "service-1", configuration = Service1LoadBalancerConfig.class)
public class Service1LoadBalancerConfig {
    @Bean
    @Order(Integer.MIN_VALUE)
    public ServiceInstanceListSupplier discoveryClientServiceInstanceListSupplier(
            DiscoveryClient discoveryClient, Environment env,
            LoadBalancerZoneConfig zoneConfig,
            ApplicationContext context) {
        isValid = true;
        ServiceInstanceListSupplier delegate = new SameZoneOnlyServiceInstanceListSupplier(
                new DiscoveryClientServiceInstanceListSupplier(discoveryClient, env),
                zoneConfig
        );
        ObjectProvider<LoadBalancerCacheManager> cacheManagerProvider = context
                .getBeanProvider(LoadBalancerCacheManager.class);
        if (cacheManagerProvider.getIfAvailable() != null) {
            return new CachingServiceInstanceListSupplier(
                    delegate,
                    cacheManagerProvider.getIfAvailable()
            );
        }
        return delegate;
    }
}
```
In this way you can configuration customized specifications for each service.

And If you feel it is troublesome, I suggest you applying a default configuration using `LoadBalancerClients`. For example:
```
@LoadBalancerClients(defaultConfiguration = {DefaultLoadBalancerConfig.class})
[...]

PR #1010, opened 2021-09-16
Fixes gh-1009

Ferioney:
Implement client specific LoadBalancedRetryFactory

### DIFF
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/blocking/retry/ClientBlockingLoadBalancedRetryFactory.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/blocking/retry/ClientBlockingLoadBalancedRetryFactory.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2012-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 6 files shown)


OlgaMaciaszek:
@Ferioney  - thanks for submitting this. Will review on Monday.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/blocking/retry/ClientBlockingLoadBalancedRetryFactory.java
```diff
@@ -0,0 +1,52 @@
...
+
+	public ClientBlockingLoadBalancedRetryFactory(LoadBalancerProperties loadBalancerProperties,
+			ClientLoadBalancerProperties clientLoadBalancerProperties) {
+		this.loadBalancerProperties = loadBalancerProperties;
+		this.clientLoadBalancerProperties = clientLoadBalancerProperties;
+	}
+
+	@Override
+	public LoadBalancedRetryPolicy createRetryPolicy(String service, ServiceInstanceChooser serviceInstanceChooser) {
+		final LoadBalancerProperties properties = clientLoadBalancerProperties.getClientLoadBalancerProperties(service)
```
OlgaMaciaszek, line 46: @Ferioney , as discussed [here](https://github.com/spring-cloud/spring-cloud-commons/issues/914#issuecomment-840422056), we will wait with resolving properties per client in general till the feature is implemented in Boot, so the changes in this class should be removed. Specifcally, the issue here is that either all properties are resolved "per client" for a given client or none, while in the final version, we would want it to be possible to make that distinction on property level. For now, a simpler implementation only for the specific properties can be added to fix this issue, similarly to what is being done [here](https://github.com/spring-cloud/spring-cloud-commons/blob/779d482730dc2c8b988056e594ec664733e95010/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java#L66) and [here](https://github.com/spring-cloud/spring-cloud-commons/blob/779d482730dc2c8b988056e594ec664733e95010/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java#L130-L131).

OlgaMaciaszek (changes requested):
@Ferioney thanks for submitting the PR. Please see the comment I've added and adjust the PR.

PR #1028, opened 2021-10-22
Adds config per load balancer client.

spencergibb:
Client specific configuration goes under `spring.cloud.loadbalancer.clients.<serviceId>.*`. Defaults are set using `spring.cloud.loadbalancer.*`.

Fixes gh-914


Aloren:
Thank you! That was a long-awaited feature ❤️

Issue #1031, opened 2021-10-27
LoadBalancer: add custom conditionals to work with per-client config


OlgaMaciaszek:
https://github.com/spring-cloud/spring-cloud-commons/pull/1028#pullrequestreview-786936462

https://github.com/spring-cloud/spring-cloud-commons/pull/1028#issuecomment-949725869

PR #1033, opened 2021-10-28
Handle conditionals for client properties

OlgaMaciaszek:
Make properties other than within `@ConfigurationProperties` work per-client, with the following global properties exclusions:

- `spring.cloud.loadbalancer.enabled`
- `spring.cloud.loadbalancer.retry.enabled` - if enabled, still possible to disable per-client only with properties
- `spring.cloud.loadbalancer.cache.enabled` - if enabled, still possible to disable per-client with custom configuration
- `spring.cloud.loadbalancer.stats.micrometer.enabled`

Fixes gh-1031.


ryanjbaxter (approved):
Could probably use some unit tests of `LoadBalancerEnvironmentPropertyUtils` to make sure its behaving as expected

OlgaMaciaszek:
> Could probably use some unit tests of `LoadBalancerEnvironmentPropertyUtils` to make sure it's behaving as expected

Thanks :) I assumed it's sort of a self-tested feature that should be covered by pre-existing tests as we change the implementation without changing expected behaviour, but will add a few tests just for that class.

Issue #930, opened 2021-03-18
Add spring.profiles.active to IdUtils.getUnresolvedServiceId()


spencergibb:
See https://github.com/spring-cloud/spring-cloud-bus/issues/113 for details.

StitzL:
I have simply added the spring.profiles.active to IdUtils.DEFAULT_SERVICE_ID_STRING at 2nd position, with an empty default.

This has the downside of leaving the double-colon artifact "::" between application name and instance index in case no profiles are active.

Should that be unacceptable, I'd offer a more careful but also more complex solution directly in  IdUtils.getUnresolvedServiceId()

StitzL:
Finally found the time to submit an improved solution in #1064 building on your guidance from [your comments in #1046](https://github.com/spring-cloud/spring-cloud-commons/pull/1046#issuecomment-982668273)

PR #1046, opened 2021-11-25
Add spring.profiles.active to IdUtils.getUnresolvedServiceId()

StitzL:
Profiles added in second position.
Default: empty String.
Adapt unit tests.
Fixes gh-930.

### DIFF
--- a/spring-cloud-commons/src/test/java/org/springframework/cloud/commons/util/IdUtilsTests.java
+++ b/spring-cloud-commons/src/test/java/org/springframework/cloud/commons/util/IdUtilsTests.java
@@ -117,7 +117,7 @@ public class IdUtilsTests {
 	@Test
 	public void testServiceIdDefaults() {
 		this.env.setProperty("cachedrandom.application.value", "123abc");
-		then("application:0:123abc").isEqualTo(IdUtils.getResolvedServiceId(this.env));
+		then("application::0:123abc").isEqualTo(IdUtils.getResolvedServiceId(this.env));
 	}
 
 	@Test
@@ -125,7 +125,8 @@ public class IdUtilsTests {
 		env.setProperty("vcap.application.name", "vcapname");
 		env.setProperty("vcap.application.instance_index", "vcapindex");
 		env.setProperty("vcap.application.instance_id", "vcapid");
-		then("vcapname:vcapindex:vcapid").isEqualTo(IdUtils.getResolvedServiceId(env));
+		env.setProperty("spring.profiles.active", "123profile");
+		then("vcapname:123profile:vcapindex:vcapid").isEqualTo(IdUtils.getResolvedServiceId(env));
 	}
 
 	@Test
@@ -133,7 +134,8 @@ public class IdUtilsTests {
 		env.setProperty("spring.application.name", "springname");
 		env.setProperty("spring.application.index", "springindex");
 		env.setProperty("cachedrandom.springname.value", "123abc");
-		then("springname:springindex:123abc").isEqualTo(IdUtils.getResolvedServiceId(env));
+		env.setProperty("spring.profiles.active", "123profile");
+		then("springname:123profile:springindex:123abc").isEqualTo(IdUtils.getResolvedServiceId(env));
 	}
 
 	@Test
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
To avoid the double colon problem I wonder if we could check if there are profiles active and if so use one string and if not use the string we have today?

StitzL:
Sure! I would suggest to implement such a conditional in `IdUtils.getUnresolvedServiceId()`.

That would ensure backwards-compatibility of `IdUtils.DEFAULT_SERVICE_ID_STRING` - anyone who directly uses that public constant will not see the profiles-section.

StitzL:
Implementing a check for active Spring profiles in `IdUtils.getUnresolvedServiceId()` is not possible due to the fully static nature of `IdUtils`. So, strictly speaking, any such solution will no longer implement the requirement of gh-930. 

At a minimum, I'll have to overload `IdUtils.getUnresolvedServiceId()`, adding a second method implementation that takes a `PropertyResolver` as parameter to evaluate the condition. The pre-existing implementation of `IdUtils.getUnresolvedServiceId()` would call this new method with `null` (and should probably be deprecated). The only other local call site, `IdUtils.getResolvedServiceId(PropertyResolver)`, could pass on its `PropertyResolver` instance.

Alternatively, I could lift that condition to `IdUtils.getResolvedServiceId(PropertyResolver)`, where I can use the `PropertyResolver` to perform the check.

In both cases, `IdUtils.getUnresolvedServiceId()` will not benefit from the resolution of active Spring profiles.

Please let me know if any of these solutions seems to be a good fit for solving gh-930 despite its disadvantages.

ryanjbaxter:
Yeah makes sense.  Ultimately the original problem (https://github.com/spring-cloud/spring-cloud-bus/issues/113) cannot be fixed without either making a breaking change, or changing the code in spring cloud bus to call a different method.  IMO its better to probably fix the bug by adding another method and then making the change in Spring Cloud Bus.

What I think is we do the following:
1.) Introduce a new method called `getUnresolvedServiceIdWithActiveProfile` which returns the string with the active profile property
2.) Add some logic to `getResolvedServiceId` to pick the right string based on whether there is an active profile
3.) Deprecate `getUnresolvedServiceId`
4.) Make the change in Spring Cloud Bus to call `getUnresolvedServiceIdWithActiveProfile` if there is an active profile

@spencergibb what do you think?

StitzL:
Sounds very reasonable to me. I'll wait if @spencergibb has any other suggestions, but in the end I will submit a new PR anyways, as the described re-work is such a big change in approach that it doesn't even fit the summary of this one.

spring-cloud-commons/src/test/java/org/springframework/cloud/commons/util/IdUtilsTests.java
```diff
@@ -117,31 +117,34 @@ public void testUnresolvedServiceId() {
 	@Test
 	public void testServiceIdDefaults() {
 		this.env.setProperty("cachedrandom.application.value", "123abc");
-		then("application:0:123abc").isEqualTo(IdUtils.getResolvedServiceId(this.env));
+		then("application::0:123abc").isEqualTo(IdUtils.getResolvedServiceId(this.env));
```
spencergibb, line 120: I'm a bit worried about this. This value gets parsed in some places and this would break things.

ryanjbaxter:
@StitzL I spoke to Spencer and he is on board with what I proposed here https://github.com/spring-cloud/spring-cloud-commons/pull/1046#issuecomment-982668273

StitzL:
Okay, sure. As I mentioned, I will open another PR for that changeset, hopefully this week.

StitzL:
I have re-submitted an improved solution that follows your guidance from the discussion above as #1064

Been24:
> 



> Yeah makes sense. Ultimately the original problem ([spring-cloud/spring-cloud-bus#113](https://github.com/spring-cloud/spring-cloud-bus/issues/113)) cannot be fixed without either making a breaking change, or changing the code in spring cloud bus to call a different method. IMO its better to probably fix the bug by adding another method and then making the change in Spring Cloud Bus.
> 
> What I think is we do the following: 1.) Introduce a new method called `getUnresolvedServiceIdWithActiveProfile` which returns the string with the active profile property 2.) Add some logic to `getResolvedServiceId` to pick the right string based on whether there is an active profile 3.) Deprecate `getUnresolvedServiceId` 4.) Make the change in Spring Cloud Bus to call `getUnresolvedServiceIdWithActiveProfile` if there is an active profile
> 
> @spencergibb what do you think?

Hi, what about the 4th step?

ryanjbaxter:
See https://github.com/spring-cloud/spring-cloud-bus/commit/aa9a485abf1d81995ce3d3160020e3793ef869a4

PR #1054, opened 2021-12-16
Support uri without port for default service instance

neiljpowell:
The current DefaultServiceInstance getUri() method does not support URIs without a port. This is problematic when using the Spring Cloud Loadbalancer for DNS subdomain-based targets. For example service instances in different K8 clusters/CF foundations. This pull request adds a simple condition for constructing the uri string based on the port being > 0.

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
@@ -80,13 +80,15 @@ public class DefaultServiceInstance implements ServiceInstance {
 	/**
 	 * Creates a URI from the given ServiceInstance's host:port.
 	 * @param instance the ServiceInstance.
-	 * @return URI of the form (secure)?https:http + "host:port" or "host" (port<=0).
+	 * @return URI of the form (secure)?https:http + "host:port". Scheme port default used if port not set.
 	 */
 	public static URI getUri(ServiceInstance instance) {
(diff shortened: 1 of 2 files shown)


spencergibb:
There's always a port. If it's http it's 80 if it's https it's 443

neiljpowell:
@spencergibb Thanks for the quick response. Always a port true but the absence of a port allows use of the standard scheme defaults. I actually noticed when testing the HealthCheckServiceInstanceSupplier.  If a port was not explicitly set, it resulted in a hard to troubleshoot error message resulting from a failed isAlive() (https://github.com/spring-cloud/spring-cloud-commons/blob/8807b4dac035c2247987d22065ac03b8c86777e3/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/HealthCheckServiceInstanceListSupplier.java#L121).  This fixed the issue and removed the need for explicit port setting.

neiljpowell:
@spencergibb Alternatively, a condition can simply set the port to the scheme default, if not provided. That would align with an expectation of it being present and eliminate having to explicitly set it. The DefaultServiceInstance is a handy class. I suspect others might like it defaulting the port too.  Would you prefer/accept that approach?

spencergibb:
Yes, I'd prefer that approach

neiljpowell:
@spencergibb Thanks for the feedback. Let me know what you think now.

OlgaMaciaszek (approved):
LGTM. @spencergibb wdyt?

PR #1057, opened 2022-01-06
Improve readability of method "isRetryException"

stefanbirkner:
The "isRetryExhausted" check only depends on the throwable itself. The
new code makes this more clear because the check is no longer part of
anyMatch's lambda expression. I also extracted the method
"isInstanceOfRetryType" because it makes it obvious that we do the same
check on the throwable itself and its cause.

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/RetryableLoadBalancerExchangeFilterFunction.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/RetryableLoadBalancerExchangeFilterFunction.java
@@ -66,7 +66,7 @@ public class RetryableLoadBalancerExchangeFilterFunction implements LoadBalanced
 
 	private static final Log LOG = LogFactory.getLog(RetryableLoadBalancerExchangeFilterFunction.class);
 
-	private static final List<Class<? extends Throwable>> exceptions = Arrays.asList(IOException.class,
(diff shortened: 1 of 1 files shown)


spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/RetryableLoadBalancerExchangeFilterFunction.java
```diff
@@ -229,10 +229,15 @@ private boolean shouldRetryNextServiceInstance(LoadBalancerRetryPolicy retryPoli
...
-						|| Exceptions.isRetryExhausted(throwable));
+		if (throwable == null) {
+			return false;
+		}
+		return isInstanceOfRetryType(throwable) || isInstanceOfRetryType(throwable.getCause())
+				|| Exceptions.isRetryExhausted(throwable);
+	}
+
+	private boolean isInstanceOfRetryType(Throwable throwable) {
+		return typesThatTriggerRetry.stream().anyMatch(type -> type.isInstance(throwable));
```
stefanbirkner, line 240: I would like to replace it with 
```suggestion
		return throwable instanceof IOException
		        ||  throwable instanceof TimeoutException
		        ||  throwable instanceof RetryableStatusCodeException;
```
because it is simpler and it moves the information about the types close to where they are used. For me that makes the code easier to read. What do you think?

OlgaMaciaszek:
Closing due to lack of activity.

PR #1064, opened 2022-02-04
Consider spring.profiles.active when resolving service ID.

StitzL:
New constant IdUtils.DEFAULT_SERVICE_ID_WITH_ACTIVE_PROFILES_STRING refers to profiles in second position.
IdUtils.getResolvedServiceId chooses new constant as template only if profiles are active, and thus remains backwards-compatible.
Add unit tests for new behaviour.
Fixes gh-930.

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/IdUtils.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/IdUtils.java
@@ -29,6 +29,7 @@ public final class IdUtils {
 	// @checkstyle:off
 	public static final String DEFAULT_SERVICE_ID_STRING = "${vcap.application.name:${spring.application.name:application}}:${vcap.application.instance_index:${spring.application.index:${local.server.port:${server.port:0}}}}:${vcap.application.instance_id:${cachedrandom.${vcap.application.name:${spring.application.name:application}}.value}}";
 
+	public static final String DEFAULT_SERVICE_ID_WITH_ACTIVE_PROFILES_STRING = "${vcap.application.name:${spring.application.name:application}:${spring.profiles.active}}:${vcap.application.instance_index:${spring.application.index:${local.server.port:${server.port:0}}}}:${vcap.application.instance_id:${cachedrandom.${vcap.application.name:${spring.application.name:application}}.value}}";
 	// @checkstyle:on
 
 	private IdUtils() {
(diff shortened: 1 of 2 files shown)


StitzL:
This solution was created with your guidance in [a comment of #1046](https://github.com/spring-cloud/spring-cloud-commons/pull/1046#issuecomment-982668273). I'll be happy to take a shot at the remaining changes for spring-cloud-bus as well.

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/IdUtils.java
```diff
@@ -29,6 +29,7 @@
 	// @checkstyle:off
 	public static final String DEFAULT_SERVICE_ID_STRING = "${vcap.application.name:${spring.application.name:application}}:${vcap.application.instance_index:${spring.application.index:${local.server.port:${server.port:0}}}}:${vcap.application.instance_id:${cachedrandom.${vcap.application.name:${spring.application.name:application}}.value}}";
```
ryanjbaxter, line 30: Mark this as `@Deprecated` so we can remove it in the future

ryanjbaxter (approved):

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/IdUtils.java
```diff
@@ -64,7 +65,13 @@ public static String getDefaultInstanceId(PropertyResolver resolver, boolean inc
 	 * @return A unique id that can be used to uniquely identify a service
 	 */
 	public static String getResolvedServiceId(PropertyResolver resolver) {
-		return resolver.resolvePlaceholders(getUnresolvedServiceId());
```
ryanjbaxter, line 67: Probably worth adding a comment to the code explaining why this logic is here and that we should remove this method in the future in favor of `getUnresolvedServiceIdWithActiveProfile`.  Mark this method as `@Deprecated`

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/IdUtils.java
StitzL: Resolved by a51f4b2.

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/IdUtils.java
StitzL: Resolved by a51f4b2.

StitzL:
Is the case that `spring.profiles.active` is set to an empty String a possibility that I have to consider in [IdUtils.java#L75](https://github.com/spring-cloud/spring-cloud-commons/blob/07ae351d1c2b39629e85e94744aa4b0310e128f6/spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/IdUtils.java#L75)?

ryanjbaxter:
@StitzL In the future, your new property resolver string should take care of that case, no?

StitzL:
Yes, to an extent, as the passing test below demonstrates. An empty String value for `spring.profiles.active` will result in e.g. `"springname::1234:123abc"`, as the new method is used. When `spring.profiles.active` has not been set at all, and the old method is used, the result will be `"springname:1234:123abc"`.

	@Test
	public void testServerPortServiceIdWithActiveProfile() {
		env.setProperty("spring.application.name", "springname");
		env.setProperty("server.port", "1234");
		env.setProperty("cachedrandom.springname.value", "123abc");
		then("springname:1234:123abc").isEqualTo(IdUtils.getResolvedServiceId(env));

		env.setProperty("spring.profiles.active", "");
		then("springname::1234:123abc").isEqualTo(IdUtils.getResolvedServiceId(env));
	}

I could account for that in the condition in [IdUtils.java#L75](https://github.com/spring-cloud/spring-cloud-commons/blob/07ae351d1c2b39629e85e94744aa4b0310e128f6/spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/IdUtils.java#L75), but then the old method and constant cannot be removed, and tokenizing by `:` will not work reliably. 

I just wanted to point it out, but in my eyes this is a small unevenness we'll have to live with.

ryanjbaxter:
Ah yes, you can revert those deprecations then, sorry for the confusion.

spencergibb:
I'd like to avoid the double colon output of `springname::1234:123ab`.

StitzL:
Okay, with 91aaf03 I made sure that regardless whether `spring.profiles.active` has not been set at all, or has been set to an empty String value, no double colon appears, as in both cases, the old template is chosen. Since this means the old code has to stay, I have removed the deprecations again, but added a test that asserts this behaviour. 

Please let me know what you think.

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/IdUtils.java
StitzL: Superseded by 91aaf03.

spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/IdUtils.java
StitzL: Superseded by 91aaf03.

PR #1078, opened 2022-03-09
Remove unnecessary sign check, whether positive or negative

jizhuozhi:
The sign bit is ignored, this allows pos to loop sequentially from 0 to Integer.MAX_VALUE.
So there is no need to check whether positive or negative.


jizhuozhi:
This way we can reduce instruction cycles in any case

jizhuozhi:
@OlgaMaciaszek Hello, PTAL. 
I found there is no need to check the sign of pos, whether positive or negative. And the extra checks come with more overhead.

jizhuozhi:
Just like the same code written in C, and compiled with `gcc` and `gcc -O3`.

OlgaMaciaszek (approved):
Thanks for submitting the PR @jizhuozhi. LGTM.

PR #1101, opened 2022-04-24
add override annotation for Specification child class

mroccyen:
add override annotation for Specification child class


OlgaMaciaszek (approved):
Thanks, @mroccyen. LGTM.

PR #1107, opened 2022-05-17
Fix LocalServerPort import path

onionpsy:
The path should be `org.springframework.boot.test.web.server.LocalServerPort` instead of `org.springframework.boot.web.server.LocalServerPort`

See https://github.com/spring-projects/spring-boot/commit/48f8021d88ef0591d7b931b2a9bc6d4d48247d58


onionpsy:
Bug has been fixed in the meantime https://github.com/spring-cloud/spring-cloud-commons/commit/c2a22346cde819971263c47b321a81a7908c4f40

Issue #1063, opened 2022-02-03
Weight-based LoadBalancer strategy


mario45211:
Hi Team,

**Is your feature request related to a problem? Please describe.**
I want to load balance traffic between target service's instances using tag and corresponding weight, provided by the configuration. That feature would help me to freely redirect significant part of traffic to a group of instances while the rest to e.g. instances with new feature. The best example of such strategy of LB is blue-green deployment.

**Describe the solution you'd like**
```
spring:
  cloud:
    consul:
      discovery:
        server-list-query-tags: # probably not the best config key but it's only example 
          service-name:
            - tag: group-1
              weight: 80
            - tag: group-2
              weigth: 20 
```

**Additional context**
Example of such implemention in Consul splitting configuration:
```hcl
Kind = "service-splitter"
Name = "<service-name>"
Splits = [
  {
    Weight        = 10
    ServiceSubset = "blue"
  },
  {
    Weight        = 90
    ServiceSubset = "green"
  },
]

```
Please, let me know what you think about such feature.

Best regards

OlgaMaciaszek:
@jizhuozhi  is working on this issue.

jizhuozhi:
> @jizhuozhi is working on this issue.

Received👌

PR #1103, opened 2022-05-02
Enable config ServiceInstance weight and provide WeightedLoadBalancer

jizhuozhi:
#1102

Issue #1102, opened 2022-04-30
Provide a mechanism to configure the weight of server, and a weighted-loadbalancing strategy 


jizhuozhi:
**Is your feature request related to a problem? Please describe.**
I want to configure my application's weight and the client load balancer could use 
weighted-strategy to choose the ServiceInstance.

There is a common use case, we have three group ECS servers, having 1C4G, 2C4G and 4C4G 
(assuming they have the same number of machines). With the diferent number of CPU cores, 
the number of requests they can handle is different, so we would like to allocate 1/7 traffic to 
the 1C4G group, 2/7 traffic to the 2C4G group and 4/7 traffic to the 4C4G group.

**Describe the solution you'd like**
I can accept both configuration methods
1. using `spring.application.weight`
```yml
spring:
  application:
    name: application
    weight: 100
```
2. using `spring.cloud.{registry}.discovery.metadata.weight`
```yml
spring:
  cloud:
    consul:
      discovery:
        metadata:
          weight: 100
```
If not configured, using 100 by default.
And then, in the loadbalancer we provide a new loadbalancer named as `WeightedRoundRobinLoadBalancer` 
or just `WeightedLoadBalancer` using EDF scheduling algorithm.

jizhuozhi:
There are a series of weighted selection algorithms
1. EDF(Earliest Deadline First) algorithm with PriorityQueue

    It's using a PriorityQueue to maintain the next access sequence, the front of the queue is the next element. It's using O(nlogn) to build heap, every time a request is received using O(1) to peek the next element and using O(logn) to maintain the next access sequence (fixDown).

    But it has a limitation, the elements must be stable. If elements was changed, we must add to or delete from the queue. In spring-cloud-loadbalancer, the actual ServiceInstance list from the supplier is related to the request, it's unstable and hard to known which one was added or deleted. So the another solution is drop the older queue and rebuild a new queue with given elements.

2. EDF(Earliest Deadline First) algorithm without PriorityQueue

    It's not using any data struct to maintain the next access sequence, but visiting the ServiceInstance list and calculate the Earliest Deadline element. Every time a request is received using O(n) to peek the next element and O(1) to recalculate the deadline of the element.

    It seems better in unstable condition, but should modify the ServiceInstance interface to record the deadline.

3. Random Binary Search algorithm

[...]

PR #1111, opened 2022-05-19
Enable config ServiceInstance weight and provide WeightedLoadBalancer

jizhuozhi:
@OlgaMaciaszek Hello, I am resubmitting a new PR from #1103 , which all commits cherry picked and squashed

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
@@ -20,6 +20,7 @@ import java.net.URI;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Default implementation of {@link ServiceInstance}.
@@ -28,6 +29,7 @@ import java.util.Objects;
  * @author Tim Ysewyn
  * @author Charu Covindane
  * @author Neil Powell
+ * @author Zhuozhi Ji
  */
 public class DefaultServiceInstance implements ServiceInstance {
 
@@ -45,6 +47,30 @@ public class DefaultServiceInstance implements ServiceInstance {
 
 	private URI uri;
 
+	private int weight;
+
+	private final AtomicInteger currentWeight = new AtomicInteger(0);
+
+	/**
+	 * @param instanceId the id of the instance.
+	 * @param serviceId the id of the service.
+	 * @param host the host where the service instance can be found.
+	 * @param port the port on which the service is running.
+	 * @param secure indicates whether or not the connection needs to be secure.
+	 * @param metadata a map containing metadata.
+	 * @param weight weight of the instance.
+	 */
+	public DefaultServiceInstance(String instanceId, String serviceId, String host, int port, boolean secure,
+			Map<String, String> metadata, int weight) {
+		this.instanceId = instanceId;
+		this.serviceId = serviceId;
+		this.host = host;
+		this.port = port;
+		this.secure = secure;
+		this.metadata = metadata;
+		this.weight = weight;
+	}
+
 	/**
 	 * @param instanceId the id of the instance.
 	 * @param serviceId the id of the service.
@@ -55,12 +81,7 @@ public class DefaultServiceInstance implements ServiceInstance {
 	 */
 	public DefaultServiceInstance(String instanceId, String serviceId, String host, int port, boolean secure,
 			Map<String, String> metadata) {
-		this.instanceId = instanceId;
-		this.serviceId = serviceId;
-		this.host = host;
-		this.port = port;
-		this.secure = secure;
-		this.metadata = metadata;
+		this(instanceId, serviceId, host, port, secure, metadata, 0);
 	}
 
 	/**
@@ -124,6 +145,16 @@ public class DefaultServiceInstance implements ServiceInstance {
 	}
 
 	@Override
+	public int getWeight() {
+		return weight;
+	}
+
+	@Override
+	public AtomicInteger getCurrentWeight() {
+		return currentWeight;
+	}
+
+	@Override
 	public boolean isSecure() {
 		return secure;
 	}
@@ -154,10 +185,15 @@ public class DefaultServiceInstance implements ServiceInstance {
 		}
 	}
 
+	public void setWeight(int weight) {
+		this.weight = weight;
+	}
+
 	@Override
 	public String toString() {
 		return "DefaultServiceInstance{" + "instanceId='" + instanceId + '\'' + ", serviceId='" + serviceId + '\''
-				+ ", host='" + host + '\'' + ", port=" + port + ", secure=" + secure + ", metadata=" + metadata + '}';
+				+ ", host='" + host + '\'' + ", port=" + port + ", secure=" + secure + ", metadata=" + metadata
+				+ ", weight=" + weight + ", currentWeight=" + currentWeight + '}';
 	}
 
 	@Override
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/ServiceInstance.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/ServiceInstance.java
@@ -18,12 +18,14 @@ package org.springframework.cloud.client;
 
 import java.net.URI;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Represents an instance of a service in a discovery system.
  *
  * @author Spencer Gibb
  * @author Tim Ysewyn
+ * @author Zhuozhi Ji
  */
 public interface ServiceInstance {
 
@@ -69,6 +71,20 @@ public interface ServiceInstance {
 	 */
 	default String getScheme() {
 		return null;
+	}
+
+	/**
+	 * @return The raw weight of the service instance.
+	 */
+	default int getWeight() {
+		return 0;
+	}
+
+	/**
+	 * @return The current weight of the service instance.
+	 */
+	default AtomicInteger getCurrentWeight() {
+		return new AtomicInteger(0);
 	}
 
 }
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/WeightedLoadBalancer.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/WeightedLoadBalancer.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2012-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.loadbalancer.core;
+
+import java.util.List;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import reactor.core.publisher.Mono;
+
+import org.springframework.beans.factory.ObjectProvider;
+import org.springframework.cloud.client.ServiceInstance;
+import org.springframework.cloud.client.loadbalancer.DefaultResponse;
+import org.springframework.cloud.client.loadbalancer.EmptyResponse;
+import org.springframework.cloud.client.loadbalancer.Request;
+import org.springframework.cloud.client.loadbalancer.Response;
+
+/**
+ * A Weighted implementation of {@link ReactorServiceInstanceLoadBalancer}.
+ *
+ * @author Zhuozhi JI
+ */
+public class WeightedLoadBalancer implements ReactorServiceInstanceLoadBalancer {
+
+	private static final Log log = LogFactory.getLog(WeightedLoadBalancer.class);
+
+	final String serviceId;
+
+	ObjectProvider<ServiceInstanceListSupplier> serviceInstanceListSupplierProvider;
+
+	/**
+	 * @param serviceInstanceListSupplierProvider a provider of
+	 * {@link ServiceInstanceListSupplier} that will be used to get available instances
+	 * @param serviceId id of the service for which to choose an instance
+	 */
+	public WeightedLoadBalancer(ObjectProvider<ServiceInstanceListSupplier> serviceInstanceListSupplierProvider,
+			String serviceId) {
+		this.serviceInstanceListSupplierProvider = serviceInstanceListSupplierProvider;
+		this.serviceId = serviceId;
+	}
+
+	@SuppressWarnings("rawtypes")
+	@Override
+	public Mono<Response<ServiceInstance>> choose(Request request) {
+		ServiceInstanceListSupplier supplier = serviceInstanceListSupplierProvider
+				.getIfAvailable(NoopServiceInstanceListSupplier::new);
+		return supplier.get(request).next()
+				.map(serviceInstances -> processInstanceResponse(supplier, serviceInstances));
+	}
+
+	private Response<ServiceInstance> processInstanceResponse(ServiceInstanceListSupplier supplier,
+			List<ServiceInstance> serviceInstances) {
+		Response<ServiceInstance> serviceInstanceResponse = getInstanceResponse(serviceInstances);
+		if (supplier instanceof SelectedInstanceCallback && serviceInstanceResponse.hasServer()) {
+			((SelectedInstanceCallback) supplier).selectedServiceInstance(serviceInstanceResponse.getServer());
+		}
+		return serviceInstanceResponse;
+	}
+
+	private Response<ServiceInstance> getInstanceResponse(List<ServiceInstance> instances) {
+		if (instances.isEmpty()) {
+			if (log.isWarnEnabled()) {
+				log.warn("No servers available for service: " + serviceId);
+			}
+			return new EmptyResponse();
+		}
+
+		ServiceInstance best = null;
+		int bestCurrentWeight = 0;
+		int total = 0;
+
+		for (ServiceInstance instance : instances) {
+			int weight = instance.getWeight();
+			int currentWeight = instance.getCurrentWeight().addAndGet(weight);
+			total += weight;
+
+			// Use bestCurrentWeight to ensure strictly increasing
+			if (best == null || currentWeight > bestCurrentWeight) {
+				best = instance;
+				bestCurrentWeight = currentWeight;
+			}
+		}
+
+		best.getCurrentWeight().addAndGet(-total);
+
+		return new DefaultResponse(best);
+	}
+
+}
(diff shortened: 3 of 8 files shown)


OlgaMaciaszek:
Fixes gh-1063.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
```diff
@@ -45,22 +47,41 @@ public class DefaultServiceInstance implements ServiceInstance {
 
 	private URI uri;
 
+	private int weight;
```
OlgaMaciaszek, line 50: In keeping with the response to https://github.com/spring-cloud/spring-cloud-commons/issues/1113, in order to ensure that these values (weight and current weight) are being passed back and forth between the service registry and the LB clients, they should be kept and updated in the metadata.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/ServiceInstance.java
```diff
@@ -71,4 +73,18 @@ default String getScheme() {
 		return null;
 	}
 
+	/**
+	 * @return The raw weight of the service instance.
+	 */
+	default int getWeight() {
```
OlgaMaciaszek, line 79: New interface methods should not be necessary - we should rely on metadata entries.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/WeightedLoadBalancer.java
```diff
@@ -0,0 +1,103 @@
...
+	@Override
+	public Mono<Response<ServiceInstance>> choose(Request request) {
+		ServiceInstanceListSupplier supplier = serviceInstanceListSupplierProvider
+				.getIfAvailable(NoopServiceInstanceListSupplier::new);
+		return supplier.get(request).next()
+				.map(serviceInstances -> processInstanceResponse(supplier, serviceInstances));
+	}
+
+	private Response<ServiceInstance> processInstanceResponse(ServiceInstanceListSupplier supplier,
+			List<ServiceInstance> serviceInstances) {
```
OlgaMaciaszek, line 66: Looks good.

OlgaMaciaszek (changes requested):
Hello, @jizhuozhi , have added some comments -  please adjust. Also, the build fails on checkstyle - please fix.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/ServiceInstance.java
jizhuozhi: But there is a problem if using metadata directly, weight is a number but the value type of metadata is String, so we will do unnecessary `Integer.parseInt` cause each time we will get the same value.

spring-cloud-loadbalancer/src/test/java/org/springframework/cloud/loadbalancer/core/LoadBalancerBenchmarkTests.java
```diff
@@ -0,0 +1,164 @@
...
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Zhuozhi Ji
+ */
+@State(Scope.Thread)
+public class LoadBalancerBenchmarkTests {
```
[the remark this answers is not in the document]
jizhuozhi: I will fix that

spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
jizhuozhi: In my original thought, weight must be by `final` to ensure the weight of entry is immutable, but it's not because I chose to keep the style consistent with the other fields. I will response more in the issue.

jizhuozhi:
Hello @OlgaMaciaszek, comments and checkstyle are fixed, but there is still a points of contention: 

    Should we add an interface method `getWeight` and use field `weight` to record the weight of `ServiceInstance`, but not use `metadata` directly.

In the earliest version, I had tried use `metadata` directly. But as we can see, in our current use cases, `metadata` is never updated, so each time `metadata.getOrDefault("weight", "0")` called will get same **string value**, to use it in LoadBalancer we must use `	int weight = Integer.parseInt(instance.getMetadata().getOrDefault("weight", "0"));` to parse to **int value**, it's slowly and unnecessary.

As for why could user set `weight` in constructor and setter but not always read from `metadata`, is considered that `LoadBalancer` could be static (no service discovery). In this condition, if we request must use `metadata` to pass `weight`, the user side code will be complicated and ugly

```Java
// Set weight with metadata
var metadata = new HashMap<>();
metadata.put("weight", "100");
// var metadata = Collections.singletonMap("weight", "100");
var instance = new DefaultServiceInstance("instanceId", "serviceId", "localhost", 80, false, metadata);

// Set weight without metadata
var instance = new DefaultServiceInstance("instanceId", "serviceId", "localhost", 80, false, metadata, 100);
```

spencergibb:
We should create a new interface and implementation, that way we don't use metadata and don't pollute the base interface and default impl

spencergibb:
We should create a new interface and implementation, that way we don't use metadata and don't pollute the base interface and default impl

jizhuozhi:
> We should create a new interface and implementation, that way we don't use metadata and don't pollute the base interface and default impl

Hello, @spencergibb , in the earliest version, I had tried this solution by providing another interface named [`WeightedServiceInstance`](https://github.com/spring-cloud/spring-cloud-commons/pull/1103/commits/f9ccc10ec9d85df1e312778ec456c6de084350b1#diff-1516eef08e1b0adc8700f3c027da2afdd4fdcaadc3a855855933b211c32da35d), but I don't know what to do when ServiceInstance is not an instance of this interface. Here is the commit link(https://github.com/spring-cloud/spring-cloud-commons/pull/1103/commits/f9ccc10ec9d85df1e312778ec456c6de084350b1).

OlgaMaciaszek:
The `DefaultServiceInstance` probably should not have any methods regarding weight. Maybe the `WeightedLoadBalancer` could handle the instances and use some default behaviours when the instance is not a `WeightedServiceInstance`? Also, please include a proposal defining the instance weight by the users (via properties?) to the PR.

jizhuozhi:
> The `DefaultServiceInstance` probably should not have any methods regarding weight. Maybe the `WeightedLoadBalancer` could handle the instances and use some default behaviours when the instance is not a `WeightedServiceInstance`? Also, please include a proposal defining the instance weight by the users (via properties?) to the PR.

Hello, @OlgaMaciaszek. This may not be the main use case for Spring Cloud.

Some traditional enterprises do not have complete cloud infrastructure, including service registration, service discovery and api gateway, or they use RMI-style registry, but do not support RESTful-style service registration (this is what I have experienced). However, when integrating between such systems, it is necessary to avoid single points of failure through load balancing. 

Here is an example of the code, I used ConfigurationProperties to autowire from a yml file
```Java
@Data
@Component
@ConfigurationProperties(prefix = "service")
public static class ServiceProperties {
    private Map<String, List<DefaultServiceInstance>> instances;
}
```

```yml
service:
  instances:
    hello-world:
      - instance-id: instance-1
        service-id: hello-world
        host: 127.0.0.1
        port: 8080
        weight: 200 # 2 cores
      - instance-id: instance-2
        service-id: hello-world
        host: 127.0.0.2
        port: 8080
        weight: 400 # 4 cores
```

After that we can create a LoadBalancer Bean that gets the name from the environment and the list corresponding to the name from our properties
```Java
    @Bean
    ReactorLoadBalancer<ServiceInstance> weightedLoadBalancer(Environment environment,
                                                            ServiceProperties serviceProperties) {
        String name = environment.getProperty(LoadBalancerClientFactory.PROPERTY_NAME);
        List<ServiceInstance> instances = new ArrayList<>(serviceProperties.instances.get(name));
        return new WeightedLoadBalancer(new SimpleObjectProvider<>(new StaticServiceInstanceListSupplier(name, instances)), name);
    }

    class StaticServiceInstanceListSupplier implements ServiceInstanceListSupplier {
        private final String serviceId;

        private final List<ServiceInstance> instances;

        StaticServiceInstanceListSupplier(String serviceId, List<ServiceInstance> instances) {
            this.serviceId = serviceId;
            this.instances = instances;
        }

        @Override
        public String getServiceId() {
            return serviceId;
        }

        @Override
            return Flux.just(instances);
        }
    }
```

jizhuozhi:
> The `DefaultServiceInstance` probably should not have any methods regarding weight. Maybe the `WeightedLoadBalancer` could handle the instances and use some default behaviours when the instance is not a `WeightedServiceInstance`? Also, please include a proposal defining the instance weight by the users (via properties?) to the PR.

I thought about it, because all instances are offered by the same supplier, we can assume that all weighted or unweighted. So we could peek the first element to check whether weighted, if not we could delegate to `RoundRobinLoadBalancer` to do unweighted choose.

jizhuozhi:
> > Here is an example of the code, I used ConfigurationProperties to autowire from a yml file
> 
> We'll need to make it usable for all our users before merging, so the PR should include changes that will allow for mapping `weight` values provided by the users in properties for instances (based on `instanceId`, but I am thinking based on zone or on other specific metadata value might be useful) without them having to create their own configurations; let me know if you have any idea on how to handle this - if not I could try come up with sth.

How to get/set weights from the registry depends on the registry and discovery client implementation, including Spring Cloud Eureka, Spring Cloud Consul and Spring Cloud Kubernetes.

Here is an example that [apisix working with Eureka](https://apisix.apache.org/docs/apisix/discovery/#how-convert-eurekas-instance-data-to-apisixs-node), it uses `eureka.instance.metadata-map.weight`. As my original issue #1102 , it uses `metadata` to set weight of the service instance when instance started.

OlgaMaciaszek:
> This is not part of this PR. That the code snippet just a simple what user would be when using WeightedLoadBalancer without service registry and discovery client.

Yes, and it's something that needs to be implemented (use with SR) before we merge it.

jizhuozhi:
Hello, @OlgaMaciaszek . I have a better idea.

As @spencergibb said, we can provide a `WeightedServiceInstance` interface and `DefaultWeightedServiceInstance` as default implementation, which all behaviors are delegated but weight related not. 

And then, maybe we don't need a `LoadBalancer` to choose the best weight/currentWeight instance, but we could move to `WeightedServiceInstanceListSupplier`. `WeightedServiceInstanceListSupplier` has a final field `Function<ServiceInstance, Integer> weightFunction`, which used to calculate the weight of instance, and if `weightFunction` is not given, we read the 
weight from `metadata`

```Java
	private final Function<ServiceInstance, Integer> weightFunction;

	public WeightedServiceInstanceListSupplier(ServiceInstanceListSupplier delegate) {
		this(delegate, METADATA_WEIGHT_FUNCTION);
	}

	public WeightedServiceInstanceListSupplier(ServiceInstanceListSupplier delegate, Function<ServiceInstance, Integer> weightFunction) {
		super(delegate);
		this.weightFunction = weightFunction;
	}
```

After that, weight related is as same as `HintBased`, `ZonePreference` and `SameInstancePreference`, we provide two method in builder `withWeighted()` and `withWeighted(Function<ServiceInstance, Integer> weightFunction)` to decide how to map the weight of instance.

jizhuozhi:
> Function<ServiceInstance, Integer> weightFunction

To avoid unnecessary boxing and unboxing, it should be `IntFunction<ServiceInstance> weightFunction`

jizhuozhi:
Hello, @OlgaMaciaszek , I submitted some changes, PTAL.

1. Split weight related methods to `WeightedServiceInstance`.
2. Remove `AtomicInteger getCurrentWeight()` method, using [`WeightedLoadBalancer::calculateCurrentWeight`](https://github.com/spring-cloud/spring-cloud-commons/pull/1111/files#diff-10d2beb0bc9a6e3eb81e639e73d5c3583c58421faefa6058a1a2a033c87a64f9R142) instead.
3. Allow user setting `ToIntFunction<ServiceInstance> weightFunction` to calculate the weight of instance. If not given, we will check whether instance of `WeightedServiceInstance`. At last, we will try to calculate weight from `metadata`, if not set in `metadata`, it will be `DEFAULT_WEIGHT`.

OlgaMaciaszek:
Hi @jizhuozhi - thanks a lot. I'm away for a few days - will get back to it next week.

jizhuozhi:
> Hi @jizhuozhi - thanks a lot. I'm away for a few days - will get back to it next week.

Hello, @OlgaMaciaszek , no problem, I will continue to think if there is a better solution

jizhuozhi:
Hello @OlgaMaciaszek , I've made two main changes since the last discussion.
1. I tried removing the new interface and just using the `weightFunction` to calculate the weight of the instance, this helps keep it as simple as possible
2. I limit the maximum number of samples each time choose is executed to avoid slowing down the execution of the program due to too many instances. 

Here is the optimized microbenchmark, we can see that for very many instances, the performance can remain stable, and the unit test can ensure that the distribution accuracy can reach 99.9% in the case of concurrency.
```
Benchmark                                                                  (hostCount)   Mode  Cnt       Score       Error  Units
WeightedLoadBalancerBenchmarkTests.weightedLoadBalancerChoose                        1  thrpt    5   55385.974 ± 14624.403  ops/s
WeightedLoadBalancerBenchmarkTests.weightedLoadBalancerChoose                       10  thrpt    5   54463.344 ± 12883.199  ops/s
WeightedLoadBalancerBenchmarkTests.weightedLoadBalancerChoose                      100  thrpt    5   42551.335 ±  9318.646  ops/s
WeightedLoadBalancerBenchmarkTests.weightedLoadBalancerChoose                     1000  thrpt    5   45279.520 ±  4289.937  ops/s
WeightedLoadBalancerBenchmarkTests.weightedLoadBalancerChoose                    10000  thrpt    5   43439.806 ±  9629.446  ops/s
WeightedLoadBalancerBenchmarkTests.weightedLoadBalancerChooseConcurrently            1  thrpt    5  193488.917 ± 17728.115  ops/s
WeightedLoadBalancerBenchmarkTests.weightedLoadBalancerChooseConcurrently           10  thrpt    5  187978.163 ± 43241.231  ops/s
WeightedLoadBalancerBenchmarkTests.weightedLoadBalancerChooseConcurrently          100  thrpt    5  172112.049 ± 51418.301  ops/s
WeightedLoadBalancerBenchmarkTests.weightedLoadBalancerChooseConcurrently         1000  thrpt    5  157070.282 ± 51588.981  ops/s
WeightedLoadBalancerBenchmarkTests.weightedLoadBalancerChooseConcurrently        10000  thrpt    5  160940.176 ± 68303.753  ops/s
```

OlgaMaciaszek:
Thanks, @jizhuozhi - will review next week.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
OlgaMaciaszek: Please provide your response on this topic either here or in the issue so that we can move forward with this.

OlgaMaciaszek: Please do not resolve conversations on the PR until either a change is made or an agreement is reached in the discussion in the comments.

OlgaMaciaszek: Sorry, I now see you've added that below.

OlgaMaciaszek:
Hello @jizhuozhi - sorry for not getting back to you earlier. I think using a `WeightedServiceInstanceListSupplier` instead of resolving the weight in the LB is a very good idea; could you redraft your PR with this kind of approach? When it comes to passing the weight in the metadata, I understand the pain point of unnecessary parsing, but I'd rather do it and have all the supported SD systems work with this model than not. I agree that instance weight can be final, but I think in order to serve currently supported SD systems (Netflix Eureka, Consul, Zookeeper -not an SD, but also Kubernetes), I thoughut we'd need to get the initial weight value from the metadata. I'm happy to consider a different proposal, but please make sure it would work with all the supported SDs without the need of introducing changes to them.

jizhuozhi:
> Hello @jizhuozhi - sorry for not getting back to you earlier.

Hello, @OlgaMaciaszek, don't mention it. 

Using a `WeightedServiceInstanceListSupplier` is my initial idea, but when I implemented it I moved it to the `WeightedLoadBalancer`. Because if I move the weight calculation logic to another class, then the problem will go back to the origin, how can I make `WeightedLoadBalancer` get this value without modifying the interface.

As for unnecessary parsing, after this period of trying, this may not be a pain point at present, because through micro-benchmarks, we could see the performance overhead caused by this may be minimal. If performance becomes the bottleneck, then I prefer to use a custom `weightFunction` with cache, which is reserved for enough room for improvement.

How to calculate the weights no longer seems to be the main problem, so now the problem that needs to be solved is how to store this constantly changing current weight. The current implementation is to use `ConcurrentHashMap`, but this may cause memory leaks, and if weak references are used, it will bring accuracy problems. So it's still debatable.

OlgaMaciaszek:
> how can I make WeightedLoadBalancer get this value without modifying the interface and ensure that users can always complete the configuration with minimal changes.

You could verify if it is a `WeightedInstance` - if not, just use some default weight that could be set by properties.

> The current implementation is to use ConcurrentHashMap, but this may cause memory leaks, and if weak references are used, it will bring accuracy problems.

Maybe you could do a quick analysis and some benchmark testing of the solutions you're considering with their drawbacks and advantages, put a summary and then we discuss it here to pick the best one?

jizhuozhi:
> You could verify if it is a WeightedInstance - if not, just use some default weight that could be set by properties.

Like this?
https://github.com/spring-cloud/spring-cloud-commons/pull/1111/commits/1293960d7752ebc9f37db7836a339c2f99bce383#diff-10d2beb0bc9a6e3eb81e639e73d5c3583c58421faefa6058a1a2a033c87a64f9L132-L140

The reason was described at https://github.com/spring-cloud/spring-cloud-commons/pull/1111#issuecomment-1185483682

OlgaMaciaszek:
> Like this?
Yes, like this, just making sure to include some kind of default for non-weighted instances.

jizhuozhi:
> > Like this?
> > Yes, like this, just making sure to include some kind of default for non-weighted instances.

Thanks, I will revert it. 

The final point of contention is how to store the current weights after each calculation and ensure that no memory leaks occur. Do you have any good idea?

jizhuozhi:
> The final point of contention is how to store the current weights of mixed instances after each calculation and ensure that no memory leaks occur.

Hello, @OlgaMaciaszek , the final solution is to check the type of `ServiceInstance`. 

If `ServiceInstance` is an instance of `WeightedServiceInstance`, then use the `getCurrentWeight` method of the instance directly, otherwise save it in `ConcurrentReferenceHashMap` (with `Collections.synchronizedMap(new) WeakHashMap >())` is equivalent), the accuracy will be sacrificed at this time, but it is guaranteed that there will be no memory leaks.

PTAL, thanks.

OlgaMaciaszek:
Thanks, @jizhuozhi - will take a look this week.

jizhuozhi:
Hello @OlgaMaciaszek, thanks for reviewing. Actually there is a concern here, if it is acceptable I will close this PR and resubmit.

For the solution of expanding the list by weight in the supplier, it is a feasible solution, which runs very fast, and can avoid the swarming problem by shuffling the cards. 

I'm worried about a very large heap memory usage for large weight lists. Although the occupied space can be reduced by the greatest common divisor, because of formula $n + 1 \equiv 1 \pmod n$, once there are two adjacent numbers, the gcd will be invalid.

OlgaMaciaszek:
Yes, @jizhuozhi , please do create one based on a Weighted ServiceInstanceListSupplier. For memory issues, one other thing that could be considered is caching; we use it already for cached load-balancing, providing a default Evictor-based caching and a production-appropriate Caffeine-based caching (see `[CachingServiceInstanceListSupplier](https://github.com/spring-cloud/spring-cloud-commons/blob/main/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/CachingServiceInstanceListSupplier.java)`). Maybe for the WeightedServiceInstanceListSupplier we could use some kind of opt-in only caching mechanism as well, to allow people move from a `ConcurrentHashMap` to a proper cache if they need to in production? We'd have to POC to verify if it makes sense, though.

PR #1120, opened 2022-06-11
Performance improvement: refer to https://github.com/spring-cloud/spr…

superbible:
Performance improvement: refer to https://github.com/spring-cloud/spring-cloud-release/issues/244, the cause of the performance regression is: every time when LoadBalancer is called, it query service instance from service register center. The solution is, when LoadBalancer query service instance list, cache the result and refresh it every 30 seconds by default. Now the performance equals to Netflix Ribbon.

### DIFF
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/DiscoveryClientServiceInstanceListSupplier.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/DiscoveryClientServiceInstanceListSupplier.java
@@ -19,6 +19,9 @@ package org.springframework.cloud.loadbalancer.core;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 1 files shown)


marcingrzejszczak:
Please sign the cla and write missing tests

spencergibb:
Will this not already do that https://github.com/spring-cloud/spring-cloud-commons/tree/main/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/cache

superbible:
When use Zookeeper as service register center, the performance regression happens. But not when use Nacos.

There're cache system for Load Balancer already, but it's doesn't work by default, but give a warning: "Spring Cloud LoadBalancer is currently working with the default cache. While this cache implementation is useful for development and tests, it's recommended to use Caffeine cache in production.You can switch to using Caffeine cache, by adding it and org.springframework.cache.caffeine.CaffeineCacheManager to the classpath."
This is the bug. By Default, the Default Cache doen't work. As the suggestion, not matter use Caffeine or Evictor (default), explicit dependecies are required in pom.xml.

So I think probably we can just add Caffeine dependecies by default, or just change the warning to "Spring Cloud LoadBalancer is currently working without cache...", the latter is used in pull request.

PR #1135, opened 2022-08-11
Support LoadBalancer child context creation during AOT processing

OlgaMaciaszek:


spring-cloud-context/src/main/java/org/springframework/cloud/context/named/NamedContextFactory.java
```diff
@@ -126,33 +161,20 @@ protected AnnotationConfigApplicationContext createContext(String name) {
...
+			context.setClassLoader(parent.getClassLoader());
 		}
 		else {
-			context = new AnnotationConfigApplicationContext();
-		}
-		if (this.configurations.containsKey(name)) {
-			for (Class<?> configuration : this.configurations.get(name).getConfiguration()) {
-				context.register(configuration);
-			}
+			context = new GenericApplicationContext();
```
spencergibb, line 169: Should this get the AotDetector treatment as well?

spring-cloud-loadbalancer/pom.xml
```diff
@@ -128,6 +128,11 @@
 			<artifactId>reactor-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-core-test</artifactId>
```
spencergibb, line 133: Interesting. Doesn't spring-boot-starter-test bring this in?

spencergibb (approved):
LGTM, a couple of small questions.

spring-cloud-context/src/main/java/org/springframework/cloud/context/named/NamedContextFactory.java
OlgaMaciaszek: Good catch. I think for child contexts we always have the parent, but maybe there's some edge-case where we don't... I guess better to have it there as well.

Issue #1144, opened 2022-09-05
request-based-sticky-session interferes with round robin balancer resulting in uneven load distribution


icguy:
Take this scenario for instance: I use round robin balancing and [withRequestBasedStickySession](https://github.com/spring-cloud/spring-cloud-commons/blob/68d43a177ed8f2367674fff58b8aedc2e05591ce/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/RequestBasedStickySessionServiceInstanceListSupplier.java#L40) with just two service instances. In the following scenario all requests will be sent to the same instance:

1. request: no instance preference cookie, gets sent to first instance
2. request: cookie is set to prefer the first instance, gets sent to first instance
3. request: no instance preference cookie, gets sent to first instance again.

The third request should be sent to the second service instance to allow for more evenly distributed load.

The round robin balancer [uses an internal counter](https://github.com/spring-cloud/spring-cloud-commons/blob/66b295529e138a528bbc6db36b8da25f4b94bf3b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/RoundRobinLoadBalancer.java#L105-L107) to keep track of which service to choose, it will increment the counter after each call and take the modulo to choose the instance. The problem is that even though the second request explicitly required to be forwarded to the specified instance the counter still gets incremented.

jizhuozhi:
Hello, @icguy and @OlgaMaciaszek . I think there is no need to move position in `RoundRobinLoadBalancer` when there is only 1 instance, I will submit a Pull Request to fix this issue.

PR #1152, opened 2022-10-05
Do not move position when there is only 1 instance

jizhuozhi:
Fixes [gh-1144](https://github.com/spring-cloud/spring-cloud-commons/issues/1144)
Do not move position when there is only 1 instance, especially some suppliers have already filtered instances.

### DIFF
--- a/spring-cloud-loadbalancer/src/test/java/org/springframework/cloud/loadbalancer/core/RoundRobinLoadBalancerTests.java
+++ b/spring-cloud-loadbalancer/src/test/java/org/springframework/cloud/loadbalancer/core/RoundRobinLoadBalancerTests.java
@@ -59,7 +59,7 @@ class RoundRobinLoadBalancerTests {
 		ServiceInstanceListSupplier supplier = mock(ServiceInstanceListSupplier.class);
 		when(supplier.get(any())).thenReturn(Flux.just(Collections.singletonList(new DefaultServiceInstance())));
 		RoundRobinLoadBalancer loadBalancer = new RoundRobinLoadBalancer(new SimpleObjectProvider<>(supplier),
-				"shouldStartFromZeroWhenPositiveOverflow", 0);
+				"shouldNotMovePositionIfOnlyOneInstance", 0);
 		loadBalancer.choose().block();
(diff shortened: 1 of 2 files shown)


jizhuozhi:
Hello, @OlgaMaciaszek , PLAT.

spring-cloud-loadbalancer/src/test/java/org/springframework/cloud/loadbalancer/core/RoundRobinLoadBalancerTests.java
```diff
@@ -52,6 +54,18 @@ void shouldOrderEnforcedWhenPositiveOverflow() {
 		assertOrderEnforced(MAX_VALUE);
 	}
 
+	@Test
+	void shouldNotMovePositionIfOnlyOneInstance() {
+		ServiceInstanceListSupplier supplier = mock(ServiceInstanceListSupplier.class);
+		when(supplier.get(any())).thenReturn(Flux.just(Collections.singletonList(new DefaultServiceInstance())));
+		RoundRobinLoadBalancer loadBalancer = new RoundRobinLoadBalancer(new SimpleObjectProvider<>(supplier),
+				"shouldNotMovePositionIfOnlyOneInstance", 0);
```
OlgaMaciaszek, line 62: Please add an empty line after this to visually divide the test setup section.

OlgaMaciaszek (changes requested):
Thansk for the PR @jizhuozhi. Makes sense. I've just added a comment about a cosmetic change in test - please address. Also, in order for the fix to also make it to `2021.x` release train, could you please submit this against `3.1.x` instead of `main`?

jizhuozhi:
> Thansk for the PR @jizhuozhi. Makes sense. I've just added a comment about a cosmetic change in test - please address. Also, in order for the fix to also make it to `2021.x` release train, could you please submit this against `3.1.x` instead of `main`?

Aha, this PR is to `3.1.x` but not `main`

OlgaMaciaszek (approved):
Thanks for the quick fix, @jizhuozhi. LGTM.

PR #1160, opened 2022-10-25
Destroy and initialize beans within the correct context

### LINKED ISSUE #1158: ConfigurationPropertiesRebinder is calling destroyBean on beans that it does not own
Originally seen in https://github.com/spring-projects/spring-boot/issues/30395, but now also in a user project https://github.com/spring-projects/spring-boot/issues/32850.

Analysis by @mbhave:

The `ConfigurationPropertiesRebinder` calls `destroyBean` and `initializeBean` on beans that it does not own, ie those bean definitions are in the parent context. When Boot’s `ConfigurationPropertiesBindingPostProcessor` is called it does not find the bean definition (because it’s not in the current context). It seems that Spring Cloud during rebind should call destroy and `initializeBean` on the application context which contains the bean definition, using the `containsLocalBean` check to go up the hierarchy. Applying post processors on a bean that the context does not own, while nothing stops you from doing that, is not expected.

ryanjbaxter:
Fixes #1158

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/context/properties/ConfigurationPropertiesRebinder.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/context/properties/ConfigurationPropertiesRebinder.java
@@ -89,30 +89,37 @@ public class ConfigurationPropertiesRebinder
 		if (!this.beans.getBeanNames().contains(name)) {
 			return false;
 		}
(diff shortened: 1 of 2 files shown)


spencergibb (approved):
Does this need to go in 3.1.x?

ryanjbaxter:
@spencergibb I cherry picked it to 3.1.x as well

PR #1165, opened 2022-10-30
Lazy fills service instances to reduce max latency

jizhuozhi:
When we have a lot of upstream services, and their total weight is large, then if we expand all instances in one request, our processing time will be very long. To solve this problem, use a lazily computed list to expand all elements to an unexpanded position only when the unexpanded position is visited.

There is an extreme case here. When using `RandomLoadBalancer`, it is possible to select a very late node for the first time, so the delay will still be large. But when we use `RoundRobinLoadBalancer`, we only move the position of one element at a time, so the effect is negligible.

### DIFF
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/LazyWeightedServiceInstanceList.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/LazyWeightedServiceInstanceList.java
@@ -22,7 +22,8 @@ import java.util.List;
 import org.springframework.cloud.client.ServiceInstance;
 
 /**
- * A {@link List} implementation that weighted and lazy fills {@link ServiceInstance}.
+ * A {@link List} implementation that lazily fills weighted {@link ServiceInstance}
+ * objects.
  *
  * @author Zhuozhi Ji
  * @see WeightedServiceInstanceListSupplier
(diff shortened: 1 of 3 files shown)


jizhuozhi:
Hello, @OlgaMaciaszek . I create a PR that lazy fills service instances to reduce max latency, PLAT.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/LazyWeightedServiceInstanceList.java
```diff
@@ -0,0 +1,136 @@
...
+	LazyWeightedServiceInstanceList(List<ServiceInstance> instances, int[] weights) {
+		this.instances = instances;
+		this.weights = weights;
+		this.init();
+	}
+
+	private void init() {
+		// Calculate the greatest common divisor (GCD) of weights, max weight and the
+		// total number of elements after expansion.
+		int gcd = 0;
```
OlgaMaciaszek, line 53: Please rename variable to `greatestCommonDivisor`.

OlgaMaciaszek (changes requested):
Thanks @jizhuozhi. Looks like a nice performance improving refactoring of the previous PR. The logic looks good, however I've added some comment regarding the code style and naming - please address.

OlgaMaciaszek (approved):
LGTM.

PR #1167, opened 2022-11-03
Add note about removing configuration values and performing refreshes

### LINKED ISSUE #1161: ConfigurationPropertiesRebinder#rebind()  works in wrong result by deleting the cofniguration
the code like this:
```
@Data
@ToString
@ConfigurationProperties(value = MonitorPluginConfiguration.PREFIX)
public class MonitorPluginConfiguration {
    public static final String PREFIX = "monitor";
    private Map<String, String> plugins;
}
```

the yaml configuration in nacos:

```
monitor:
  plugins:
    p1: 1
    p2: 2
    p3: 3
    p4: 4
```
if i delete the key-value  `p4: 4` in the configuration from nacos ，the value for field MonitorPluginConfiguration#plugins will not be deleted.

the changed configuration in nacos after delete the  the key-value  `p4: 4`
```
monitor:
  plugins:
    p1: 1
    p2: 2
    p3: 3
```

I trace the code , i find the method in **`org.springframework.boot.context.properties.bind.MapBinder#merge`**，
```
        @Override
	protected Map<Object, Object> merge(Supplier<Map<Object, Object>> existing, Map<Object, Object> additional) {
		Map<Object, Object> existingMap = getExistingIfPossible(existing);
		if (existingMap == null) {
			return additional;
		}
		try {
			existingMap.putAll(additional);
			return copyIfPossible(existingMap);
		}
		catch (UnsupportedOperationException ex) {
			Map<Object, Object> result = createNewMap(additional.getClass(), existingMap);
			result.putAll(additional);
			return result;
		}
	}
```
the parmater `additional `value is `{p1=1, p2=2, p3=3}`
image

and the variable `existingMap` value is` {p1=1, p2=2, p3=3, p4=4}`
image

for the code `existingMap.putAll(additional)`, it will save the original value ` {p1=1, p2=2, p3=3, p4=4}`.
so the result is bad.
I think that it is good to execute `existingMap.clear()` before  executing the code `existingMap.putAll(additional)`
The code like this:

```
        @Override
	protected Map<Object, Object> merge(Supplier<Map<Object, Object>> existing, Map<Object, Object> additional) {
		Map<Object, Object> existingMap = getExistingIfPossible(existing);
		if (existingMap == null) {
			return additional;
		}
		try {
			existingMap.clear(); // add this line
			existingMap.putAll(additional);
			return copyIfPossible(existingMap);
		}
		catch (UnsupportedOperationException ex) {
			Map<Object, Object> result = createNewMap(additional.getClass(), existingMap);
			result.putAll(additional);
			return result;
		}
	}
```

this issue I have raise at https://github.com/spring-projects/spring-boot/issues/32877, but I got the reply like this

image

ryanjbaxter:
Fixes #1161

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-commons.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-commons.adoc
@@ -214,6 +214,11 @@ Specifically, anything that depends on those beans cannot rely on them being upd
 In that case, it is rebuilt on a refresh and its dependencies are re-injected.
 At that point, they are re-initialized from the refreshed `@Configuration`).
 
+NOTE:  Removing a configuration value and then performing a refresh will not update the presence of the configuration value.
+The configuration property must be present in order to update the value after a refresh.  If you are relying on the presence of
+a value in your application you might want to switch your logic to rely on its absence instead.  Another option would be to rely
+on the value changing rather than not being present in the application's configuration.
+
 === Encryption and Decryption
 
 Spring Cloud has an `Environment` pre-processor for decrypting property values locally.


docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -214,6 +214,10 @@ Specifically, anything that depends on those beans cannot rely on them being upd
 In that case, it is rebuilt on a refresh and its dependencies are re-injected.
 At that point, they are re-initialized from the refreshed `@Configuration`).
 
+NOTE:  Removing a configuration value and then performing a refresh will not update the presence of the configuration value.
+The configuration property must be present in order to update the value after a refresh.  If you are relying on the presence of
+a value in your application you might want to switch your logic to rely on its absence instead.
```
spencergibb, line 219: > to rely on its absence instead.

Or a value that changes rather than depending on presence/absence.

ryanjbaxter: Updated

spencergibb (approved):

Issue #823, opened 2020-09-14
SimpleServcieInstance should override getScheme()


dagerber:
I'm submitting a bugreport for

- **spring-cloud-commons**
- class **org.springframework.cloud.client.discovery.simple.SimpleDiscoveryProperties**

Versions used:

- spring-boot: 2.3.3
- spring-cloud: Hoxton.SR7
- spring-cloud-commons: 2.2.4.RELEASE

When using SimpleDiscoveryClient, non-HTTPS Urls do not work, they are always resolved as https:// URLs.
We are using https in production, but http in Tests to avoid the need to setup certificates for localhost.
There are workarounds, but I think this is a bug that should be fixed in spring-cloud-commons

e.g. 

"spring.cloud.discovery.client.simple.instances.foo-service[0].uri= http://localhost:8889",
"spring.cloud.discovery.client.simple.instances.foo-service[0].secure = false",

foo-service is always resolved as **https**://localhost:8889, because SimpleServcieInstance (inner class of SimpleServiceProperties) does not override getScheme().

LoadBalancerUriTools calls getScheme() and in case of SimpleServiceInstance always receives null and thus always sets Scheme to https.

Proposed fix for org.springframework.cloud.client.discovery.simple.SimpleDiscoveryProperties#SimpleServiceInstance

`
public String getScheme() {
   return this.isSecure() ? "https" : "http";
 }
`

spencergibb:
This should be done in `DefaultServiceInstance` as well.

dagerber:
Has been fixed with https://github.com/spring-cloud/spring-cloud-commons/commit/72762d76ed9eccc5876e5bf099c2495c4c44ff3e

dagerber:
Sorry, thought this was solved, but it is not.

dagerber:
https://github.com/spring-cloud/spring-cloud-commons/pull/926

spencergibb:
Closed via #926

spencergibb:
Reverted, at least commons, gateway and contract have behavior that depends on the null implementation that will need to be addressed in the future.

OlgaMaciaszek:
Requires more discussion on impact in GW. Reverting the change and reopening.

PR #1168, opened 2022-11-08
Allow overriding getScheme on DefaultServiceInstance.

OlgaMaciaszek:
Fixes gh-823.

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors.
+ * Copyright 2012-2022 the original author or authors.
  *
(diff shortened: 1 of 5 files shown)


spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
```diff
@@ -148,8 +180,8 @@ public void setUri(URI uri) {
 		this.uri = uri;
 		this.host = this.uri.getHost();
 		this.port = this.uri.getPort();
-		String scheme = this.uri.getScheme();
-		if ("https".equals(scheme)) {
+		scheme = this.uri.getScheme();
+		if (Arrays.asList("https", "wss").contains(scheme)) {
```
ryanjbaxter, line 184: Any concern on case?
I would make the list a class variable instead of creating it everytime setUri is called.

ryanjbaxter (commented):
LGTM

PR #1170, opened 2022-11-15
Initial implementation of httpclient 5

ryanjbaxter:


spring-cloud-commons/src/main/java/org/springframework/cloud/commons/httpclient/HttpClientConfiguration.java
```diff
@@ -33,8 +33,34 @@
 public class HttpClientConfiguration {
 
 	@Configuration(proxyBeanMethods = false)
-	@ConditionalOnProperty(name = "spring.cloud.httpclientfactories.apache.enabled", matchIfMissing = true)
+	@ConditionalOnProperty(name = "spring.cloud.httpclientfactories.apache5.enabled", matchIfMissing = true)
```
spencergibb, line 36: I probably leave `apache`

spencergibb: Oh, you've got both?

ryanjbaxter: Right implementations for both, depending on whats on the classpath

ryanjbaxter: They would have to put apache4 on the classpath AND enable to property enabled

PR #1174, opened 2022-11-23
Add flexibility to zone preference in load balancing

BenoitKessler:
Enables to easily create custom load balancer configurations with zone preference, but a different zone then the one defined by spring.cloud.loadbalancer.zone

### DIFF
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/ServiceInstanceListSupplierBuilder.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/ServiceInstanceListSupplierBuilder.java
@@ -225,6 +225,21 @@ public final class ServiceInstanceListSupplierBuilder {
 	}
 
 	/**
+	 * Adds a {@link ZonePreferenceServiceInstanceListSupplier} to the
+	 * {@link ServiceInstanceListSupplier} hierarchy.
+	 * @param zoneName desired zone for zone preference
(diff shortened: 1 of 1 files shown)


BenoitKessler:
Currently, if I want to implement a custom load balancer configuration with zone preference, I could do : 
`public class CustomLoadBalancerConfiguration {`
`    @Bean`
`    public ServiceInstanceListSupplier customDiscoveryClientServiceInstanceListSupplier(final ConfigurableApplicationContext context) {`
`       return ServiceInstanceListSupplier.builder()`
`           .withBlockingDiscoveryClient()`
`           .withZonePreference()`
`           .build(context);`
`    }`
`}`
And in my configuration add
`@LoadBalancerClient(value = "authProvider", configuration = CustomLoadBalancerConfiguration.class)`
But this would use the zone defined by `spring.cloud.loadbalancer.zone` which is not helping me.

With this change, I want to be able to define a specific zone preference for each custom load balancer configuration : 
`public class CustomLoadBalancerConfiguration {`
`    @Bean`
`    public ServiceInstanceListSupplier customDiscoveryClientServiceInstanceListSupplier(final ConfigurableApplicationContext context) {`
`        String authProviderZone = context.getEnvironment().getProperty("auth.provider.loadbalancer.zone");`
`        return ServiceInstanceListSupplier.builder()`
`            .withBlockingDiscoveryClient()`
`            .withZonePreference(authProviderZone)`
`            .build(context);`
`    }`
`}`

OlgaMaciaszek (approved):
Thanks @BenoitKessler . LGTM.

PR #1188, opened 2023-01-08
Allow empty keystore in SSLContextFactory. Fixes #1183

### LINKED ISSUE #1183: SSLContextFactory should allow for no key store
I am using 3.1.4. My spring cloud config server has a self signed cert. I did not want to add its certificate to java trust store, so I wanted to use TlsProperties to provide the trust store. I can't do that without providing the keystore.

Currently the trust store is optional, I think the keystore needs to be optional too, as not always we need to use ssl client verification.

I can made a PR with the change, but I am not sure if removing this requirement in a straightforward way is OK for you.

freafrea:
Fixes https://github.com/spring-cloud/spring-cloud-commons/issues/1183


spencergibb:
Can you add a test?

freafrea:
Sure, although this class did not have any test that I could find so it might take me a while to understand how to write it in order to be consistent with other tests.

PR #1199, opened 2023-01-19
Refactor code to simplify calculation logic

jizhuozhi:
The implemented logic is the same as #1184, both are O(1) schedulers, but the code implementation is simplified through double queues, and recursive processing that may cause errors is avoided

### DIFF
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/LazyWeightedServiceInstanceList.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/LazyWeightedServiceInstanceList.java
@@ -36,7 +36,7 @@ class LazyWeightedServiceInstanceList extends AbstractList<ServiceInstance> {
 
 	private final Object expandingLock = new Object();
 
-	private final O1ServiceInstanceSelector selector;
+	private O1ServiceInstanceSelector selector;
 
 	private volatile int position = 0;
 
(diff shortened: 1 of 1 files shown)


spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/LazyWeightedServiceInstanceList.java
```diff
@@ -82,77 +83,61 @@ static int greatestCommonDivisor(int a, int b) {
...
-		static final int MODE_QUEUE = 1;
-
-		final ServiceInstance[] instances;
-
-		final int[] weights;
-
-		final int greatestCommonDivisor;
-
-		final Queue<Entry> queue;
+	static class O1ServiceInstanceSelector {
```
OlgaMaciaszek, line 86: Maybe rename to sth like `WeightedInstanceSelector` as this puts focus on the main feature.

OlgaMaciaszek (commented):
Thanks for the PR, @jizhuozhi - this in fact is simpler and more readable. Have added just one cosmetic comment - please address it.

OlgaMaciaszek (changes requested):
Thanks for the PR, @jizhuozhi - this in fact is simpler and more readable. Have added just one cosmetic comment - please address it.

jizhuozhi:
Hello, @OlgaMaciaszek . All conversation has resolved, PLAT :)

OlgaMaciaszek (approved):
Thanks, @jizhuozhi. LGTM.

PR #1228, opened 2023-04-21
Add application listener to locate property sources during bootstrap

ryanjbaxter:
Also adds support for activating profiles using spring.profiles.active from bootstrap property source listeners.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-commons.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-commons.adoc
@@ -54,6 +54,10 @@ The additional property sources are:
 An example would be properties from the Spring Cloud Config Server.
 See "`<<customizing-bootstrap-property-sources>>`" for how to customize the contents of this property source.
 
+NOTE:  Prior to Spring Cloud 2021.0.7 `PropertySourceLocators` (including the ones for Spring Cloud Config) were run during
+the main application context and not in the Bootstrap context.  You can force `PropertySourceLocators` to be run during the
+Bootstrap context by setting `spring.cloud.config.initialize-on-context-refresh=true` in `bootstrap.[properties | yaml]`.
+
 * "`applicationConfig: [classpath:bootstrap.yml]`" (and related files if Spring profiles are active): If you have a `bootstrap.yml` (or `.properties`), those properties are used to configure the bootstrap context.
 Then they get added to the child context when its parent is set.
 They have lower precedence than the `application.yml` (or `.properties`) and any other property sources that are added to the child as a normal part of the process of creating a Spring Boot application.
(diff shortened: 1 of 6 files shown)


ryanjbaxter:
Partial fix for https://github.com/spring-cloud/spring-cloud-config/issues/1922

ryanjbaxter:
Profiles from the bootstrap context get applied to the child context here
https://github.com/spring-cloud/spring-cloud-commons/blob/main/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/BootstrapApplicationListener.java#L372
When `ParentContextApplicationContextInitializer.initialize` sets the parent context to the bootstrap context it inherits the active profiles.  

This all happens after the environment is prepared when config data runs.  
Here is where we prepare the environment 
https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java#L306

And here is where the parent context is set
https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java#L310

docs/src/main/asciidoc/spring-cloud-commons.adoc
```diff
@@ -54,6 +54,10 @@ The additional property sources are:
 An example would be properties from the Spring Cloud Config Server.
 See "`<<customizing-bootstrap-property-sources>>`" for how to customize the contents of this property source.
 
+NOTE:  Prior to Spring Cloud 2021.0.7 `PropertySourceLocators` (including the ones for Spring Cloud Config) were run during
```
spencergibb, line 57: Not sure I like calling out specific version in the docs. Maybe, "previously"?

spencergibb (approved):

docs/src/main/asciidoc/spring-cloud-commons.adoc
ryanjbaxter: But after the next release then that statement wont be correct

spencergibb: @Buzzardo what do you suggest? Do you know what other teams do in these situations?

spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/BootstrapApplicationListener.java
```diff
@@ -294,6 +294,13 @@ private void apply(ConfigurableApplicationContext context, SpringApplication app
 		target.addAll(getOrderedBeansOfType(context, ApplicationContextInitializer.class));
 		application.setInitializers(target);
 		addBootstrapDecryptInitializer(application);
+
+		// Get the active profiles from the bootstrap context and set them in main
+		// application
+		// environment. This allows any profiles activated during bootstrap to be
+		// activated when
+		// config data runs in the main application context.
+		environment.setActiveProfiles(context.getEnvironment().getActiveProfiles());
```
qnnn, line 303: @ryanjbaxter  This may not fully resolve all issues mentioned in the original issue. As noted, when the spring.profiles.active configuration is sourced from a config server, it does not take effect in the main application context. The root cause, which needs fixing here, is similar: setting spring.profiles.active before the main application context is ready. However, there are other configurations besides spring.profiles.active that also need to be set in advance, such as those starting with spring.main.
I lean towards utilizing BootstrapApplicationListener to proactively locate all remote configurations for the main application context, similar to what ConfigDataLoader does.

PR #1236, opened 2023-05-17
replace poor punctuation and naming

GierSmith:
The name of the variable is o_O does not conform to Java's naming rules. 
"..." uses the wrong punctuation


OlgaMaciaszek (changes requested):
Hello @GierSmith, thanks for submitting a PR. `...` has correct punctuation - you do not put spaces in between. Please revert that change. I agree the exception name could be improved, but we try to avoid non-explicit, single-letter variable names in favour of more meaningful naming. Please rename `e` to `exception`. Also, please remember to update copyright dates (above) to `-2023` in any changed java files.

GierSmith:
The `...` in the previous version was not an English character, my commit changed it to an English character, I think this is the correct change.

OlgaMaciaszek:
@Buzzardo - could you please take a look?

Buzzardo (approved):

PR #1255, opened 2023-07-18
Make EncryptionBootstrapConfiguration.EncryptionHints static

izeye:
This PR changes the `EncryptionBootstrapConfiguration.EncryptionHints` to `static`.


spencergibb:
Can I ask why you want it static. Not opposed to the change, just curious.

izeye:
@spencergibb Ah, sorry for missing context. I thought that if an inner class doesn't have any references to its outer class, making it `static` is preferred in general.

izeye:
I'm not sure as I don't know how the `aot.factories` file works internally, but it seems to require its outer class' instantiation for `RuntimeHintsRegistrar`.

See https://github.com/spring-cloud/spring-cloud-commons/blob/main/spring-cloud-context/src/main/resources/META-INF/spring/aot.factories

OlgaMaciaszek (changes requested):
Thanks, @izeye; if we want to do any changes regarding hint class structure, I'd rather move them to package-level classes in separate file as that's the convention Spring framework has moved to for hints classes.

izeye:
@OlgaMaciaszek Thanks for the feedback! I updated as you suggested.

OlgaMaciaszek (approved):
LGTM

PR #1286, opened 2023-09-29
Fix #1285 Spring Cloud activates unexpected profiles when some conditions met

### LINKED ISSUE #1285: Spring Cloud activates unexpected profiles when some conditions met
According to [3.1. Adding Active Profiles](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.profiles.adding-active-profiles):

> "The ***spring.profiles.active*** property follows the same ordering rules as other properties: The highest ***PropertySource*** wins. This means that you can specify active profiles in ***application.properties*** and then ***replace*** them by using the command line switch."

Hence, I would expect ***spring.profiles.active*** can only introduce profiles in property source with highest priority, while ***spring.profiles.include*** can accept any profiles from all valid property sources.

3.1.6 is the last version that holds above rules. Since 3.1.7, profiles under *spring.profiles.active* from all property sources are merged, which does not follow the existing behaviours.

I have quickly written a demo and detailed descriptions for this issue [here](https://github.com/leovx/spring-cloud-profile-bug-demo).

## Trigger Conditions:
0. Using Spring Cloud 2021.0.8 (Spring Cloud Common 3.1.7) onwards
1. Using Spring Boot 2.7.x onwards
2. In application.yml, we give (for out-of-box experience, run locally for dev and test once after clone)
> spring.profiles.active: local
3. In command line or any other remote property source which has higher priority, e.g. Kubernetes ConfigMap, we give
> spring.profiles.active: prod, secure
> #intended to override and replace value in application.yml
4. Provide a bean that implements *org.springframework.cloud.bootstrap.config.PropertySourceLocator*

## Buggy Behaviour
Instead of accepting "*prod*" and "*secure*" profiles only, it accepts *local* profile as well.
> INFO 5445 --- [           main] com.jinghao.demo.DemoApplication         : The following 3 profiles are active: "prod", "secure", "local"

## Potential Hazard
In this demostration project, 2 beans of AuthManager would be candidates to use - LocalBypassAuthManager should only be used in local development environment due to some reasons (say something cannot be installed locally but should not impact your BAU functional implementation), while NormalAuthManager should be applied in any valid remote environment - especially Production. In this case, both "prod" and "local" profiles are activated and hence the illegal bean of LocalBypassAuthManager will be used in Production making any authentication tokens bypass and become valid.

leovx:
https://github.com/spring-cloud/spring-cloud-commons/issues/1285


yanhom1314:
I recently upgraded to version 3.1.4 and encountered an issue with retrieving active profiles from different PropertySources.

leovx:
> I recently upgraded to version 3.1.4 and encountered an issue with retrieving profiles from different PropertySources. I want to know if this is considered a bug, as previous versions did not have this problem.

Not quite aware of your problem, coz I only found this issue comparing 3.1.7 and 3.1.6. You can check if your issue persists in the latest version and possibly provide a POC.

ryanjbaxter (changes requested):
You will need to change the PR to be against the 3.1.x branch and also fix the checkstyle errors.

leovx:
> You will need to change the PR to be against the 3.1.x branch and also fix the checkstyle errors.

Code style issues fixed and rebased to 3.1.x.

Issue #1288, opened 2023-10-10
deterministic subsetting algorithm to limit the number of instances


jizhuozhi:
**Is your feature request related to a problem? Please describe.**
Many RPCs use connection pools and long connections to reduce TCP handshake overhead (including RPC over HTTP). In a very large-scale cluster, there may be tens of thousands of instances (such as a search engine). If the client establishes connections with all instances, all available TCP ports will be quickly exhausted.

**Describe the solution you'd like**
In order to solve this problem, in very large-scale clusters, each client should only select a limited subset to establish connections. Google SRE provides a deterministic subsetting algorithm solution, see https://sre.google/sre-book/load-balancing-datacenter

PR #1289, opened 2023-10-10
feat(lb): deterministic subsetting algorithm

jizhuozhi:
[gh-1288](https://github.com/spring-cloud/spring-cloud-commons/issues/1288)

### DIFF
--- a/docs/modules/ROOT/pages/spring-cloud-commons/loadbalancer.adoc
+++ b/docs/modules/ROOT/pages/spring-cloud-commons/loadbalancer.adoc
@@ -403,9 +403,9 @@ public class CustomLoadBalancerConfiguration {
 	}
 ----
 
-TIP: Each service instance will be assigned a unique `instanceId` by default, and different `instanceId` will often select different subsets. Normally you don't need to pay attention to it, but if you need to have multiple instances select the same subset, you can set it via `spring.cloud.loadbalancer.subset.instance-id` (supports placeholders).
+TIP: By default, each service instance is assigned a unique `instanceId`, and different `instanceId` values often select different subsets. Normally, you need not pay attention to it. However, if you need to have multiple instances select the same subset, you can set it with `spring.cloud.loadbalancer.subset.instance-id` (which supports placeholders).
 
-TIP: The size of the subset is set to 100 by default, you can also set it via `spring.cloud.loadbalancer.subset.size`
+TIP: By default, the size of the subset is set to 100. You can also set it with `spring.cloud.loadbalancer.subset.size`.
 
 [[spring-cloud-loadbalancer-starter]]
 == Spring Cloud LoadBalancer Starter
--- a/docs/modules/ROOT/pages/spring-cloud-commons/loadbalancer.adoc
+++ b/docs/modules/ROOT/pages/spring-cloud-commons/loadbalancer.adoc
@@ -379,6 +379,34 @@ For `WebClient`, you need to implement and define `LoadBalancerClientRequestTran
 If multiple transformers are defined, they are applied in the order in which Beans are defined.
 Alternatively, you can use `LoadBalancerRequestTransformer.DEFAULT_ORDER` or `LoadBalancerClientRequestTransformer.DEFAULT_ORDER` to specify the order.
 
+[[loadbalancer-subset]]
+== Spring Cloud LoadBalancer Subset
+
+`SubsetServiceInstanceListSupplier` implements a https://sre.google/sre-book/load-balancing-datacenter/[deterministic subsetting algorithm] to select a limited number of instances in the `ServiceInstanceListSupplier` delegates hierarchy.
+
+You can configure it either by setting the value of `spring.cloud.loadbalancer.configurations` to `subset` or by providing your own `ServiceInstanceListSupplier` bean -- for example:
+
+[[subset-custom-loadbalancer-configuration-example]]
+[source,java,indent=0]
+----
+public class CustomLoadBalancerConfiguration {
+
+	@Bean
+	public ServiceInstanceListSupplier discoveryClientServiceInstanceListSupplier(
+			ConfigurableApplicationContext context) {
+		return ServiceInstanceListSupplier.builder()
+					.withDiscoveryClient()
+					.withSubset()
+					.withCaching()
+					.build(context);
+	    }
+	}
+----
+
+TIP: Each service instance will be assigned a unique `instanceId` by default, and different `instanceId` will often select different subsets. Normally you don't need to pay attention to it, but if you need to have multiple instances select the same subset, you can set it via `spring.cloud.loadbalancer.subset.instance-id` (supports placeholders).
+
+TIP: The size of the subset is set to 100 by default, you can also set it via `spring.cloud.loadbalancer.subset.size`
+
 [[spring-cloud-loadbalancer-starter]]
 == Spring Cloud LoadBalancer Starter
 
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/annotation/LoadBalancerClientConfiguration.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/annotation/LoadBalancerClientConfiguration.java
@@ -217,6 +217,15 @@ public class LoadBalancerClientConfiguration {
 					.build(context);
 		}
 
+		@Bean
+		@ConditionalOnBean(DiscoveryClient.class)
+		@ConditionalOnMissingBean
+		@Conditional(SubsetConfigurationCondition.class)
+		public ServiceInstanceListSupplier subsetServiceInstanceListSupplier(ConfigurableApplicationContext context) {
+			return ServiceInstanceListSupplier.builder().withBlockingDiscoveryClient().withSubset().withCaching()
+					.build(context);
+		}
+
 	}
 
 	@Configuration(proxyBeanMethods = false)
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/annotation/LoadBalancerClientConfiguration.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/annotation/LoadBalancerClientConfiguration.java
@@ -143,6 +143,15 @@ public class LoadBalancerClientConfiguration {
 					.build(context);
 		}
 
+		@Bean
+		@ConditionalOnBean(ReactiveDiscoveryClient.class)
+		@ConditionalOnMissingBean
+		@Conditional(SubsetConfigurationCondition.class)
+		public ServiceInstanceListSupplier subsetServiceInstanceListSupplier(ConfigurableApplicationContext context) {
+			return ServiceInstanceListSupplier.builder().withDiscoveryClient().withSubset().withCaching()
+					.build(context);
+		}
+
 	}
 
 	@Configuration(proxyBeanMethods = false)
@@ -349,6 +358,16 @@ public class LoadBalancerClientConfiguration {
 		public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
 			return LoadBalancerEnvironmentPropertyUtils.equalToForClientOrDefault(context.getEnvironment(),
 					"configurations", "weighted");
+		}
+
+	}
+
+	static class SubsetConfigurationCondition implements Condition {
+
+		@Override
+		public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+			return LoadBalancerEnvironmentPropertyUtils.equalToForClientOrDefault(context.getEnvironment(),
+					"configurations", "subset");
 		}
 
 	}
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/ServiceInstanceListSupplierBuilder.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/ServiceInstanceListSupplierBuilder.java
@@ -35,6 +35,7 @@ import org.springframework.cloud.loadbalancer.cache.LoadBalancerCacheManager;
 import org.springframework.cloud.loadbalancer.config.LoadBalancerZoneConfig;
 import org.springframework.cloud.loadbalancer.support.LoadBalancerClientFactory;
 import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.core.env.PropertyResolver;
 import org.springframework.http.HttpStatus;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
@@ -296,6 +297,16 @@ public final class ServiceInstanceListSupplierBuilder {
 		DelegateCreator creator = (context, delegate) -> {
 			LoadBalancerClientFactory factory = context.getBean(LoadBalancerClientFactory.class);
 			return new HintBasedServiceInstanceListSupplier(delegate, factory);
+		};
+		creators.add(creator);
+		return this;
+	}
+
+	public ServiceInstanceListSupplierBuilder withSubset() {
+		DelegateCreator creator = (context, delegate) -> {
+			PropertyResolver resolver = context.getBean(PropertyResolver.class);
+			LoadBalancerClientFactory factory = context.getBean(LoadBalancerClientFactory.class);
+			return new SubsetServiceInstanceListSupplier(delegate, resolver, factory);
 		};
 		creators.add(creator);
 		return this;
--- a/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/SubsetServiceInstanceListSupplier.java
+++ b/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/SubsetServiceInstanceListSupplier.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2012-2023 the original author or authors.
+ *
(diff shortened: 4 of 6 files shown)


jizhuozhi:
Hello @OlgaMaciaszek , PTAL

spencergibb:
Please be patient, it has only been one day since you submitted it.

jizhuozhi:
Sorry, I was impatient

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/SubsetServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,80 @@
...
+ * <a href="https://sre.google/sre-book/load-balancing-datacenter/">deterministic
+ * subsetting</a> to limit the number of instances provided by delegate.
+ *
+ * @author Zhuozhi Ji
+ */
+public class SubsetServiceInstanceListSupplier extends DelegatingServiceInstanceListSupplier {
+
+	static final int DEFAULT_SUBSET_SIZE = 100;
+
+	final String instanceId;
```
OlgaMaciaszek, line 40: The naming of this field does not seem very clear to me; we use the term `instanceId` to signify the id of each of the instances available in the ServiceRegistry, for a given `serviceId`. I see here you use it to generate a `clientId`. I suggest you rename the field to `clientId` and perform any necessary transformations while setting its value.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/SubsetServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,80 @@
...
+
+	final int subsetSize;
```
OlgaMaciaszek, line 42: I think all the fields in this class could be `private`.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/SubsetServiceInstanceListSupplier.java
OlgaMaciaszek: I'm thinking, we should also add a configuration properties  for the `clientId` and `subsetSize` along with their default values, that allow the users to set it and in this class we could just inject the corresponding `@ConfiguationProperties` bean through constructor instead of passing the values itself. We'd then just need one constructor (see `HintBasedServiceInstanceListSupplier` as a reference).

OlgaMaciaszek (changes requested):
Hello @jizhuozhi, thanks for submitting the PR. I have done an initial review after looking at the class structure. I have added some comments - please address. Also, please add documentation. I will take a look at the actual algorithm implementation and tests tomorrow and provide a second review then.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/SubsetServiceInstanceListSupplier.java
jizhuozhi: Got, I will fix them

OlgaMaciaszek: Basically, there are two things. One is the id of the `LoadBalancer` context responsible for the communication with a given external service that encompasses all the `ServiceInstanceListSupplier` and `ReactiveLoadBalancer` instances used to communicate with a given service; so if the app uses load-balancing to communicate with `test-service` and `demo-service`, 2 such named LoadBalancer contexts will be created, however, unless set up otherwise it happens lazily during the first load-balanced call to that service. See [the code](https://github.com/spring-cloud/spring-cloud-commons/blob/14bfa57922f344f92b93aada588a4a7d0ebaf742/spring-cloud-context/src/main/java/org/springframework/cloud/context/named/NamedContextFactory.java#L235-L239). Another thing is how the client application registers itself in a service registry - it's general `clientId`, which to me is what we should use. That is simply taken from Spring's `spring.application.name` property. See the code: https://github.com/spring-cloud/spring-cloud-commons/blob/14bfa57922f344f92b93aada588a4a7d0ebaf742/spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/simple/SimpleDiscoveryClientAutoConfiguration.java#L61-L69 and https://github.com/spring-cloud/spring-cloud-commons/blob/14bfa57922f344f92b93aada588a4a7d0ebaf742/spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/simple/reactive/SimpleReactiveDiscoveryClientAutoConfiguration.java#L59-L60, so to me that's what you should take. Then, for every LoadBalancer context, you may create a `SubsettingServiceInstanceListSupplier` bean with the same `clientId` (taken from `spring.application.name`), but this will not be a problem, since each of these beans would communicate with a different service. Let me know if you have any questions or comments on this.

OlgaMaciaszek: And the property field for the `subsetSize` should go to `LoadBalancerProperties` and be retrieved from there.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/SubsetServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,80 @@
...
+
+import org.springframework.cloud.client.ServiceInstance;
+
+/**
+ * A {@link ServiceInstanceListSupplier} implementation that uses
+ * <a href="https://sre.google/sre-book/load-balancing-datacenter/">deterministic
+ * subsetting</a> to limit the number of instances provided by delegate.
+ *
+ * @author Zhuozhi Ji
+ */
```
OlgaMaciaszek, line 35: Please add `@since 4.1.0` to the Javadoc.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/SubsetServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	public SubsetServiceInstanceListSupplier(ServiceInstanceListSupplier delegate, String instanceId, int subsetSize) {
+		super(delegate);
+		this.instanceId = instanceId;
+		this.subsetSize = subsetSize;
+	}
+
+	@Override
+	public Flux<List<ServiceInstance>> get() {
```
OlgaMaciaszek, line 59: The algorithm implementation looks good.

spring-cloud-loadbalancer/src/test/java/org/springframework/cloud/loadbalancer/core/SubsetServiceInstanceListSupplierTest.java
```diff
@@ -0,0 +1,124 @@
...
+import reactor.core.publisher.Flux;
+
+import org.springframework.cloud.client.DefaultServiceInstance;
+import org.springframework.cloud.client.ServiceInstance;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class SubsetServiceInstanceListSupplierTest {
```
OlgaMaciaszek, line 35: Please add javdoc ("Tests for {@link linkToClass}."). Please add `@author` tag to the javadoc.

spring-cloud-loadbalancer/src/test/java/org/springframework/cloud/loadbalancer/core/SubsetServiceInstanceListSupplierTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+	private final DiscoveryClientServiceInstanceListSupplier delegate = mock(
+			DiscoveryClientServiceInstanceListSupplier.class);
+
+	@Test
+	void shouldConstructCorrect() {
```
OlgaMaciaszek, line 41: As we switch from using multiple constructors to getting the values from properties (default or set), this test will have to be modified.

spring-cloud-loadbalancer/src/test/java/org/springframework/cloud/loadbalancer/core/SubsetServiceInstanceListSupplierTest.java
```diff
@@ -0,0 +1,124 @@
...
+		when(delegate.get()).thenReturn(Flux.just(instances));
+
+		SubsetServiceInstanceListSupplier supplier1 = new SubsetServiceInstanceListSupplier(delegate, "foobar1", 5);
+		List<ServiceInstance> serviceInstances1 = Objects.requireNonNull(supplier1.get().blockFirst());
+
+		SubsetServiceInstanceListSupplier supplier2 = new SubsetServiceInstanceListSupplier(delegate, "foobar2", 5);
+		List<ServiceInstance> serviceInstances2 = Objects.requireNonNull(supplier2.get().blockFirst());
+
+		assertThat(serviceInstances1).isNotEqualTo(serviceInstances2);
+	}
```
OlgaMaciaszek, line 122: Thanks for adding exhaustive test scenarios.

OlgaMaciaszek (changes requested):
Have added some comments. Please address. Please also add corresponding `ServiceInstanceListSupplierBuilder` methods and `configurations` property-based instantiation in `LoadBalancerClientConfiguration`.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/SubsetServiceInstanceListSupplier.java
jizhuozhi: Hello @OlgaMaciaszek .`spring.application.name` cannot satisfy this scenario. In fact, there are now a large number of request instances (thousands or tens of thousands), each consumer instance needs to select a deterministic subset from them. At the same time, because the consumer may have tens of thousands of instances providing consistent services, it cannot be distinguished by `spring.application.name`.

So in my mind, I prefer to use the following information registered in Eureka. Is there any way to get an id like `192.168.0.1:eureka:8761`?

image

OlgaMaciaszek: Ok, I see the issue, but I don't think getting this info from Eureka is the solution, since that's just one of the supported SR implementations, so we should only use whatever information we can get from `DiscoveryClient` that supports all the implementations, and we don't get this kind of data there; your initial `UUID.random()` default might actually be the best in this case; @spencergibb @ryanjbaxter wdyt?

OlgaMaciaszek: @jizhuozhi, I've chatted with @spencergibb, who's rightly pointed out that we could use [this method](https://github.com/spring-cloud/spring-cloud-commons/blob/41ae9a429761e7fbd465cae181be26897f7ea6a4/spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/IdUtils.java#L44-L61) for setting the `clientId`. Please take a look.

jizhuozhi: I think it will be useful that using `DEFAULT_SERVICE_ID_STRING ` as default but user could custom the expression
 https://github.com/spring-cloud/spring-cloud-commons/blob/41ae9a429761e7fbd465cae181be26897f7ea6a4/spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/IdUtils.java#L30

OlgaMaciaszek (changes requested):
Have added some comments - please address them.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/SubsetServiceInstanceListSupplier.java
```diff
@@ -0,0 +1,91 @@
...
+	private final String instanceId;
+
+	private final int size;
+
+	public SubsetServiceInstanceListSupplier(ServiceInstanceListSupplier delegate, PropertyResolver resolver,
+			ReactiveLoadBalancer.Factory<ServiceInstance> factory) {
+		super(delegate);
+		LoadBalancerProperties properties = factory.getProperties(getServiceId());
+		String instanceId = properties.getSubset().getInstanceId();
+		if (StringUtils.hasText(instanceId)) {
```
OlgaMaciaszek, line 52: Please document this entire logic.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/core/ServiceInstanceListSupplierBuilder.java
```diff
@@ -301,6 +302,16 @@ public ServiceInstanceListSupplierBuilder withHints() {
...
+	public ServiceInstanceListSupplierBuilder withSubset() {
+		DelegateCreator creator = (context, delegate) -> {
+			PropertyResolver resolver = context.getBean(PropertyResolver.class);
+			LoadBalancerClientFactory factory = context.getBean(LoadBalancerClientFactory.class);
+			return new SubsetServiceInstanceListSupplier(delegate, resolver, factory);
+		};
+		creators.add(creator);
+		return this;
+	}
+
```
OlgaMaciaszek, line 314: Please also add  `configurations` property-based instantiation in `LoadBalancerClientConfiguration`.

OlgaMaciaszek (changes requested):

OlgaMaciaszek:
Hi @jizhuozhi, will you be able to finish this PR up soon? I'd like to get this feature included in `4.1.0`.

spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/annotation/LoadBalancerClientConfiguration.java
```diff
@@ -143,6 +143,15 @@ public ServiceInstanceListSupplier weightedServiceInstanceListSupplier(Configura
 					.build(context);
 		}
 
+		@Bean
```
OlgaMaciaszek, line 146: Thanks. We should also have one with the non-reactive discovery client (`@ConditionalOnBean(DiscoveryClient.class`). Also, please provide documentation for your feature.

OlgaMaciaszek (changes requested):
Have added a comment. Please address.

OlgaMaciaszek (changes requested):
Thanks @jizhuozhi, now only documentation, including this: https://github.com/spring-cloud/spring-cloud-commons/pull/1289#discussion_r1380232560, is left. Please add corresponding changes in `docs/modules/ROOT/pages/spring-cloud-commons/loadbalancer.adoc`.

jizhuozhi:
> Thanks @jizhuozhi, now only documentation, including this: [#1289 (comment)](https://github.com/spring-cloud/spring-cloud-commons/pull/1289#discussion_r1380232560), is left. Please add corresponding changes in `docs/modules/ROOT/pages/spring-cloud-commons/loadbalancer.adoc`.

Hello, @OlgaMaciaszek . The document has been updated, PTAL

OlgaMaciaszek (approved):
Thanks for working on it, @jizhuozhi. LGTM.

OlgaMaciaszek:
Hi @Buzzardo, could you please review documentation changes?

OlgaMaciaszek:
Fixes gh-1288.

docs/modules/ROOT/pages/spring-cloud-commons/loadbalancer.adoc
```diff
@@ -379,6 +379,34 @@ For `WebClient`, you need to implement and define `LoadBalancerClientRequestTran
...
+		return ServiceInstanceListSupplier.builder()
+					.withDiscoveryClient()
+					.withSubset()
+					.withCaching()
+					.build(context);
+	    }
+	}
+----
+
+TIP: Each service instance will be assigned a unique `instanceId` by default, and different `instanceId` will often select different subsets. Normally you don't need to pay attention to it, but if you need to have multiple instances select the same subset, you can set it via `spring.cloud.loadbalancer.subset.instance-id` (supports placeholders).
```
Buzzardo, line 406: Change to:

TIP: By default, each service instance is assigned a unique `instanceId`, and different `instanceId` values often select different subsets. Normally, you need not pay attention to it. However, if you need to have multiple instances select the same subset, you can set it with `spring.cloud.loadbalancer.subset.instance-id` (which supports placeholders).

docs/modules/ROOT/pages/spring-cloud-commons/loadbalancer.adoc
```diff
@@ -379,6 +379,34 @@ For `WebClient`, you need to implement and define `LoadBalancerClientRequestTran
...
+
+TIP: The size of the subset is set to 100 by default, you can also set it via `spring.cloud.loadbalancer.subset.size`
```
Buzzardo, line 408: Change to: 

TIP: By default, the size of the subset is set to 100. You can also set it with `spring.cloud.loadbalancer.subset.size`.

Buzzardo (changes requested):

Buzzardo (approved):

PR #1291, opened 2023-10-19
Fix Add a specific description of ServiceInstance is empty

RealBeBetter:
If the ServiceInstance is empty, the prompt information is too crude, which will increase the difficulty of user investigation.


RealBeBetter:
The following is the error when using: 
image

RealBeBetter:
> Thanks, @RealBeBetter. This looks good, but can you please submit it against `3.1.x` instead of `main`, like this we will include the change in all the supported versions. Also, please update the copyright entry on top of the file to say `2012-2023`, since the file is being updated.

I have recreated a new Pull Request. Please check.

OlgaMaciaszek:
To do that you need to make a change in your `3.1.x` branch, not `main`. And then create the PR. from your `3.1.x` to Spring Cloud `3.1.x`.

RealBeBetter:
> To do that you need to make a change in your `3.1.x` branch, not `main`. And then create the PR. from your `3.1.x` to Spring Cloud `3.1.x`.


Sorry, I have resubmitted this time. The address is: 
[pull request](https://github.com/spring-cloud/spring-cloud-commons/pull/1297)

PR #1297, opened 2023-11-11
Fix Add a specific description of ServiceInstance is empty

RealBeBetter:
As you can see, this is the re-submitted pull request.


OlgaMaciaszek (commented):
Thanks @RealBeBetter. LGTM.

PR #1321, opened 2024-01-16
Default value of `spring.cloud.loadbalancer.retry.enabled` remains consistent

DanielLiu1123:
This PR addresses one issue: The default value for `spring.cloud.loadbalancer.retry.enabled` is true, but this has not been consistently applied in some code.


OlgaMaciaszek:
Hello, @DanielLiu1123. Thank you for working on this PR. I have reviewed it. However, in this case, we will not be accepting it, since this behaviour is there on purpose, [as stated in the docs](https://docs.spring.io/spring-cloud-commons/reference/spring-cloud-commons/common-abstractions.html#retrying-failed-requests). That's because while with the blocking implementations, there's a dependency that needs to be added to trigger the retry behaviour (Spring Retry), in the case of the non-blocking implementation, there's no such starter, as the retry implementation is part of reactor-core, and the flag is used to opt into it.

Issue #1302, opened 2023-11-28
WebClient's URI_TEMPLATE_ATTRIBUTE ignored by MicrometerStatsLoadBalancerLifecycle


JaroslawDembek:
Version: 4.0.4

### Problem
When `spring.cloud.loadbalancer.stats.micrometer.enabled: true` using WebClient with loadbalancing can lead to generation of huge number of metrics when path contains params, e.g. `/test/{somePerCallId}/...` what eventually with high workload can cause even OOM.

### Common approach
Spring metrics collectors for WebClient uses URI_TEMPLATE_ATTRIBUTE (which contains placeholder instead of actual param value) to overcome problem mention above.

### Solution:
Make MicrometerStatsLoadBalancerLifecycle URI_TEMPLATE_ATTRIBUTE aware.
Now this loggic is hidden in `LoadBalancerTags` utility class.
```java
private static String getPath(RequestData requestData) {
     return requestData.getUrl() != null ? requestData.getUrl().getPath() : UNKNOWN;
}
```
Possibly MicrometerStatsLoadBalancerLifecycle could be aware of client type.

### Alternative:
Set `spring.cloud.loadbalancer.stats.micrometer.enabled: false` and deliver you own metrics collector as bean implementing `LoadBalancerLifecycle` to override `LoadBalancerTags.getPath` with e.g.
```java
private static String getPath(RequestData requestData) {
   if (requestData.getAttributes() != null) {
	var uriTemplate = (String) requestData.getAttributes().get(URI_TEMPLATE_ATTRIBUTE);
	if (uriTemplate != null) {
		return uriTemplate;
	}
  }
  return requestData.getUrl() != null ? requestData.getUrl().getPath() : UNKNOWN;
}
```

OlgaMaciaszek:
Hi @JaroslawDembek, thanks for raising the issue.  Would you like to submit a PR with the changes? If yes, please keep in mind that since this introduces a possibly breaking behaviour change, then method would need to use a flag to go the old path or the new one, with the flag already being deprecated since we'd transition to the new behaviour as the only one with the next major release, and all the changes should be submitted against the `4.0.4`. If you don't want to work on the PR, no worries - we'll take care of it.

JaroslawDembek:
I am already using overrided version of metric collector with SpringBoot3 so it shouldn't be a big task.
I will jump on it next week.

JaroslawDembek:
[...]

PR #1327, opened 2024-01-26
URI_TEMPLATE_ATTRIBUTE used when available. Fixes gh-1302

JaroslawDembek:


OlgaMaciaszek:
Hello @JaroslawDembek, thanks a lot. `4.0.x` is out of support now (the last release just happening today), so let's do the PR against `main` instead (I'm sorry for any confusion here). The feature flag would just be a `boolean` property that could be accessed from `MicrometerStatsLoadBalancerLifecycle`, as it is a bean. Then in `getPath`, you'd go one way or another, depending on the value of that property. The build is failing - please take a look.  Let me know if you need additional pointers.

JaroslawDembek:
Deprecated base branch.

PR #1339, opened 2024-02-19
Allow building @LoadBalanced RestClient in component constructor.

OlgaMaciaszek:
Load balanced RestClient with deferring postprocessor.

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerAutoConfiguration.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerAutoConfiguration.java
@@ -33,12 +33,12 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.client.loadbalancer.reactive.ReactiveLoadBalancer;
+import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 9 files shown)


spencergibb (approved):

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerAutoConfiguration.java
```diff
@@ -79,26 +75,33 @@ public SmartInitializingSingleton loadBalancedRestTemplateInitializerDeprecated(
...
-
 	@Bean
 	@ConditionalOnMissingBean
 	public LoadBalancerRequestFactory loadBalancerRequestFactory(LoadBalancerClient loadBalancerClient) {
 		return new LoadBalancerRequestFactory(loadBalancerClient, transformers);
 	}
 
-	@Configuration(proxyBeanMethods = false)
+	@AutoConfiguration
+	static class DeferringLoadBalancerInterceptorConfig {
```
ryanjbaxter, line 85: I assume these inner classes get autoconfigured because the outerclass is specified in spring.factories?

OlgaMaciaszek:
I'm really sorry for the mess around that deprecation @wind57. In fact, it was being load-balanced [with the previous change](https://github.com/spring-cloud/spring-cloud-commons/pull/1319), however, with the caveat that a component constructor would be to early to  build the `RestClient` instance with a `@LoadBalanced RestClient.Builder`, which is why it's been reworked again.

OlgaMaciaszek:
The `@LoadBalanced` annotation over in:
`@LoadBalanced
	@Autowired(required = false)
	private List<RestClient.Builder> restClientBuilders = Collections.emptyList();`
works as a qualifier; that's also how it's always been done for `RestTemplate`, just too early for `RestClient` for some of the scenarios, since there we work with builder instances and not actual client instances.

OlgaMaciaszek:
No worries, it's not very straightforward.

PR #1349, opened 2024-04-04
Add the ability for FailsafeTextEncryptor to have a delegate.

ryanjbaxter:
Related to https://github.com/spring-cloud/spring-cloud-config/issues/2330


spencergibb (approved):
Maybe document here why a delegate, since it's odd without context.

PR #1360, opened 2024-06-29
multi-main-zone-based-load-balancing

seal90:
To ensure the stability of the online environment, before releasing it to the production environment, a pre-release environment will be used to use all the middleware in production to verify whether the new changes are stable. When it is released to production, a small amount of traffic will be used to verify whether the changes meet expectations. At this time, isolation of the main environment and grayscale isolation are required.


OlgaMaciaszek:
Hello @seal90. We have discussed it as a team and we think this is very specific to your setup, so we're not going to be merging this. The way to go is to keep using your custom `ServiceInstanceListSupplier`, which is a supported scenario. However, thank you for your work on this PR.

PR #1403, opened 2024-10-11
Support @LoadBalanced RestTemplateBuilder

OlgaMaciaszek:
Fixes gh-1402


spencergibb (approved):
To summarize, you took the rest template post processor, pulled out the common bits into an abstract base class then added the rest template builder support. Did I get that right?

ryanjbaxter (approved):

OlgaMaciaszek:
> To summarize, you took the rest template post processor, pulled out the common bits into an abstract base class then added the rest template builder support. Did I get that right?

Yes, just it was the `RestClient.Builder` post-processor.

PR #1445, opened 2025-01-20
Remove workaround for data source cycle

quaff:
It should be fixed at Spring Boot side, see https://github.com/spring-projects/spring-boot/commit/2f83a6714bf258c3fcc8ca92c89f02cf2043c97a

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/RefreshAutoConfiguration.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/RefreshAutoConfiguration.java
@@ -23,9 +23,6 @@ import java.util.Set;
 import org.springframework.aop.scope.ScopedProxyUtils;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanFactory;
-import org.springframework.beans.factory.InitializingBean;
-import org.springframework.beans.factory.ListableBeanFactory;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.config.BeanDefinition;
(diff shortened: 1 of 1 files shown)


quaff:
I have a feeling that `JpaInvokerConfiguration` is not needed anymore with recent Spring Framework. @dsyer

spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/RefreshAutoConfiguration.java
```diff
@@ -158,27 +153,6 @@ public String toString() {
...
-	@Configuration(proxyBeanMethods = false)
-	@ConditionalOnClass(name = "javax.persistence.EntityManagerFactory")
-	protected static class JpaInvokerConfiguration implements LoadTimeWeaverAware, InitializingBean {
-
-		@Autowired
-		private ListableBeanFactory beanFactory;
-
-		@Override
-		public void afterPropertiesSet() {
-			String cls = "org.springframework.boot.autoconfigure.jdbc.DataSourceInitializerInvoker";
```
ryanjbaxter, line 170: `DataSourceInitializerInvoker` was removed in the Spring Boot commit you referenced, correct?

Therefore the `if` statement below never actually gets executed, right?

quaff: Not exactly, Spring Boot rework DataSource initialization to fix it entirely, it can be proved since `@ConditionalOnClass(name = "javax.persistence.EntityManagerFactory")` never matches since it renamed to `jarkata.persistence.EntityManagerFactory`.

PR #1450, opened 2025-01-21
Move field ContextRefresher.REFRESH_ARGS_PROPERTY_SOURCE down to LegacyContextRefresher

quaff:
It's specific to LegacyContextRefresher.

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/ConfigDataContextRefresher.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/ConfigDataContextRefresher.java
@@ -92,9 +92,6 @@ public class ConfigDataContextRefresher extends ContextRefresher
 			postProcessor.postProcessEnvironment(environment, application);
 		}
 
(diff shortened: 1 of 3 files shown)


spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/ConfigDataContextRefresher.java
```diff
@@ -92,9 +92,6 @@ protected void updateEnvironment() {
 			postProcessor.postProcessEnvironment(environment, application);
 		}
 
-		if (environment.getPropertySources().contains(REFRESH_ARGS_PROPERTY_SOURCE)) {
```
ryanjbaxter, line 95: Why is this removed?

quaff: Because this `PropertySource` is added by `LegacyContextRefresher`, It's not involved if `ConfigDataContextRefresher` is using.

PR #1456, opened 2025-01-23
Deprecate unused EnvironmentUtils

quaff:
EnvironmentUtils is not used in all spring-cloud projects, see https://github.com/search?q=org%3Aspring-cloud%20getSubProperties&type=code


spencergibb (changes requested):
It will need to be deprecated before being removed in the next major later this year.

Issue #1329, opened 2024-01-26
Spring cloud bug with placeholders with comma in spring.profiles.include


alex-kormukhin:
**Describe the bug**
If we use placeholders with comma in spring.profiles.include (and spring.profiles.active) property like this:
`spring.profiles.include=${ENVIRONMENT_PROFILE_NAME:some_profile1,some_profile2}`

vanilla spring boot application works fine:
![2024-01-26_190115](https://github.com/spring-cloud/spring-cloud-commons/assets/99888299/7aa7942c-11df-4c6f-925e-56fb71afd80f)

but spring cloud application (with consul config, for example) **with spring-cloud-starter-bootstrap** (bootstrap mode) can't resolve such placeholders:
![2024-01-26_190209](https://github.com/spring-cloud/spring-cloud-commons/assets/99888299/9ce508b9-91c3-416a-9087-a2557bfb91eb)

If also set 
`spring.cloud.config.initialize-on-context-refresh=true`
cloud application crashes on start like this:
`2024-01-26T19:28:33.299+03:00 ERROR 3808 --- [           main] o.s.boot.SpringApplication               : Application run failed

org.springframework.cloud.consul.config.ConsulPropertySources$PropertySourceNotFoundException: java.lang.IllegalArgumentException: Illegal character in path at index 48: http://localhost:8500/v1/kv/config/application,${ENVIRONMENT_PROFILE_NAME:some_profile1/?recurse&token=
        at org.springframework.cloud.consul.config.ConsulPropertySources.createPropertySource(ConsulPropertySources.java:143) ~[spring-cloud-consul-config-4.1.0.jar:4.1.0]
        at org.springframework.cloud.consul.config.ConsulPropertySourceLocator.locate(ConsulPropertySourceLocator.java:88) ~[spring-cloud-consul-config-4.1.0.jar:4.1.0]
        at org.springframework.cloud.bootstrap.config.PropertySourceLocator.locateCollection(PropertySourceLocator.java:50) ~[spring-cloud-context-4.1.0.jar:4.1.0]
        at org.springframework.cloud.consul.config.ConsulPropertySourceLocator.locateCollection(ConsulPropertySourceLocator.java:71) ~[spring-cloud-consul-config-4.1.0.jar:4.1.0]
[...]

PR #1464, opened 2025-02-06
Let `spring.profiles.[include|active]` accept list as well as comma separated value in bootstrapProperties

### LINKED ISSUE #1417: spring.profiles.include must be a comma separated value in contrast to to non-cloud usage
**Describe the bug**
According to [the documentation](https://docs.spring.io/spring-boot/reference/features/profiles.html#features.profiles.adding-active-profiles) for `spring.profiles.include` the value for `spring.profiles.include` is a list.

```
spring.profiles.include[0]=common
spring.profiles.include[1]=local
```

But using this in a configuration file provided via cloud config it does not work. Instead it expects the value to be a comma separated string value:

```yaml
spring.profiles.include=common,local
```

The offending code:

https://github.com/spring-cloud/spring-cloud-commons/blob/main/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/config/PropertySourceBootstrapConfiguration.java#L306-L317

In Spring Boot the configuration setting is processed as a string list:

https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigDataEnvironment.java#L291-L310

The difference in behavior is not documented anywhere.

quaff:
Fix GH-1417

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/config/PropertySourceBootstrapConfiguration.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/config/PropertySourceBootstrapConfiguration.java
@@ -34,6 +34,8 @@ import org.springframework.boot.context.config.Profiles;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.boot.context.properties.bind.Bindable;
 import org.springframework.boot.context.properties.bind.Binder;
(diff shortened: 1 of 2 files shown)


quaff:
It fix https://github.com/spring-cloud/spring-cloud-commons/issues/1329 also.

quaff:
Could any team member review?

PR #1466, opened 2025-02-07
Polish note about @ConfigurationProperties can't be refreshed

quaff:


spencergibb:
Your change implies other types can't be refreshed. Can you describe why you want this change?

quaff:
> Your change implies other types can't be refreshed. Can you describe why you want this change?

My point is not only Java Records but any type using constructor binding.

See https://github.com/spring-projects/spring-boot/blob/1e73a7cf8b376ea4b5ead9e56df093e5354c4de7/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/BindMethod.java#L33C18-L33C43

spencergibb:
Then say that rather than value object.

spencergibb (changes requested):
Change value object to something about constructor binding.

quaff:
I copied terms from Javadoc of Spring Boot, WDYT?

PR #1467, opened 2025-02-07
Use framework's AopProxyUtils to take care of CGLIB Proxy

### LINKED ISSUE #1308: spring.cloud.refresh.never-refreshable not ignoring class
**Describe the bug**
Hi folks,

I have a class name set in spring.cloud.refresh.never-refreshable to be ignored for refresh. 

This however doesn't work because Spring adds some EnhancerBySpringCGLIB<somenumeric> to the class name.

Thus, this snippet  org.springframework.cloud.context.properties.ConfigurationPropertiesRebinder#getNeverRefreshable fails when calling getClass().getName()

**Sample**
1.)  set prop: spring.cloud.refresh.never-refreshable=com.sth.IgnoreMe

2.) bean.getClass().getName() - returns during runtime a modified com.sth.IgnoreMe$$EnhancerBySpringCGLIB$$aaffeeaa and the comparison against the prop value of com.sth.IgnoreMe fails.

quaff:
Fix GH-1308

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/context/properties/ConfigurationPropertiesRebinder.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/context/properties/ConfigurationPropertiesRebinder.java
@@ -21,6 +21,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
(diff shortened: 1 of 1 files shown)


spencergibb (changes requested):
Add a test that fails before this change

quaff:
@nitron2x Could you provide a minimal reproducer example?

Issue #1322, opened 2024-01-17
InstanceRegisteredEvent can implement ResolvableTypeProvider interface for better convenience


XhstormR:
Due to Java type erasure, currently I have to manually determine the type of the event's config when listening to this event:
```kotlin
@Component
class EurekaInstanceConfigPostProcessor(
    private val eureka: EurekaRegistration,
) : ApplicationListener<InstanceRegisteredEvent<*>> {

    override fun onApplicationEvent(event: InstanceRegisteredEvent<*>) {
        if (event.config is EurekaInstanceConfigBean){
            eureka.applicationInfoManager.registerAppMetadata(......)
        }
    }
}
```
If InstanceRegisteredEvent can implement ResolvableTypeProvider, i can directly listening and set it's config type:
```kotlin
@Component
class EurekaInstanceConfigPostProcessor(
    private val eureka: EurekaRegistration,
) : ApplicationListener<InstanceRegisteredEvent<EurekaInstanceConfigBean>> {

    override fun onApplicationEvent(event: InstanceRegisteredEvent<EurekaInstanceConfigBean>) {
        eureka.applicationInfoManager.registerAppMetadata(......)
    }
}
```

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/ResolvableTypeProvider.html

quaff:
It should be fixed by https://github.com/spring-cloud/spring-cloud-commons/pull/1468.

PR #1468, opened 2025-02-07
Make InstanceRegisteredEvent implements ResolvableTypeProvider

quaff:
Fix GH-1322


ryanjbaxter:
This will need to wait for the next major since it adds an interface to a public class.  It could break binary compatibility since this is effecting remote events.

PR #1490, opened 2025-05-13
5.0.x

OlgaMaciaszek:
Adjust to changes in Framework.

### DIFF
--- a/spring-cloud-commons/src/test/java/org/springframework/cloud/configuration/SpringBootDependencyTests.java
+++ b/spring-cloud-commons/src/test/java/org/springframework/cloud/configuration/SpringBootDependencyTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors.
+ * Copyright 2012-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,6 +27,7 @@ import static org.assertj.core.api.BDDAssertions.then;
 
 /**
  * @author Marcin Grzejszczak
+ * @author Olga Maciaszek-Sharma
  */
 public class SpringBootDependencyTests {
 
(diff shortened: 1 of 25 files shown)


spring-cloud-commons/src/test/java/org/springframework/cloud/configuration/SpringBootDependencyTests.java
```diff
@@ -204,15 +205,15 @@ private void verifyCurrentVersionFromManifest(String version) {
 
 	@Test
 	public void should_match_against_current_predicate() {
-		List<String> acceptedVersions = Collections.singletonList("3.0");
```
spencergibb, line 207: I didn't think this one needs to change

spring-cloud-loadbalancer/src/test/java/org/springframework/cloud/loadbalancer/core/XForwardedHeadersTransformerTests.java
```diff
@@ -66,9 +66,9 @@ void shouldAppendXForwardedHeadersIfEnabled() {
 
 		ClientRequest newRequest = transformer.transformRequest(request, serviceInstance);
 
-		assertThat(newRequest.headers()).containsKey("X-Forwarded-Host");
+		assertThat(newRequest.headers().containsHeader("X-Forwarded-Host")).isTrue();
 		assertThat(newRequest.headers().getFirst("X-Forwarded-Host")).isEqualTo("spring.io");
-		assertThat(newRequest.headers()).containsKey("X-Forwarded-Proto");
+		assertThat(newRequest.headers().containsHeader("X-Forwarded-Host")).isTrue();
```
spencergibb, line 71: This should stay proto?

spencergibb (approved):

spring-cloud-commons/src/test/java/org/springframework/cloud/configuration/SpringBootDependencyTests.java
OlgaMaciaszek: If we don't change it, we'll get:
```
java.lang.AssertionError: 
Expecting empty but was: "Spring Boot is not compatible with this Spring Cloud release train"
```
in the test.

spring-cloud-loadbalancer/src/test/java/org/springframework/cloud/loadbalancer/core/XForwardedHeadersTransformerTests.java
OlgaMaciaszek: Good catch, thanks. Fixing it.

PR #1491, opened 2025-05-14
Add LoadBalancer integration for Spring Interface Clients AutoConfiguration

OlgaMaciaszek:


docs/modules/ROOT/pages/spring-cloud-commons/loadbalancer.adoc
```diff
@@ -554,3 +554,14 @@ NOTE: Starting with `4.1.0`, we have introduced the `callGetWithRequestOnDelegat
 
 Since `4.0.0`, Spring Cloud LoadBalancer supports Spring AOT transformations and native images. However, to use this feature, you need to explicitly define your `LoadBalancerClient` service IDs. You can do so by using the `value` or `name` attributes of the `@LoadBalancerClient` annotation or as values of the `spring.cloud.loadbalancer.eager-load.clients` property.
 
+== LoadBalancer Integration for Spring Interface Clients AutoConfiguration
+
+Since `5.0.0`, Spring Cloud LoadBalancer supports https://docs.spring.io/spring-framework/reference/7.0-SNAPSHOT/integration/rest-clients.html#rest-http-interface[Spring Interface Clients] AutoConfiguration through the `LoadBalancerRestClientHttpServiceGroupConfigurer` and `LoadBalancerWebClientHttpServiceGroupConfigurer`.
+
+For each Interface Client group, if the group `baseUrl` (defined under the
```
ryanjbaxter, line 561: Some examples illustrating these difference scenarios might make things clearer

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerRestClientHttpServiceGroupConfigurer.java
```diff
@@ -0,0 +1,103 @@
...
+			ReactiveLoadBalancer.Factory<ServiceInstance> loadBalancerClientFactory) {
+		this.loadBalancerInterceptorSupplier = SingletonSupplier
+			.ofNullable(loadBalancerInterceptorProvider.getIfAvailable());
+		this.clientServiceProperties = clientServiceProperties;
+		this.loadBalancerClientFactory = loadBalancerClientFactory;
+	}
+
+	@Override
+	public void configureGroups(@NonNull Groups<RestClient.Builder> groups) {
+		DeferringLoadBalancerInterceptor loadBalancerInterceptor = loadBalancerInterceptorSupplier.get();
```
ryanjbaxter, line 72: In what scenario would `DeferringLoadBalancerInterceptor` not be available?

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/reactive/LoadBalancerWebClientHttpServiceGroupConfigurer.java
```diff
@@ -0,0 +1,104 @@
...
+			ReactiveLoadBalancer.Factory<ServiceInstance> loadBalancerClientFactory) {
+		this.loadBalancerFilterFunctionSupplier = SingletonSupplier
+			.ofNullable(exchangeFilterFunctionProvider.getIfAvailable());
+		this.clientServiceProperties = clientServiceProperties;
+		this.loadBalancerClientFactory = loadBalancerClientFactory;
+	}
+
+	@Override
+	public void configureGroups(Groups<WebClient.Builder> groups) {
+		DeferringLoadBalancerExchangeFilterFunction<LoadBalancedExchangeFilterFunction> loadBalancerFilterFunction = loadBalancerFilterFunctionSupplier
```
ryanjbaxter, line 72: Same as above, when would this not be available?

OlgaMaciaszek: TBH, mainly if the user has manually played with the configuration and botched it.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerRestClientHttpServiceGroupConfigurer.java
OlgaMaciaszek: TBH, mainly if the user has manually played with the configuration and botched it.

ryanjbaxter: Would it be better then for the auto-configuration to not even run in that case?

OlgaMaciaszek: Honestly, I don't really expect that to happen unless someone completely misconfigures the application, more of a defensive setup, but if you think it's too much, I could just not use the `ofNullable` option to retrieve it and instead expect it to be there.

ryanjbaxter: It just seems easier to me (without looking too deeply into it) to not enable the auto-configuration if the classes we need to do so are not present.  At least that is more of the pattern we have followed in the past.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerProperties.java
```diff
@@ -572,4 +585,22 @@ public void setIncludePath(boolean includePath) {
 
 	}
 
+	public static class InterfaceClients {
```
spencergibb, line 588: Seems a bit strange for the load balancer to know about interface clients

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerRestClientHttpServiceGroupConfigurer.java
```diff
@@ -0,0 +1,103 @@
...
+	private final SingletonSupplier<DeferringLoadBalancerInterceptor> loadBalancerInterceptorSupplier;
+
+	private final HttpClientServiceProperties clientServiceProperties;
+
+	public LoadBalancerRestClientHttpServiceGroupConfigurer(
+			ObjectProvider<DeferringLoadBalancerInterceptor> loadBalancerInterceptorProvider,
+			HttpClientServiceProperties clientServiceProperties,
+			ReactiveLoadBalancer.Factory<ServiceInstance> loadBalancerClientFactory) {
+		this.loadBalancerInterceptorSupplier = SingletonSupplier
+			.ofNullable(loadBalancerInterceptorProvider.getIfAvailable());
```
spencergibb, line 65: Why not just use the provider rather than wrapping it?

OlgaMaciaszek: Right that didn't make much sense. What I actually was planning to do is: https://github.com/spring-cloud/spring-cloud-commons/pull/1491/commits/6d479fa3c89d64badb11a57f3a648e2ef292869, so to be able to store the value after first access, but at the same time differ until required to avoid issues with too early access that we;ve seen sometimes.

spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerProperties.java
OlgaMaciaszek: Removed

docs/modules/ROOT/pages/spring-cloud-commons/loadbalancer.adoc
ryanjbaxter: I dont see them 🧐

ryanjbaxter (approved):

PR #1496, opened 2025-05-29
remove duplicate line in document

JongP:
there's duplicate line in config property, `spring.cloud.loadbalancer.health-check.interval`
it seems it's unnecessary

I'm new to open source. please let me know anything wrong.

### DIFF
--- a/docs/modules/ROOT/partials/_configprops.adoc
+++ b/docs/modules/ROOT/partials/_configprops.adoc
@@ -44,7 +44,6 @@
 |spring.cloud.loadbalancer.enabled | `+++true+++` | Enables Spring Cloud LoadBalancer.
 |spring.cloud.loadbalancer.health-check.initial-delay | `+++0+++` | Initial delay value for the HealthCheck scheduler.
 |spring.cloud.loadbalancer.health-check.interval | `+++25s+++` | Interval for rerunning the HealthCheck scheduler.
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
Thanks for reporting this!

This is actually a bit non-intuitive.  To solve this we actually don't need to change `_configprops.adoc`, this file is automatically generated during the build so any change there will just be overwritten eventually and the problem will show up again.

The problem here is that we have added the property unnecessarily here: https://github.com/spring-cloud/spring-cloud-commons/blob/4.1.x/spring-cloud-commons/src/main/resources/META-INF/additional-spring-configuration-metadata.json#L62-L67

When the metadata will automatically be generated from the `@ConfigurationProperties` class here https://github.com/spring-cloud/spring-cloud-commons/blob/4.1.x/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/LoadBalancerProperties.java#L225

That is why it ends up duplicated in `_configprops.adoc`.

The real change needs to be made in `additional-spring-configuration-metadata.json` basically removing that property.

Then if you run `./mvnw clean package -Pdocs` you should see the elimination of the duplicate property in `_configprops.adoc`.

Also it would be great if you make this PR against the 4.1.x branch and we can then merge it forward to the other branches since they all have the same problem.

JongP:
@ryanjbaxter 
thanks for review!

I followed your instructions and found there's no duplicate metadata in `target/classes/META-INF/spring-configuration-metadata.json `.


target branch is also updated.

Issue #1515, opened 2025-06-25
Support per-service load balancing strategy configuration in Spring Cloud LoadBalancer


SkyeBeFreeman:
**Is your feature request related to a problem? Please describe.**
I want to configure different load balancing strategies (e.g., roundrobin, random, or custom) for different services individually. Currently, I’m not sure if Spring Cloud LoadBalancer supports per-service strategy configuration similar to spring.cloud.loadbalancer.clients.<service-name>.configurations. Without this, I have to apply a global strategy for all services, which is not flexible enough for my use case.

**Describe the solution you'd like**
I would like to be able to specify the load balancing strategy on a per-service basis via configuration, for example:

```yaml
spring:
  cloud:
    loadbalancer:
      clients:
        serviceA:
          strategy: roundrobin
        serviceB:
          strategy: random
```
or a similar approach that allows assigning different built-in or custom load balancing strategies to each service.

**Describe alternatives you've considered**
Currently, I can only configure a global load balancing strategy for all services or implement custom logic programmatically, which is more complex and less convenient than a configuration-based approach.

**Additional context**
This feature would improve flexibility and make it easier to tailor load balancing behavior per service, especially in microservice architectures where different services may have different load balancing requirements.

If possible, I would be happy to try implementing this feature and contribute a PR.

Thanks for your consideration!

PR #1516, opened 2025-06-25
feat:support setting load balancing strategy per service.

SkyeBeFreeman:
solves #1515


SkyeBeFreeman:
@spencergibb Please give me suggestions or feedback. Thank you.

PR #1517, opened 2025-06-26
Fix incorrect usage of `@AutoConfiguration`

quaff:
>> Generally, auto-configuration classes are top-level classes


spencergibb (commented):
Are those classes listed in the auto configuration file?

quaff:
> Are those classes listed in the auto configuration file?

No.

PR #1532, opened 2025-07-17
Add interface clients support for circuit breaker

OlgaMaciaszek:
Sample branches that demonstrate the usage: https://github.com/OlgaMaciaszek/interface-clients-autoconfiguration-demo/tree/client-with-loadbalancer-and-cb and https://github.com/OlgaMaciaszek/interface-clients-autoconfiguration-demo/tree/client-with-loadbalancer-webflux-cb

Fixes gh-1538.


docs/modules/ROOT/pages/spring-cloud-circuitbreaker.adoc
```diff
@@ -107,25 +107,73 @@ Customizer.once(circuitBreaker -> {
...
+When fallback classes are configured using the `@HttpServiceFallbackAnnotation`,
+CircuitBreaker adapter decorators are added:
 - `CircuitBreakerAdapterDecorator` is used with `RestClient`
 - `ReactiveCircuitBreakerAdapterDecorator` is used with `WebClient`
 
+[NOTE]
+====
+You can disable CircuitBreaker integration for HTTP service clients by setting the appropriate property:
+
+- For blocking (`RestClient`) clients: `spring.cloud.circuitbreaker.http-services.enabled=false`
```
ryanjbaxter, line 126: Is this necessary?  Why would they add the annotations if they didn't want to use them?

docs/modules/ROOT/pages/spring-cloud-circuitbreaker.adoc
```diff
@@ -107,25 +107,73 @@ Customizer.once(circuitBreaker -> {
...
+
+This configuration results in:
+
+- `DefaultFallbacks` used as a global fallback for all services not explicitly handled
+- `GroupAndServiceSpecificFallbacks` used only for `BillingService` and `ShippingService` within the `"billing"` group
+
+[NOTE]
+====
+- The fallback class and its methods must be `public`
+- Fallback methods must not be annotated with `@HttpExchange`-specific annotations
```
ryanjbaxter, line 169: Change to "Fallback methods must not be annotated with `@HttpExchange`"?

spring-cloud-commons/src/main/java/org/springframework/cloud/client/circuitbreaker/httpservice/HttpServiceFallback.java
```diff
@@ -0,0 +1,115 @@
...
+
+	/**
+	 * The optional list of service interfaces this fallback class applies to.
+	 * <p>
+	 * If omitted, the fallback is treated as a default for all service interfaces within
+	 * the applicable group.
+	 * </p>
+	 * @return the service interfaces this fallback applies to
+	 */
+	Class<?>[] forService() default {};
```
spencergibb, line 87: Maybe drop the `for` here and below?

spencergibb (approved):

docs/modules/ROOT/pages/spring-cloud-circuitbreaker.adoc
ryanjbaxter: Ah I see.  This is fine for a milestone.  I wonder if we should add some kind of marker class to `spring-cloud-circuitbreaker` to indicate its on the classpath and only run things if it is, thoughts?

OlgaMaciaszek: Right. Makes sense. Will create a separate PR.

docs/modules/ROOT/pages/spring-cloud-circuitbreaker.adoc
OlgaMaciaszek: Ok.

Issue #1580, opened 2025-09-30
Wrong set of active profiles as property substitution does not take place inside values


hosea:
Hello,

I've detected the following behavior when defining profiles using the following values (bootstrap.yml):

spring:
profiles:
active: stage_${DEPLOYMENT_STAGE:PRD}, domain_${LOADBALANCER_DOMAIN:mydomain}, env_${DEPLOYMENT_ENVIRONMENT:px}

Spring reports 6 active profiles, not 3. The reason is: the first 3 profiles are resolved using property substitution. So the result are 3 profiles: stage_PRD, domain_mydomain, env_px (assuming default-values).
But then additional 3 profiles are added: stage_${DEPLOYMENT_STAGE:PRD}, domain_${LOADBALANCER_DOMAIN:mydomain}, env_${DEPLOYMENT_ENVIRONMENT:px}
because the property substitution does not take place a second time.

I found the place: PropertySourceBootstrapConfiguration#resolvePlaceholdersInProfiles does not recognize placeholders inside the profile-value.

Kind regards
Hans

PR #1581, opened 2025-09-30
1580: Fixes property substitution problem for active profiles

hosea:
Fix fur Issue 1580: https://github.com/spring-cloud/spring-cloud-commons/issues/1580


ryanjbaxter:
Thanks!  Can you add a test that validates the fix?

PR #1604, opened 2025-11-22
Creates InstanceProperties

spencergibb:
This allows SimpleDiscoveryClient and SimpleReactiveDiscoveryClient to use InstanceProperties rather than DefaultServiceInstance.

This then allows ServiceInstance to remove nullable from host and serviceId.

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/DefaultServiceInstance.java
@@ -81,7 +81,7 @@ public class DefaultServiceInstance implements ServiceInstance {
 
 	@Override
 	public URI getUri() {
-		return ServiceInstance.getUri(this);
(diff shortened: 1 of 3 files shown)


spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/simple/InstanceProperties.java
```diff
@@ -0,0 +1,145 @@
...
+
+public class InstanceProperties {
+
+	private @Nullable String instanceId;
+
+	private @Nullable String serviceId;
+
+	private @Nullable String host;
+
+	private int port;
```
ryanjbaxter: How is `port` and `secure` not `@Nullable`?  They have no default value and are not set in the default constructor

ryanjbaxter: 🤦‍♂️

PR #1632, opened 2026-01-13
Fix decryption of indexed (list) properties.

### LINKED ISSUE #1253: Properties decryption not working correctly when they are on a list (yaml)
When we have properties on a list that are encrypted and we overwrite them on another profile they will keep their initial value and ignore other profiles. This happens even to other properties but if we add them also ciphered on the other profiles then it will work correctly. I have a sample application to reproduce the issue: https://github.com/FuRi0n/decrypt

In the sample, with this yml file:

```yml
application:
  list:
    - plain: good
      cipher: '{cipher}4ffef96df11b9e9ca7b33aa5ccd60b6ed8406ec5edb68618b2469208fceb2626'

---

application:
  list:
    - plain: updated
      cipher: 'updated'
spring:
  config:
    activate:
      on-profile: decrypt
```

When running 'decrypt' profile it will print on console plain='good' and cipher='bad' but, if we update the 'cipher' value on 'decrypt' profile to be also encrypted then we will get plain='updated' and cipher='updated'

The expected behaviour is that properties in the profile with higher priority must overwrite other profiles even if those properties values are not encrypted.

kvmw:
This pull request fixes #1253.  

It also fixes the following cases:

- In current (before this fix) implementation, when there is a list (indexed) property with encrypted items in a property source, **All** lists are copied to the new decrypted property-source, not only the one that has encrypted items. 

For example, in the following case only `my-list` should be copied to the decrypted property source but current implementation copies `another-list` as well. 
```
  my-list:
    - foo: '{cipher}4ffef9'
    - foo: 'plain'
    - foo: 'another plain value'
 
  another-list:
    - bar: plain
    - bar: another plain   
```

- When searching for properties, during decryption, relaxed-binding should be considered. Otherwise It is not possible to override the value of `foo.text : '{cipher}4ffef9'` by  adding `FOO_TEXT : 'updated value'`  in another property source with higher priority.

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/AbstractEnvironmentDecrypt.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/AbstractEnvironmentDecrypt.java
@@ -16,20 +16,19 @@
 
 package org.springframework.cloud.bootstrap.encrypt;
 
-import java.util.ArrayList;
-import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
-import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.regex.Pattern;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
-import org.springframework.core.env.CompositePropertySource;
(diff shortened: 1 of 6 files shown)


ryanjbaxter:
Could you submit this against the 4.3.x branch instead?

kvmw:
closing in favour of #1634 1634

PR #1662, opened 2026-03-10
Recreate proxied @ConfigurationProperties beans on rebind

### LINKED ISSUE #1616: ConfigurationProperties initial values not restored on DELETE /actuator/env
**Describe the bug**

We are setting a configuration value via `POST /actuator/env` and later attempt to restore the original value by running `DELETE /actuator/env`.
However, this does not work — Spring does **not** restore the values defined in the field declaration.

**Sample**

```java
@ConfigurationProperties("my-properties")
class MyProperties {
    private String name = "initial";
}
```

Change the value via actuator and delete the change:
```
curl -H 'Content-Type: application/json' --data '{"name": "my-properties.name", "value": "from-actuator"}' -X POST http://localhost:8080/actuator/env

curl -X DELETE http://localhost:8080/actuator/env
```

After this, the value is still `from-env` although I would have expected `initial` again.

It seems like`JavaBeanBinder` is not able to do that. (This is even tested [here](https://github.com/spring-projects/spring-boot/blob/main/core/spring-boot/src/test/java/org/springframework/boot/context/properties/bind/JavaBeanBinderTests.java#L147))

seonghyeoklee:
## Summary

When a `@ConfigurationProperties` bean is wrapped in an AOP proxy and properties are removed from the `Environment`, calling `rebind()` previously left field initializers unrestored (they became `null` / `0` / `false` instead of their declared default values).

This change modifies `ConfigurationPropertiesRebinder.rebind()` so that when the bean is an AOP proxy, a fresh instance is created via `createBean()` and the proxy's `TargetSource` is replaced, ensuring field initializers are properly restored.

Non-proxied beans continue to use the existing `destroyBean()` + `initializeBean()` behavior for backwards compatibility.

Fixes gh-1616

## Changes

- `ConfigurationPropertiesRebinder`: detect proxied beans and replace the proxy target with a freshly created instance
- Added `ConfigurationPropertiesRebinderFieldInitializerIntegrationTests` to verify field initializer restoration

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/context/properties/ConfigurationPropertiesRebinder.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/context/properties/ConfigurationPropertiesRebinder.java
@@ -21,6 +21,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
+import org.springframework.aop.framework.Advised;
 import org.springframework.aop.scope.ScopedProxyUtils;
 import org.springframework.aop.support.AopUtils;
 import org.springframework.beans.BeansException;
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
Can you merge in the latest changes from main so the build will pass?

seonghyeoklee:
Hi @ryanjbaxter, I've addressed the review feedback in c73c01c7:

1. **Nested proxy prevention**: `freshBean` is now unwrapped via `ProxyUtils.getTargetObject()` if `createBean()` returns an AOP proxy, avoiding double-proxied targets.
2. **Error safety**: The fresh instance is created *before* destroying the old target. If `setTargetSource()` fails, the fresh bean is cleaned up and the original target remains intact.

Tests pass. Would appreciate a review when you get a chance. Thanks!

ryanjbaxter:
Superseded by https://github.com/spring-cloud/spring-cloud-commons/pull/1680

Issue #1676, opened 2026-04-02
Add Grpc Client for Spring Cloud Loadbalancer


KouShenhai:
**Is your feature request related to a problem? Please describe.**
Currently, the spring-rpc project has not yet been adapted to support Spring Cloud load balancing. This requirement is quite common, and it is hoped that this functionality can be added.

**Describe the solution you'd like**
This is a Spring Cloud gRPC client identified by  `discovery://service-name` .

**Describe alternatives you've considered**
There is not,

**Additional context**
Refer to the open-source project [grpc-spring](https://github.com/grpc-ecosystem/grpc-spring)

PR #1677, opened 2026-04-02
Add Grpc Client for Spring Cloud Loadbalancer

KouShenhai:


pom.xml
```diff
@@ -88,6 +88,8 @@
 		<evictor.version>1.0.0</evictor.version>
 		<bouncycastle-bcprov-jdk18on.version>1.83</bouncycastle-bcprov-jdk18on.version>
 		<jspecify.enabled>true</jspecify.enabled>
+		<spring-grpc.version>1.0.2</spring-grpc.version>
```
spencergibb, line 91: This should all be managed by spring boot. Since this won't be merged until summer time (when we start on our new minor), this will be based on boot 4.2, but in the meantime you can start with 4.1 and spring grpc 1.1

spencergibb (changes requested):

spencergibb:
While this is a good start. I think it's premature to do this. We won't integrate this until boot 4.2. I think we need to do some design work as a team as well.

KouShenhai:
@spencergibb Okay, thank you.

KouShenhai:
https://github.com/spring-cloud/spring-cloud-commons/issues/1676

PR #1696, opened 2026-06-09
docs: warn against combining @RefreshScope with @ControllerAdvice beans

### LINKED ISSUE #1452: @RefreshScope and component-scanned @ControllerAdvice beans.
This is in 2.1.3.RELEASE, and Spring Boot 2.1.7.

Beans that are annotated with both @RefreshScope and @ControllerAdvice, and which are component-scanned (as opposed to explicit configuration) have their @ModelAttribute methods executed twice per incoming request, as opposed to once.  One of the executions is via the proxy, the other is direct.

Simple example:
```java
package test;

import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;

@SpringBootApplication
public class TestMain
{
    public static void main(final String[] args)
    {
        SpringApplication.run(TestMain.class, args);
    }
}
```
----

```java
package test;

import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import org.springframework.beans.factory.annotation.Value;
import org.springframework.cloud.context.config.annotation.RefreshScope;
import org.springframework.web.bind.annotation.ControllerAdvice;
import org.springframework.web.bind.annotation.ModelAttribute;

import javax.servlet.http.HttpServletRequest;

@RefreshScope
@ControllerAdvice
public class TestAdvice
{
    private static final Logger logger = LoggerFactory.getLogger(TestAdvice.class);

    @Value("${test}")
    private String testConfigValue;
    
    @ModelAttribute
    public void adviceMethod(final HttpServletRequest httpServletRequest)
    {
        logger.info("advice method executed with " + testConfigValue);
    }
}
```
----

```java
package test;

import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RestController;

@RestController
public class TestController
{
    @GetMapping(value = "/test")
    public String test()
    {
        return "test";
    }
}
```
---

Requests to the controller log as so:

2019-08-20 14:05:07.237  INFO 5156 --- [nio-8080-exec-5] test.TestAdvice                          : advice method executed with testvalue
2019-08-20 14:05:07.237  INFO 5156 --- [nio-8080-exec-5] test.TestAdvice                          : advice method executed with testvalue

Remove the @RefreshScope, and they log just once.

won-seoop:
## Summary

Closes spring-cloud/spring-cloud-config#1452

When a `@ControllerAdvice` bean is also annotated with `@RefreshScope` and is component-scanned, Spring MVC detects it in two ways: once through the scope proxy and once directly. This causes methods annotated with `@ModelAttribute` (and potentially other MVC lifecycle methods) to execute **twice per request**, which is a subtle and hard-to-debug behavior.

The recommended workaround (confirmed by @spencergibb in the issue) is to avoid putting `@RefreshScope` on the `@ControllerAdvice` and instead extract the refreshable configuration into a separate `@RefreshScope` or `@ConfigurationProperties` bean that the advice injects as a normal dependency.

Added a `WARNING` callout to the `@RefreshScope` section in `application-context-services.adoc` documenting this limitation and the recommended approach.

## Test plan

- [ ] Docs build with Antora renders the WARNING callout correctly
- [ ] Wording is consistent with maintainer guidance from the issue

### DIFF
--- a/docs/modules/ROOT/pages/spring-cloud-commons/application-context-services.adoc
+++ b/docs/modules/ROOT/pages/spring-cloud-commons/application-context-services.adoc
@@ -214,6 +214,10 @@ Specifically, anything that depends on those beans cannot rely on them being upd
 In that case, it is rebuilt on a refresh and its dependencies are re-injected.
 At that point, they are re-initialized from the refreshed `@Configuration`).
 
+WARNING: Do not combine `@RefreshScope` with Spring MVC lifecycle beans such as `@ControllerAdvice`.
+Spring MVC detects `@ControllerAdvice` beans both through the scope proxy and directly, which causes `@ModelAttribute` methods to execute twice per request.
+Instead, extract refreshable configuration into a separate `@RefreshScope` bean (or a `@ConfigurationProperties` bean) that your `@ControllerAdvice` injects as a regular dependency.
+
 NOTE:  Removing a configuration value and then performing a refresh will not update the presence of the configuration value.
 The configuration property must be present in order to update the value after a refresh.  If you are relying on the presence of
 a value in your application you might want to switch your logic to rely on its absence instead.  Another option would be to rely


ryanjbaxter:
I would like to confirm this is still an issue as the original issue is quite old

PR #1730, opened 2026-08-25
Make GenericScope proxy locking conditional on ScopeCache

### LINKED ISSUE #630: GenericScope with ThreadLocalScopeCache heavy concurrency contention on ReadWriteLock
Symptom: 
* Application is not processing as many concurrent requests as configured in message listener container
* Thread dumps of application show many threads WAITING on the same ReadWriteLock.

Situation:
GenericScope is used with ThreadLocalScopeCache to scope some beans to a single message-processing operation. (It is a like a web "RequestScope" but for message listeners) At the end of each message processing operation, `GenericScope.destroy()` is invoked to end the life cycle of these message-scoped beans and clear the ThreadLocalScopeCache for that thread.

Since GenericScope uses pluggable ScopeCaches, it never occurred to me that it would employ a concurrent locking mechanism that bypassed the scope cache and locks on a global scope. The documentation doesn't mention it. But unfortunately, this is what happens. Whenever an operation is invoked on a bean coming from the scope, `GenericScope$LockedScopedProxyFactoryBean` acquired a read lock on **the name of the bean**, instead of on the instance of the bean. 

`ReadWriteLock readWriteLock = this.scope.getLock(this.targetBeanName);`

Even though the beans from a ThreadLocalScopeCache can only ever be seen by the thread to which they are scoped, all calls on all threads to the same bean share the same lock. This wouldn't cause any contention in itself, but the reason for acquiring the read lock, is that calls to the bean should be disallowed while the target of the bean is being destroyed. That's why the `destroy()` and `destroy(String beanName)` methods in GenericScope lock the corresponding write lock for the **bean name**.

`Lock lock = this.locks.get(wrapper.getName()).writeLock();`

Unfortunately that effect of that is that even if only one Threads' ThreadLocalScope is being destroyed at the end of a message, it still write-locks all beans in all threads at the same time, heavily reducing concurrency.
In fact, for a ThreadLocalScope, no locking is required at all, as no concurrent access to the beans in the scope is possible.

This leads me to the conclusion that this locking behaviour should be implemented in the implementation of `ScopeCache`, not in `GenericScope`, as the required behaviour varies with the `ScopeCache` implementations.

The locking-behaviour was introduced in 2017 in this commit: https://github.com/spring-cloud/spring-cloud-commons/commit/c28ca5de91408e6c64a2c9a3e1977b8f104940ac#diff-afa0715eafc2b0154475fe672dab70e4

The description suggests that it was done to fix a problem in a subclass of `GenericScope`: `RefreshScope`.

Unfortunately this had the effect of making the GenericScope much less generic, and the combination of GenericScope with a ThreadLocalScopeCache not very usable anymore.

Workarounds: 
In our case, the beans currently in the scope are simple and do not require life cycle callbacks (yet), so we can get by with just invoking `ThreadLocalScopeCache.clear()` and not doing life cycle management on the beans.
Another workaround would be to copy the code of `GenericScope` and removing the locking behaviour.

Solutions:
The details of the problem in RefreshScope that were fixed with the introduction of locking are not clear to me, but it seems to me that the locking behaviour should be in the `ScopeCache` implementation, in this case `StandardScopeCache`, so that the GenericScope can still be meaningfully used with other ScopeCache implementations that do not require locking.

akenra:
## Context

`GenericScope` guards every proxied invocation with a read lock keyed by the bean *name*, and `destroy()` takes the matching write lock. Those locks live in a single map on the scope instance, so **all** scope instances sharing one `GenericScope` contend - even when the configured `ScopeCache` confines instances to a single thread (e.g. `ThreadLocalScopeCache`). For message-listener-style usage (one thread-local scope per consumer thread), a single instance's teardown write-locks the bean name for every thread.

The analysis in #630 was endorsed by @dsyer back in 2020:

> The analysis looks good. Do we have a pull request?

Six years later - here is that pull request. Nothing was ever merged in between; the locking code is unchanged since the original 2017 commit referenced in the issue.

## What this PR does

1. **New `ScopeCache.requiresLocking()`** (`default boolean`, returns `true`) documenting the contract: caches whose instances are confined to a single thread return `false`. Concurrency semantics belong to the cache implementation - the reporter's instinct - without contorting the storage-only contract into an invocation-guard.
2. **`ThreadLocalScopeCache` overrides it to `false`**, and `GenericScope`'s `LockedScopedProxyFactoryBean` skips read-lock acquisition entirely when the cache does not require locking. The gh-349 `UndeclaredThrowableException` unwrap is preserved on both paths; destruction-side write locks are intentionally left untouched (uncontended when no readers exist, keeps the diff minimal).
3. **Bonus contention fix in the same spirit:** `BeanLifecycleWrapper` synchronized on `this.name` - a shared, typically-interned String - so independent scope instances creating same-named beans serialized against each other during lazy creation and destruction. The monitor is now the wrapper instance itself, which is exactly the state being guarded. Single-wrapper-per-name-per-cache holds via the existing `putIfAbsent` semantics of both in-tree caches.

## Backward compatibility

- Default behavior is unchanged by construction: `StandardScopeCache` (and therefore `RefreshScope`) never override `requiresLocking()`, so every refresh path keeps identical locking. All existing `RefreshScope*Tests` pass unmodified.
- Adding a default method preserves binary compatibility for already-compiled custom `ScopeCache` implementations.
- No framework-internal code calls `setScopeCache(...)` with a non-default cache today; the affected surface is custom scopes like the reporter's.

## Testing

- New `GenericScopeLockingTests` (first direct test coverage this area has had): contract defaults, proxy invocation consults locks for `StandardScopeCache` / skips for `ThreadLocalScopeCache`, and cross-instance creation of same-named beans proceeds without blocking. Tests went red before the fix (compile gap + behavioral failures on the old `GenericScope`) and green after.
- Full `spring-cloud-context` suite: 232 tests, no failures from this change. (One pre-existing, unrelated error in `EncryptorFactoryTests.testWithRsaPrivateKey` reproduces identically on clean `main`; it is environment-related PEM parsing and untouched here.)

Fixes #630

## Notes for reviewers

- Happy to split the `BeanLifecycleWrapper` monitor-granularity fix into a separate commit/PR if preferred.
- Pre-existing nits observed but deliberately not touched here: non-volatile `callback` field written outside synchronization, and a potential NPE in `destroy(name)` if the lock map lacks an entry.

### DIFF
--- a/spring-cloud-context/src/main/java/org/springframework/cloud/context/scope/GenericScope.java
+++ b/spring-cloud-context/src/main/java/org/springframework/cloud/context/scope/GenericScope.java
@@ -304,12 +304,20 @@ public class GenericScope
 		return this.locks.get(beanName);
 	}
 
+	boolean requiresProxyLocking() {
+		return this.cache.requiresLocking();
+	}
+
 	private static class BeanLifecycleWrapperCache {
 
 		private final ScopeCache cache;
 
 		BeanLifecycleWrapperCache(ScopeCache cache) {
 			this.cache = cache;
+		}
+
+		boolean requiresLocking() {
+			return this.cache.requiresLocking();
 		}
 
 		public BeanLifecycleWrapper remove(String name) {
@@ -368,7 +376,7 @@ public class GenericScope
 
 		public Object getBean() {
 			if (this.bean == null) {
-				synchronized (this.name) {
+				synchronized (this) {
 					if (this.bean == null) {
 						this.bean = this.objectFactory.getObject();
 					}
@@ -381,7 +389,7 @@ public class GenericScope
 			if (this.callback == null) {
 				return;
 			}
-			synchronized (this.name) {
+			synchronized (this) {
 				Runnable callback = this.callback;
 				if (callback != null) {
 					callback.run();
@@ -464,16 +472,19 @@ public class GenericScope
 				return invocation.proceed();
 			}
 			Object proxy = getObject();
-			ReadWriteLock readWriteLock = this.scope.getLock(this.targetBeanName);
-			if (readWriteLock == null) {
-				if (logger.isDebugEnabled()) {
-					logger.debug("For bean with name [" + this.targetBeanName
-							+ "] there is no read write lock. Will create a new one to avoid NPE");
+			Lock lock = null;
+			if (this.scope.requiresProxyLocking()) {
+				ReadWriteLock readWriteLock = this.scope.getLock(this.targetBeanName);
+				if (readWriteLock == null) {
+					if (logger.isDebugEnabled()) {
+						logger.debug("For bean with name [" + this.targetBeanName
+								+ "] there is no read write lock. Will create a new one to avoid NPE");
+					}
+					readWriteLock = new ReentrantReadWriteLock();
 				}
-				readWriteLock = new ReentrantReadWriteLock();
+				lock = readWriteLock.readLock();
+				lock.lock();
 			}
-			Lock lock = readWriteLock.readLock();
-			lock.lock();
 			try {
 				if (proxy instanceof Advised advised) {
 					ReflectionUtils.makeAccessible(method);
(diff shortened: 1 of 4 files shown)


ryanjbaxter:
Cherry picked into 5.0.x as well

PR #1732, opened 2026-08-31
Update latest Spring Antora UI

ngocnhan-tran1996:
Spring Cloud Commons and the other Spring Cloud projects are still using the old Antora UI. The main difference is the navbar compared to [spring.io](https://spring.io/) and the reference docs.

If this is not intentional, I will create a PR to update the `doc-build` branch.


ryanjbaxter:
Thank you for opening this PR, but we have updated everything across all projects at this point

PR #1739, opened 2026-09-10
Respect discovery client order for reactive instance lookup

### LINKED ISSUE #1542: ReactiveCompositeDiscoveryClient does not respect discovery clients order
# Description
The current implementation of  `ReactiveCompositeDiscoveryClient` does not respect the order of discovery clients. Instead of returning results in the specified order, it returns the first non-empty Flux from the fastest responding client. This behavior differs from the non-reactive `CompositeDiscoveryClient`, which respects the order of discovery clients.

# Problem Analysis
The default behavior of `ReactiveCompositeDiscoveryClient` can lead to unexpected results when multiple discovery clients are configured with different priorities or when specific ordering is required for service discovery.

# Proposed Solution
To address this issue, a custom implementation of `ReactiveCompositeDiscoveryClient` can be provided. This implementation ensures that discovery clients are queried in the specified order and returns results from the first non-empty response.

### Author’s Note
_Although the author is not an expert in reactive Java, this solution appears to work correctly based on testing. Further review by experienced developers is recommended to ensure optimal performance and correctness._

```java
@Bean
@Primary
ReactiveCompositeDiscoveryClient reactiveCompositeDiscoveryClient(List<ReactiveDiscoveryClient> discoveryClients) {
    return new ReactiveCompositeDiscoveryClient(discoveryClients) {
        @Override
        public Flux<ServiceInstance> getInstances(String serviceId) {
            if (discoveryClients == null || discoveryClients.isEmpty()) {
                return Flux.empty();
            }

            return Flux.fromIterable(discoveryClients)
                    .concatMap(client -> client.getInstances(serviceId).collectList())
                    .filter(Predicate.not(List::isEmpty))
                    .next()
                    .flatMapMany(Flux::fromIterable);
        }
    };
}
```

# Usage Instructions
To apply this solution, exclude the default `ReactiveCompositeDiscoveryClientAutoConfiguration` and register the custom bean in your Spring configuration.

# Notes
- This implementation preserves the order of discovery clients.
- It collects results from each client sequentially.
- The first non-empty result is returned, respecting the configured order.

hutiefang76:
When a higher-priority discovery client responds slowly, `ReactiveCompositeDiscoveryClient` currently returns the faster lower-priority client's instances. This contradicts the ordering applied in its constructor and the behavior of `CompositeDiscoveryClient`.

Try the publishers in order with `switchIfEmpty`, falling back only when a client completes without instances. The result stays streaming and preserves all instances from the first non-empty client. The shared `CloudFlux` behavior is unchanged.

Fixes #1542.

Validation on JDK 21:

```sh
./mvnw -B -ntp -pl spring-cloud-commons -am test \
  -Dtest=ReactiveCompositeDiscoveryClientTests,ReactiveCompositeDiscoveryClientAutoConfigurationTests,CompositeDiscoveryClientUnitTests \
  -Dsurefire.failIfNoSpecifiedTests=false
```

14 tests passed, with formatting and Checkstyle checks enabled. The delayed-priority regression failed on the original implementation. Coverage also includes empty fallback, multiple instances, error propagation, and not subscribing to lower-priority publishers once a result is found. The full repository suite was not run.

Developed with Codex assistance; the diff and regression results were reviewed.

### DIFF
--- a/spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/composite/reactive/ReactiveCompositeDiscoveryClient.java
+++ b/spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/composite/reactive/ReactiveCompositeDiscoveryClient.java
@@ -16,14 +16,12 @@
 
 package org.springframework.cloud.client.discovery.composite.reactive;
 
-import java.util.ArrayList;
 import java.util.List;
 
 import reactor.core.publisher.Flux;
 
 import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.client.discovery.ReactiveDiscoveryClient;
-import org.springframework.cloud.commons.publisher.CloudFlux;
 import org.springframework.core.annotation.AnnotationAwareOrderComparator;
 
 /**
@@ -51,11 +49,11 @@ public class ReactiveCompositeDiscoveryClient implements ReactiveDiscoveryClient
 		if (discoveryClients == null || discoveryClients.isEmpty()) {
 			return Flux.empty();
 		}
-		List<Flux<ServiceInstance>> serviceInstances = new ArrayList<>();
+		Flux<ServiceInstance> serviceInstances = Flux.empty();
 		for (ReactiveDiscoveryClient discoveryClient : discoveryClients) {
-			serviceInstances.add(discoveryClient.getInstances(serviceId));
+			serviceInstances = serviceInstances.switchIfEmpty(discoveryClient.getInstances(serviceId));
 		}
-		return CloudFlux.firstNonEmpty(serviceInstances);
+		return serviceInstances;
 	}
 
 	@Override
--- a/spring-cloud-commons/src/test/java/org/springframework/cloud/client/discovery/composite/reactive/ReactiveCompositeDiscoveryClientTests.java
+++ b/spring-cloud-commons/src/test/java/org/springframework/cloud/client/discovery/composite/reactive/ReactiveCompositeDiscoveryClientTests.java
@@ -16,12 +16,15 @@
 
 package org.springframework.cloud.client.discovery.composite.reactive;
 
+import java.time.Duration;
+
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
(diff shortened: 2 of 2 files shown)


spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/composite/reactive/ReactiveCompositeDiscoveryClient.java
```diff
@@ -51,11 +49,11 @@ public Flux<ServiceInstance> getInstances(String serviceId) {
 		if (discoveryClients == null || discoveryClients.isEmpty()) {
 			return Flux.empty();
 		}
-		List<Flux<ServiceInstance>> serviceInstances = new ArrayList<>();
+		Flux<ServiceInstance> serviceInstances = Flux.empty();
 		for (ReactiveDiscoveryClient discoveryClient : discoveryClients) {
-			serviceInstances.add(discoveryClient.getInstances(serviceId));
+			serviceInstances = serviceInstances.switchIfEmpty(discoveryClient.getInstances(serviceId));
```
ryanjbaxter, line 54: What happens if the user now wants the previous behavior?  For example it might be desirable in many cases to move to the next priority client if it returns and the higher priority client is still has not responded.  I don't necessarily think that would be a bad thing.

hutiefang76: That is a valid compatibility concern. With this patch, a slow or non-terminating higher-priority publisher delays or prevents fallback; the old race can return a lower-priority client's result in that situation. Applications may rely on that latency behavior even though the client list is sorted.

I suggest keeping the existing fastest-non-empty behavior as the default and making strict priority ordering opt-in, with tests for both modes. That would address #1542 without changing existing applications by default. Would a configuration property on the reactive composite auto-configuration be the preferred way to expose this, or would you rather keep it as a constructor-level option? I'll align the API choice before extending the patch.

ryanjbaxter: > Would a configuration property on the reactive composite auto-configuration 

Meaning you would have two different beans one for the existing behavior and another one for the new behavior that would be activated by a property?

Issue #1733, opened 2026-09-02
Refresh drops @ConfigurationProperties values that no property source carries


guerricm:
Since spring-cloud-context 5.0.2, `ConfigurationPropertiesRebinder` resets every rebindable bean to
the defaults of a template instance before re-binding it. Whatever was set programmatically after
the initial binding is gone, because the re-binding only restores what the `Environment` carries.

Java 25, Spring Boot 4.1.1, Spring Cloud 2025.1.3, spring-cloud-context 5.0.3,
spring-cloud-netflix-eureka-client 5.0.2.

Sample, three Initializr projects, the application under test has no code of its own:
https://github.com/guerricm/bug-refresh

`POST /actuator/refresh` answers `[]`, so no key of the `Environment` changed. `eureka.instance`
changes anyway:

```
BEFORE                                      AFTER
hostname      : localhost                   hostname      : localhost
ipAddress     : 192.168.1.117               ipAddress     : null
metadataMap   : {management.port=8080}      metadataMap   : {}
nonSecurePort : 8080                        nonSecurePort : 80
```

An empty change set is not a condition of the bug, only the shortest proof that the refresh itself
drives the reset. `hostname` survives because `application.yaml` carries it.

The Eureka instance is then cancelled from the registry and cannot register again:
`ApplicationResource.addInstance` answers `400 Missing ip address` for a blank `ipAddr`. The
application keeps running and is simply absent from the registry.

`resetBeanToDefaults` builds its template through any no-arg constructor, whatever its visibility.
`EurekaInstanceConfigBean` declares a `private EurekaInstanceConfigBean() {}` next to the public
`EurekaInstanceConfigBean(InetUtils)` that computes `ipAddress` and `hostname`, so the template
carries `ipAddress = null`, `nonSecurePort = 80` and an empty `metadataMap`. The re-binding restores
none of the three: they come from the public constructor, from `EurekaAutoServiceRegistration`, and
from `EurekaClientAutoConfiguration:204`.

Absent in 5.0.1, introduced in 5.0.2 by #1680, extended in 5.0.3. The sample carries a
`-Pcontext-5.0.1` profile that downgrades that artifact alone; the same refresh then leaves the bean
untouched. #1680 targets a property removed from the `Environment` while the bean keeps its value,
and the reset cannot tell that apart from a value that was never in the `Environment`.

Related: #1727, same pull request, nested objects passing through `null` during the rebind. That one
[...]

Issue #1727, opened 2026-08-22
`resetBeanToDefaults` transiently nulls nested `@ConfigurationProperties` objects, NPEing concurrent request threads


rsheldon-dealeron:
### Summary

Since spring-cloud-context 5.0.2 (`resetBeanToDefaults`, #1680), a rebind empties the live
`@ConfigurationProperties` singleton before repopulating it. For a bean with **nested** property
objects, the observable intermediate state is not merely a stale or new scalar — it is `null`, so a
concurrent reader does not read a wrong value, it throws `NullPointerException`.

This is the same window PR #1709 reported and #1721 partially addressed (writer/writer only). #1709
was closed with the observation acknowledged; this report adds the failure mode it did not cover — a
hard failure rather than a wrong value — and evidence of how often it is hit in a framework where
every service has nested properties read on request threads.

### Environment

- spring-cloud-context 5.0.2 and 5.0.3 (trains 2025.1.2, 2025.1.3); Spring Boot 4.0.8. Not present in
  4.3.x/4.2.x.

### What happens

```java
@Bean
@ConfigurationProperties(prefix = "my.service")
MyProperties myProperties() { return new MyProperties(); }

public class MyProperties {
    private Timeout timeout = null;          // populated only by binding
    public Timeout getTimeout() { return timeout; }
    // ...
}
```

`resetProperties` walks the `BeanWrapper` descriptors and, for every writable property, does
`target.setPropertyValue(name, defaultsWrapper.getPropertyValue(name))`. For `timeout` that is
`setTimeout(null)`. Between `resetBeanToDefaults(bean)` and the binding inside
`initializeBean(bean, name)`, every reader of `myProperties.getTimeout().getX()` throws NPE.

`spring.cloud.refresh.never-reset-nested-types` does not help: it controls recursion *into* a non-null
nested value, not the top level `setTimeout(null)`.

### Reproduction

[rebind-reset-repro.zip](https://github.com/user-attachments/files/31326583/rebind-reset-repro.zip)  is a self-contained project — Spring Boot 4.0.8 with
`spring-cloud-context` as the only other dependency, no Spring Cloud Config, no Consul, no
`@RefreshScope`. The test is also attached on its own as [RebindEmptiesNestedPropertiesTest.java.txt](https://github.com/user-attachments/files/31326601/RebindEmptiesNestedPropertiesTest.java.txt),
[...]

Issue #1698, opened 2026-06-17
ConfigurationPropertiesRebinder StackOverflowError for certain ConfigurationProperties beans


ikaakkola:
spring-cloud-commons version 5.0.2 introduced changes to resetting beans / properties in ConfigurationPropertiesRebinder (https://github.com/spring-cloud/spring-cloud-commons/pull/1680).

These changes cause a StackOverflowError for (at least) Configuration classes that have a field of type `javax.net.ssl.SSLContext` , e.g.

```
package example;

import javax.net.ssl.SSLContext;
import org.springframework.boot.context.properties.ConfigurationProperties;
import org.springframework.context.annotation.Configuration;

@Configuration
@ConfigurationProperties(prefix = "foo")
public class FooProperties {

  private transient SSLContext sslContext;

  public synchronized SSLContext getSslContext() {
    if (sslContext != null) {
      return sslContext;
    }
    // initialize a new SSLcontext
    // ..
    // ..

    return sslContext;
  }
}
```

This works fine with the ConfigurationPropertiesRebinder from release 5.0.1 , but on 5.0.2 causes a StackOverflowError:

```
[error] java.lang.StackOverflowError
[error]         at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:101)
[error]         at java.base/java.lang.reflect.Method.invoke(Method.java:580)
[error]         at org.springframework.beans.BeanWrapperImpl$BeanPropertyHandler.getValue(BeanWrapperImpl.java:270)
[error]         at org.springframework.beans.AbstractNestablePropertyAccessor.getPropertyValue(AbstractNestablePropertyAccessor.java:623)
[error]         at org.springframework.beans.AbstractNestablePropertyAccessor.getPropertyValue(AbstractNestablePropertyAccessor.java:611)
[error]         at org.springframework.cloud.context.properties.ConfigurationPropertiesRebinder.resetProperties(ConfigurationPropertiesRebinder.java:216)
[error]         at org.springframework.cloud.context.properties.ConfigurationPropertiesRebinder.resetProperties(ConfigurationPropertiesRebinder.java:231)
[error]         at org.springframework.cloud.context.properties.ConfigurationPropertiesRebinder.resetProperties(ConfigurationPropertiesRebinder.java:231)
[...]

PR #1740, opened 2026-09-12
Apply never-reset-nested-types to writable properties

kdelay:
`spring.cloud.refresh.never-reset-nested-types` (1c3d7b0d, gh-1698) only takes effect on a property with no setter. A writable one is overwritten with the default instance's value in an earlier branch of `resetProperties`, before the exclusion is consulted, so it is still replaced on rebind and reads as `null` while the default leaves the field unset. That is gh-1727.

This consults the exclusion in the writable branch too, by declared type and by the runtime class. Nothing changes when the property is unset, and the reset stays best effort, as in gh-1733.

The test fails on `main` at 439ed3c (`getWritable()` is null after rebind) and passes with the fix. `./mvnw -pl spring-cloud-context verify`: 234 tests, 0 failures, checkstyle and javaformat clean.

Scope is that escape hatch; a primitive cannot be listed in the property.


ryanjbaxter:
Can you submit this against the 5.0.x branch?

kdelay:
Retargeted to 5.0.x in 4a4529b. The commit cherry-picked cleanly. On 5.0.x the new test fails without the fix (AssertionError at ConfigurationPropertiesRebinderResetExclusionIntegrationTests.java:86) and passes with it; ./mvnw -pl spring-cloud-context verify gives 234 tests, 0 failures, 0 checkstyle violations.

ryanjbaxter:
Can you look at the failing PR build?

kdelay:
The failure is a compile error in spring-cloud-loadbalancer, which this PR does not touch (the diff is spring-cloud-context only): reactor.cache and reactor.retry not found, both from reactor-extra.

I built unmodified 5.0.x (dd0d9f6) locally today: ./mvnw -pl spring-cloud-loadbalancer -am -DskipTests compile is BUILD SUCCESS, resolving reactor-extra 3.6.1, which has both packages. So the runner seems to have picked up a reactor-extra without them.

I pushed b362067, an identical tree to 4a4529b, for a fresh run; it needs a workflow approval.

PR #1741, opened 2026-09-12
Compare config tree values by content on refresh

### LINKED ISSUE #1687: Change detection on refreshEnvironment not working properly with configtree
Spring Boot: 4.0.5
Spring Cloud: 2025.1.1
Related issue: https://github.com/spring-projects/spring-boot/issues/50181

**Describe the bug**
When using configtree: for properties and using [ContextRefresher#refreshEnvironment()](https://github.com/spring-cloud/spring-cloud-commons/blob/main/spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/ContextRefresher.java#L99-L105), all properties from configtree property sources will be returned as changed, even if the actual value did not.

The reason is that [ContextRefresher#changes(Map, Map)](https://github.com/spring-cloud/spring-cloud-commons/blob/main/spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/ContextRefresher.java#L145) checks for equality, but the returned [PropertyFileContent](https://github.com/spring-projects/spring-boot/blob/main/core/spring-boot/src/main/java/org/springframework/boot/env/ConfigTreePropertySource.java#L265) does not implement equals/hashCode.

After posting the issue in the Spring Boot repo, the response was that this should be raised here.

**Sample**
Example project: https://github.com/schosin/spring-cloud-commons-1687

A Spring Boot application with spring-cloud-context using this configuration, and the directory containing atleast one file will trigger the issue with the auto configured ContextRefresher.

```yaml
spring:
  config:
    import:
      - configtree:some/path/to/directory/
```

arimu1:
## Summary

- Fix change detection in `ContextRefresher` when the environment uses Spring Boot `ConfigTreePropertySource`.
- `ConfigTreePropertySource.Value` (including `PropertyFileContent`) does not implement `equals`/`hashCode`, so every refresh incorrectly reported all config tree keys as changed.
- Compare config tree property values by content via `CharSequence`/`toString()` instead of reference equality.

Fixes gh-1687

## Test plan

- [x] `./mvnw -pl spring-cloud-context -am test`
- [x] Added `ContextRefresherConfigTreeTests` covering unchanged config tree refresh (no spurious keys) and changed content (key reported).


spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/ContextRefresher.java
```diff
@@ -154,6 +156,17 @@ else if (!Objects.equals(before.get(key), after.get(key))) {
 		return result;
 	}
 
+	private boolean equal(Object one, Object two) {
+		if (Objects.equals(one, two)) {
+			return true;
+		}
+		if (one instanceof ConfigTreePropertySource.Value oneValue
```
ryanjbaxter, line 163: I think we need to be a little more careful here since we are dealing with files on K8S and those files can be updated/removed between the time the refresh occurs and the comparison happens.  If one or both of the files is gone when the comparison happens an IllegalStateException may occur so we should wrap this in a try/catch to deal with that case and return false.  We should probably add a test to cover this if possible.

spring-cloud-context/src/main/java/org/springframework/cloud/context/refresh/ContextRefresher.java
```diff
@@ -154,6 +156,17 @@ else if (!Objects.equals(before.get(key), after.get(key))) {
...
+				&& two instanceof ConfigTreePropertySource.Value twoValue) {
+			return oneValue.toString().contentEquals(twoValue);
```
ryanjbaxter, line 165: I think we might want to compare bytes rather than use `toString()` since some of these values may be binary data (certs/keystores)

ryanjbaxter (changes requested):

arimu1:
Thanks for the review @ryanjbaxter — addressed both points in e57b4f7.

**Missing K8s files during comparison**
`configTreeValuesEqual` now reads both values inside a try/catch. If a mounted file disappears (or is otherwise unreadable) between refresh and comparison, we log at debug and return `false` so refresh treats it as a change instead of throwing `IllegalStateException`.

Added `equalReturnsFalseWhenConfigTreeFileIsMissingDuringComparison`: two `ALWAYS_READ` values are captured, the backing file is deleted, then `equal` is invoked — expects `false` without an exception.

**Byte comparison (binary mounts)**
Comparison uses `Arrays.equals` on bytes from `Value.getInputStream()` instead of `toString().contentEquals(...)`.

Added `refreshEnvironmentDoesNotReportUnchangedBinaryConfigTreeProperties` with synthetic cert-like bytes to ensure unchanged binary config tree properties do not trigger a false refresh.

**Tests run locally**
`mvn -pl spring-cloud-context -am test -Dtest=ContextRefresherConfigTreeTests` (Temurin 21) — all pass.

**CI note (run 34726626743)**
That failure was in `spring-cloud-loadbalancer` (`reactor.cache` / `reactor.retry` missing); `spring-cloud-context` (including our module tests) succeeded. Looks upstream/main unrelated to this PR — happy to re-check after the next CI run on this push.

ryanjbaxter (approved):
Could you submit this against the 5.0.x branch?

PR #1743, opened 2026-09-19
Match never-reset-nested-types against the declared property type

kdelay:
## What

`resetProperties` checks the `spring.cloud.refresh.never-reset-nested-types` exclusion against the declared property type only for a writable property (gh-1740). A read-only property is tested against the runtime class of its value alone, and the exclusion is a name prefix match, so a value of a differently named subtype is not recognised: the rebind descends into it and resets its fields to the defaults of a fresh instance.

## Fix

Test the declared type once per property; keep the value check on the writable branch.

## Verification

Added a test with a read-only property declared as the excluded type holding a subclass instance. Reverting the fix fails it while the two existing tests pass. `./mvnw -pl spring-cloud-context test`: 235 tests, 0 failures, including javaformat and checkstyle.


spring-cloud-context/src/main/java/org/springframework/cloud/context/properties/ConfigurationPropertiesRebinder.java
```diff
@@ -290,10 +290,16 @@ private void resetProperties(Object bean, Object defaults, Set<Object> visited)
 			if ("class".equals(propertyName)) {
 				continue;
 			}
+			// A property whose declared type the user excluded is left alone whatever its
```
ryanjbaxter: I would put this inside the try block below

ryanjbaxter (commented):
Can you submit this against the 5.0.x branch?

kdelay:
Retargeted to 5.0.x: the commit is now cherry-picked onto 21ebbd8 (cf5b7a5), same tree change as before.

Re-verified on that base: reverting only `ConfigurationPropertiesRebinder` leaves `excludedNestedTypeIsNotResetWhenTheValueIsASubclass` failing (AssertionFailedError at line 101) with the other two passing; with the change all 3 pass.

spring-cloud-context/src/main/java/org/springframework/cloud/context/properties/ConfigurationPropertiesRebinder.java
ryanjbaxter: Please address this comment

kdelay:
Done in 7c6c856: the declared-type check now sits at the top of the try block instead of before it.

Re-verified on 5.0.x: reverting only `ConfigurationPropertiesRebinder` still leaves `excludedNestedTypeIsNotResetWhenTheValueIsASubclass` failing (AssertionFailedError at line 101) with the other two passing; with the change all 3 pass. `./mvnw -pl spring-cloud-context test` is 235 tests, 0 failures.

